PR #5, opened 2026-01-20
Add Operations trait

svix-gabriel:
I want to add this trait ASAP so I can build the operations for Stream around the API.

I'm sure the trait might change, but for now I essentially just copied what was specified here http://github.com/svix/rfc/pull/30/files#diff-1f8e708b840474d3072b1c965eb090a3e30b26e8b7036c6e0ae47ef36ffb09abR54.

The only difference is I modified the API to take in `AppState`, rather than a specific `fjall::Database`. I don't know what kind of state each operation will need, so better to be flexible from the beginning.

### DIFF
--- a/src/core/operations.rs
+++ b/src/core/operations.rs
@@ -1,0 +1,12 @@
+use serde::{Deserialize, Serialize};
+
+use crate::AppState;
+
+/// Operations as defined in the RFC https://github.com/svix/rfc/pull/30/files#diff-1f8e708b840474d3072b1c965eb090a3e30b26e8b7036c6e0ae47ef36ffb09abR67.
+pub trait Operation<'a>: Serialize + Deserialize<'a> {
+    type ApplyOutput;
+    type ApplyError;
+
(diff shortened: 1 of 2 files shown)


src/core/operations.rs
```diff
@@ -0,0 +1,12 @@
...
+
+use crate::AppState;
+
+/// Operations as defined in the RFC https://github.com/svix/rfc/pull/30/files#diff-1f8e708b840474d3072b1c965eb090a3e30b26e8b7036c6e0ae47ef36ffb09abR67.
+pub trait Operation<'a>: Serialize + Deserialize<'a> {
+    type ApplyOutput;
+    type ApplyError;
+
+    #[allow(async_fn_in_trait)]
+    async fn apply(self, state: &AppState) -> Self::ApplyOutput;
```
svix-jbrown, line 11: this should return `Result<Self::ApplyOutput, Self::ApplyError>`, where ApplyError indicates a critical failure (e.g., out of disk) and that the operation cannot be retried right now

PR #8, opened 2026-01-20
Initial Stream scaffolding and create_stream RPC call

svix-gabriel:
This PR does two main things

1. NUKE the existing stream code, as it's not how we'll implement Stream
2. Add the initial `create_stream` RPC call, and a very rudimentary test.

I want to get this PR reviewed ASAP, because it also sets up some organization patterns I want.

I left quite a few FIXME's, as there are things I depend on that aren't implemented yet (or other folks are working on). But having the scaffolding for tests and the stream crate should make it easy to iterate going forward.


crates/stream/src/tables.rs
```diff
@@ -0,0 +1,96 @@
...
+        let prefix = Self::TABLE_PREFIX;
+        format!("{prefix}{key}").into()
+    }
+
+    fn fjall_key(&self) -> fjall::UserKey {
+        Self::make_fjall_key(self.get_key())
+    }
+
+    fn fjall_value(&self) -> Result<fjall::UserValue> {
+        // FIXME(@svix-gabriel) - we probably want to use something other than serde_json here.
```
svix-lucho: The fjall examples use `rmp_serde`, which worked for me as a drop-in replacement.

crates/stream/src/tables.rs
svix-lucho, line 1: I kind of want this for other modules 😛 

I started working on this the other way around (no abstracion over fjall so I ended up with very ugly code), and now I'm going to need a similar abstraction. Can we move this to a common place? not just for stream

crates/stream/src/tables.rs
```diff
@@ -0,0 +1,96 @@
...
+            .metadata_tables
+            .get(&key)?
+            .map(Self::from_fjall_value)
+            .transpose()
+    }
+}
+
+// IMPORTANT. Since these are all shared in the same fjall::Keyspace, the table prefixes must be unique.
+static_assertions::const_assert!(coyote_utils::are_all_unique(&[
+    NameToStreamRow::TABLE_PREFIX,
```
svix-james, line 11: Do they need to be in the same keyspace? Just a question -- I'm totally unclear on best practices there. In terms of "batching", though, that's database level, not keyspace level?

crates/stream/src/tables.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    fn insert(&self, state: &State) -> Result<()> {
+        state
+            .metadata_tables
+            .insert(self.fjall_key(), self.fjall_value()?)?;
+        Ok(())
+    }
+
+    fn fetch(state: &State, key: &Self::Key) -> Result<Option<Self>> {
```
svix-james: Should this stuff be async and call spawn_blocking inside? Else we'll have to do that like everywhere else in the code, right?

crates/stream/src/tables.rs
svix-gabriel: Updated.

I know we might not use messagepack for the internal serialization, but since we'll use messagepack anyway, can't hurt to make the substitute now

crates/stream/src/tables.rs
svix-gabriel: Done

crates/stream/src/tables.rs
svix-gabriel: They don't, but I think the consensus from yesterday when I went over the stream RFC was to keep them all in the same keyspace.

crates/stream/src/tables.rs
svix-gabriel: TBH I don't know.

AFAICT fjall doesn't expose any kind of async interface... so if we go down this route, we'll basically have `spawn_blocking` and similar calls _everywhere_.

I kind of assumed @svix-jplatte would have opinions here, because this also ties into how do the multithreading.

One thing I was thinking: if we want to avoid using `spawn_blocking` all the time, is to basically have the architecture as

1. 1 thread runs the axum server
2. the remaining worker threads are _not_ async, and the axum server delegates all of the work to the worker threads using async channels.

If we do that, then we won't have to worry about `spawn_blocking` and similar tokio functions.

crates/stream/src/tables.rs
svix-lucho: It's also unclear to me if there's a difference between having 1 vs 10 keyspaces, for example. I know we said 'millions' of keyspaces is wrong, but I wonder if we're missing out on something if we use one per module.

crates/stream/src/tables.rs
svix-gabriel: Ok started a discussion here https://svixhq.slack.com/archives/C0A72988962/p1768953274193959

crates/fjall-utils/src/table_row.rs
```diff
@@ -0,0 +1,40 @@
...
+    const TABLE_PREFIX: &'static str;
+
+    type Key: Display;
+
+    /// Return the field used for indexing into the table.
+    fn get_key(&self) -> &Self::Key;
+
+    fn make_fjall_key(key: &Self::Key) -> fjall::UserKey {
+        let prefix = Self::TABLE_PREFIX;
+        format!("{prefix}{key}").into()
```
svix-lucho, line 19:
```suggestion
        format!("{prefix}\0{key}").into()
```

I'd add the `\0` divider here, that fjall uses in the examples

crates/stream/src/tables.rs
svix-gabriel: So I think the difference is that each keyspace is it's own LSM tree, which I think has implications for how well compaction / maintenance tasks can work.

I think the implication here is that separate keyspaces can be useful for different workloads, since the entries won't interfere w/one another? But too many keyspaces, and you can't leverage compression/compaction as easily.

... I think

crates/stream/src/tables.rs
svix-gabriel: Ok, I don't this discussion will be resolved today, so I've just wrapped the `apply_operation` call in `spawn_blocking` for now.

svix-james (approved):

PR #10, opened 2026-01-21
Some db scaffolding

svix-james:
Maybe this is helpful, maybe just clumsy, but adding some
configuration for multiple databases, which are tentatively
differentiated as "persistent" and "ephemeral". Oh, and don't
forget the "management" db, which will store config info.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ schemars.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 serde_path_to_error.workspace = true
+serde_with.workspace = true
 tokio.workspace = true
 tokio-util.workspace = true
 tower.workspace = true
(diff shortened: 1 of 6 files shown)


svix-lucho (approved):

Cargo.toml
```diff
@@ -127,6 +128,7 @@ schemars = { version = "1.2.0", features = ["jiff02"] }
 serde = { version = "1.0.184", features = ["derive", "rc"] }
 serde_json = { version = "1.0.74", features = ["arbitrary_precision", "raw_value"] }
 serde_path_to_error = "0.1.7"
+serde_with = "3.16.1"
```
svix-jplatte, line 152: This package is unreasonably heavy for what it does, IMHO. I can replace it in a follow-up PR though.

svix-jplatte (commented):
fjall docs are saying you generally should only have one db per application, and use different keyspaces for different db concerns. Why are we deviating from that?

svix-james:
> fjall docs are saying you generally should only have one db per application, and use different keyspaces for different db concerns. Why are we deviating from that?

The reason is that Tom literally wants data stored in different files / on different mount points for if someone wants local SSD / some memory-mapped thing / external network-attached / EBS device, etc., which is slower / more reliable. It will be an optional per-configuration group to specify one or the other.

svix-james:
Actually, maybe the "management" DB should purely be a keyspace within the "persistent" DB. I'll leave it separate for now, though.

PR #13, opened 2026-01-22
Fix CreateStream::apply_operation not being deterministic

svix-gabriel:
🤦 

I forgot that the "apply operation" step has to be perfectly deterministic for the HA/Raft/consensus stuff to work.

This means when I create the Operation struct, I need to know _exactly_ what I'm inserting into the DB.

Then the apply step actually does the inserts.

I guess this isn't broken _yet_, since none of the consensus stuff isn't in. But based on my understanding of how the Raft/consensus wrapper JB's working on, this is how things should be setup.

### DIFF
--- a/crates/stream/src/operations/create_stream.rs
+++ b/crates/stream/src/operations/create_stream.rs
@@ -10,10 +10,7 @@ use std::num::NonZeroU64;
 use uuid::Uuid;
 
 pub struct CreateStream {
-    timestamp: Timestamp,
-    name: String,
-    retention_period_seconds: Option<NonZeroU64>,
-    max_byte_size: Option<NonZeroU64>,
+    stream: StreamRow,
 }
 
 pub struct CreateStreamOutput {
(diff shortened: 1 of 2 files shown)


crates/stream/src/operations/create_stream.rs
```diff
@@ -27,71 +24,63 @@ pub struct CreateStreamOutput {
 
 impl CreateStream {
     pub fn new(
+        state: &State,
```
svix-james, line 27: Something about the fact that this function is doing blocking IO things makes me nervous AF.... I know we're worrying about that later, but still....

svix-james (approved):

crates/stream/src/operations/create_stream.rs
svix-gabriel: Yup.

FWIW, this is a big reason why I think we should go with the approach I suggested (1 server thread, N-1 worker threads). And then the server thread doesn't even have access to `fjall` at all, so it's (almost) impossible for the server thread to accidentally make blocking calls.

PR #14, opened 2026-01-22
Initial KV and cache implementation

svix-lucho:
Remove the existing in-memory implementations for the Cache and the KV store modules, in favor of Fjall-backed implementations. The caching module uses the KV module as the 'backend' with the addition of an eviction worker, which is not there for KV since KV should never evict.

This also includes **basic** implementations for the 'cleanup' workers:
- Cache worker: evicts values closest to their expiration when under capacity pressure
- KV worker: removes values from the store when they reach their expiration.

There are still many improvements to be made on top of this (as you'll see from the many FIXMEs), but I wanted to put this out there sooner rather than later just to keep the changeset smaller.


crates/fjall-utils/src/table_row.rs
```diff
@@ -37,4 +38,54 @@ pub trait TableRow: Sized + Serialize + DeserializeOwned {
...
+
+    fn iter(keyspace: &fjall::Keyspace) -> Result<impl Iterator<Item = Self>> {
+        Ok(keyspace.iter().map(|g| {
+            let v = g.value().expect("iter error?");
+            let r = Self::from_fjall_value(v).expect("deserialize error?");
+            r
+        }))
+    }
+
+    fn take_tx(
```
svix-gabriel: From in person convo: we probably don't need any of the `tx` helpers, since we can get away with batching.

crates/fjall-utils/src/table_row.rs
```diff
@@ -37,4 +38,54 @@ pub trait TableRow: Sized + Serialize + DeserializeOwned {
         let key = Self::make_fjall_key(key);
         keyspace.get(&key)?.map(Self::from_fjall_value).transpose()
     }
+
+    fn iter(keyspace: &fjall::Keyspace) -> Result<impl Iterator<Item = Self>> {
+        Ok(keyspace.iter().map(|g| {
```
svix-gabriel: I think this assumes all of the keys in the Keyspace are of the same logical table, which we can't make (at least with my code).

I think the way to fix this is just to use `Keyspace::prefix`https://docs.rs/fjall/latest/fjall/struct.Keyspace.html#method.prefix, and just use the`TABLE_PREFIX`.

src/lib.rs
```diff
@@ -119,8 +119,10 @@ pub async fn run_with_prefix(cfg: Configuration, listener: Option<TcpListener>)
     let app_state = AppState {
         cfg: cfg.clone(),
         db,
-        kv_store: crate::v1::modules::kv::KvStore::new(),
-        cache_store: crate::v1::modules::cache::CacheStore::new(),
+        kv_store: crate::v1::modules::kv::KvStore::new("kv_store"),
```
svix-gabriel: Since we don't need transactions, we can probably get away with having these Stores initialized by passing in the same `Database` object we initialize above, no? (Similar to what I did with 
```rust
let stream_state = stream::State::init(&db).expect("initialing stream state");
```

above)

crates/kv/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+
+        let tables = {
+            let opts = KeyspaceCreateOptions::default();
+            db.keyspace(&kv_keyspace, || opts).unwrap()
+        };
+
+        Self { db, tables }
+    }
+
+    pub fn new_temporary(namespace: &str) -> Self {
```
svix-gabriel, line 69: The way I did this was, rather than a dedicated method for creating a temporary db, I created a temporary directory in the test scaffolding, and that handles automatic cleanup for us.

If you end up changing the stores to use the same db initialized in run_with_prefix, that means you can leverage the helpers in the test code already.

See https://github.com/svix/coyote-private/blob/main/tests/it/main.rs

and 

https://github.com/svix/coyote-private/blob/main/tests/it/stream.rs#L6

svix-lucho: yup I'll address that later

src/lib.rs
svix-lucho: yes. I will do that in another PR

crates/fjall-utils/src/table_row.rs
svix-lucho: oh yeah, good catch. I fixed it.

crates/fjall-utils/src/table_row.rs
svix-lucho: Updated

Cargo.toml
svix-jplatte, line 1: I think I have the same TOML vscode extension as you, because I have this in my config

```json
    "[toml]": {
        "editor.formatOnSave": false
    },
```

to tell it to not do this ugly

```toml
dependency = { version = "x.y", features = [
    "a",
    "b",
    "c",
] }
```

thing. Can you revert those changes? Should also make rebase easier.

svix-lucho: fixed it

svix-lucho:
I'm going to merge this since I have other changes depending on this and the high level stuff has already been reviewed. I'll fix the remaining stuff in other PRs.

PR #15, opened 2026-01-22
implementation of openraft on top of fjall

svix-jbrown:
This implements an [openraft](https://crates.io/crates/openraft) backend using fjall. It doesn't implement any of the network parts yet (nothing is wired up to axum), nor any of the discovery parts. It also doesn't use the same fjall database for storage as the main system. But! It's a start and I wanted to check it in before embarking on the next multi-thousand-line PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,10 +20,12 @@ openapi = []
 
 [dependencies]
 aide.workspace = true
+anyerror.workspace = true
 anyhow.workspace = true
 axum.workspace = true
 axum-extra.workspace = true
 base64.workspace = true
+byteorder.workspace = true
 bytes.workspace = true
 config.workspace = true
 coyote-derive.workspace = true
@@ -41,11 +43,13 @@ itertools.workspace = true
 jiff.workspace = true
 jwt-simple.workspace = true
 mime.workspace = true
+openraft.workspace = true
 opentelemetry.workspace = true
 opentelemetry-http.workspace = true
 opentelemetry-otlp.workspace = true
 opentelemetry-semantic-conventions.workspace = true
 opentelemetry_sdk.workspace = true
+paste.workspace = true
 regex.workspace = true
 rmp-serde.workspace = true
 schemars.workspace = true
@@ -65,6 +69,7 @@ tracing-subscriber = { workspace = true, features = ["json"] }
 url.workspace = true
 uuid.workspace = true
 validator.workspace = true
+zip.workspace = true
 
 [dev-dependencies]
 assert_matches.workspace = true
@@ -96,11 +101,13 @@ rust-version = "1.91"
 
 [workspace.dependencies]
 aide = { version = "=0.16.0-alpha.2", features = ["axum", "axum-json", "axum-query", "macros"] }
+anyerror = { version = "0.1.13", features = ["anyhow"] }
 anyhow = "1.0.56"
 assert_matches = "1.5.0"
 axum = "0.8.8"
 axum-extra = { version = "0.12.5", features = ["typed-header"] }
 base64 = "0.22"
+byteorder = "1.5.0"
 bytes = "1.1.0"
 clap = { version = "4.1.8", features = ["derive"] }
 clap_complete = "4.5.38"
@@ -117,7 +124,7 @@ crossbeam-skiplist = "0.1.3"
 ctor = "0.6.3"
 dirs = "6.0.0"
 dotenvy = "0.15.7"
-fjall = "3.0.1"
+fjall = { version = "3.0.1", features = ["lz4", "metrics"] }
 fjall-utils.path = "crates/fjall-utils"
 hex = "0.4.3"
 hickory-resolver = "0.25.2"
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -97,7 +109,7 @@ crossbeam = "0.8"
 crossbeam-skiplist = "0.1.3"
 ctor = "0.6.3"
 dotenvy = "0.15.7"
-fjall = "3.0.1"
+fjall = { version = "3.0.1", features = ["bytes_1", "lz4", "metrics"] }
```
svix-jplatte: Why use the bytes feature? I saw their benchmarks that indicate that this is pretty bad for memory usage.

src/core/cluster/errors.rs
```diff
@@ -0,0 +1,32 @@
+use openraft::{AnyError, NodeId as RaftNodeId, StorageError, StorageIOError};
+
+macro_rules! error_helper {
```
svix-jplatte:
```suggestion
macro_rules! storage_error_io_ctor {
```

Or something. `error_helper` is very vague.

src/core/cluster/serialized_state_machine.rs
```diff
@@ -0,0 +1,286 @@
...
+
+// This file supports serializing a bunch of Fjall keyspaces to a file
+// to be sent as a Raft snapshot. In the future, when
+// https://github.com/fjall-rs/fjall/issues/52 is done, we should use that to just
+// take a backup and then tar it up, but for now we need to actually
+// separately serialize the whole thing.
+//
+// The format is as follows:
+//
+// - The magic data "COYOTE01"
```
svix-jplatte, line 18: Coyote is only a working title, right? Maybe don't put it here? Or easy enough to deprecate and change later?

src/core/cluster/state_machine.rs
```diff
@@ -0,0 +1,448 @@
...
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+struct LastSnapshot {
+    meta: SnapshotMeta<NodeId, Node>,
+    path: PathBuf,
+}
+
+#[derive(Clone)]
+pub struct Store {
+    db: Arc<Database>,
```
svix-jplatte: Is fjall's database type not already just a handle? I saw it impls Clone. It's there any point to putting it in an `Arc`?

src/core/cluster/state_machine.rs
```diff
@@ -0,0 +1,448 @@
...
+    }
+}
+
+pin_project! {
+    pub struct StoredSnapshot {
+        #[pin]
+        file: tokio::fs::File,
+        path: PathBuf,
+    }
+}
```
svix-jplatte: Pinning a `tokio::fs::File` is pointless because it implements `Unpin`. If unclear how to handle this instead, I can send a follow-up PR.

svix-jplatte:
> It also doesn't use the same fjall database for storage as the main system. 

Why? The docs recommend one db per application.

Cargo.toml
svix-jbrown: I was experimenting with it; I'm fine to turn it off.

src/core/cluster/serialized_state_machine.rs
svix-jbrown: it's a magic number. it doesn't have to have anything to do with the final name.

src/core/cluster/state_machine.rs
svix-jbrown: I was basing on the docs, which say

> In your application, you should create a single database and keep it around for as long as needed.

But reading teh code, I do see that it's just an Arc in a trenchcoat.

src/core/cluster/state_machine.rs
svix-jbrown: ok, I can switch to `Pin::new` directly; I've just always used `pin_project` before for implementing Async* traits

svix-jbrown:
> > It also doesn't use the same fjall database for storage as the main system.
> 
> Why? The docs recommend one db per application.

It's always a good idea in databases to keep your logs as isolated as possible from your main data store to avoid priority inversions. I also think there's a good chance we may want to use something simpler for logs in the future (LSMs are kind of a silly choice for log-structured data), and having it in a separate database makes it easier to rip out.

src/core/cluster/serialized_state_machine.rs
svix-jplatte: I know but it just feels weird :sweat_smile: 
Probably just ignore me.

src/core/cluster/logs.rs
```diff
@@ -17,6 +17,8 @@ use super::raft::TypeConfig;
 
 type NodeId = <TypeConfig as RaftTypeConfig>::NodeId;
 
+/// This is an implementation of an openraft Logs store backed by fjall
+
 #[derive(Debug, PartialEq, Clone)]
 enum LogError {
```
svix-jplatte: It doesn't matter that there's a newline, this doc comment applies to `enum LogError`. Was this meant to be a non-doc comment?

Cargo.toml
```diff
@@ -131,41 +75,46 @@ jiff = { version = "0.2.18", features = ["serde"] }
...
+openraft = { version = "0.9.21", features = ["tracing-log", "anyhow", "serde", "storage-v2"] }
 opentelemetry = { version = "0.31.0", features = ["metrics"] }
 opentelemetry-http = "0.31.0"
 opentelemetry-otlp = { version = "0.31.0", default-features = false, features = ["internal-logs", "metrics", "grpc-tonic", "http-proto", "reqwest-client"] }
 opentelemetry-proto = { version = "0.31.0", default-features = false, features = ["internal-logs", "gen-tonic", "trace", "with-serde"] }
 opentelemetry-semantic-conventions = "0.31.0"
 opentelemetry_sdk = { version = "0.31.0", features = ["metrics", "rt-tokio", "rt-tokio-current-thread", "experimental_metrics_periodicreader_with_async_runtime", "experimental_trace_batch_span_processor_with_async_runtime"] }
+paste = "1.0.15"
 percent-encoding = "2.3.2"
+pin-project-lite = "0.2.16"
```
svix-jplatte: Not needed now, right?

svix-jplatte (approved):
Feel free to merge with final pieces of review feedback addressed (I don't think we have any branch protection here).

src/core/cluster/logs.rs
svix-jbrown: sure was

PR #16, opened 2026-01-23
Add stream.append RPC call

svix-gabriel:
This just adds the RPC call for writing messages to the stream. That's basically it.

Consuming messages is coming in the next commit.

The only thing really noteworthy here is how we store messages. We key on `(stream_id, msg_id)`, and the msg_id effectively an autoincrementing integer. The keys are deliberately serialized so that msgs are clustered by stream_id, and ordered by msg_id.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,0 +1,1 @@
+cargo test --workspace $@
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3444,6 +3444,7 @@ dependencies = [
  "schemars_derive 1.2.0",
  "serde",
  "serde_json",
+ "uuid",
 ]
 
 [[package]]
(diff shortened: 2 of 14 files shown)


test.sh
```diff
@@ -0,0 +1 @@
+cargo test --workspace $@
```
svix-gabriel, line 1: I'm sure someone will change this to use cargo-nextest at some point 😛

svix-jplatte: Also not a fan of bash files for dev commands. Any objections against me introducing a justfile in this repo?

svix-james: Having a `justfile` would be awesome

svix-gabriel: I'm not against a justfile, though I've never really seen the benefit over standardized script like this either.

svix-jplatte: #24

svix-gabriel:
Since there were no comments other than the testfile, I'm gonna go ahead and merge (I have other changes on top of this one I want to get out).

We can delete/replace the .sh files separately

PR #17, opened 2026-01-23
Initial implementation for Coyote Rate Limiter

svix-lucho:
Remove the existing RateLimiter implementation in favor of one based on fjall. 

This initial implementation has 2 rate limiting configurations: `FixedWindow` and `TokenBucket`. The functionality is quite straightforward. Only implemented simple versions of 2 common rate limiting algorithms, just to get us started somewhere. It should be relatively simple to add more.

```
pub struct FixedWindow {
    /// Window size
    pub size: Duration,
    /// Max tokens allowed per window
    pub tokens: u64,
}
```

```
pub struct TokenBucket {
    /// Token refill rate in tokens per refill interval
    pub refill_rate: u64,
    /// Token refill interval
    pub refill_interval: Duration,
    /// Max tokens allowed in the bucket
    pub bucket_size: u64,
}
```

Some notes:
- The rate limit config/algorithm is sent along in every `limit` call. This makes the API much nicer to use, since there is no pre-setup required. 
- If the same identifier is used with 2 different rate-limiting algorithms, the API will consider them as 2 separate resources. That's a bit weird, but I think it's an acceptable tradeoff to make the API nicer (the previous bullet point)
- I have no idea how performant this is, specially with the consensus overhead that will be on top of this API. 

**Still TBD:** evictions/expirations. I didn't use the KV/caching modules as the 'backend' for rate limiting because I think the eviction/expiration logic might end up being quite different. In the rate-limiter, we don't need to expire (delete) entries only to reinsert them again. We can keep entries in the store for longer since they can be reset when `limit` is called without having a background worker aggressively cleaning them up. I'll go deeper on this on my v2 of the rate-limiting RRC. In any case, if we end up using the KV as the backend for this, it should be very easy to change this, the logic is exactly the same.


svix-jplatte:
Wut. That was not intentional

PR #24, opened 2026-01-26
Replace lint.sh by justfile

svix-jplatte:


justfile
```diff
@@ -0,0 +1,17 @@
+fmt:
+    cargo +nightly fmt
+
+lint: fmt
```
svix-gabriel: Does this run `fmt` first, or after the other commands?

My only nit would be that `fmt` should run last in the lint step, since I've seen situations where `cargo clippy --fix` causes changes that `fmt` has to fix.

svix-gabriel (approved):

justfile
svix-jplatte: Ah, yeah I guess it would run first like this.

svix-jplatte:
I've just removed the `fmt` command following feedback - I guess everybody should have editor auto-formatting set up already, this is only there to fix up formatting issues left by `cargo clippy --fix`.

PR #25, opened 2026-01-26
Update openapi.json, generated code

svix-jplatte:
Just `cargo codegen`. I'll add a CI job.


svix-jplatte:
Looks like we have a codegen issue.

svix-jbrown (approved):

PR #30, opened 2026-01-26
Add more robust eviction for cache/kv

svix-lucho:
The caching module had a toy implementation of eviction. 

This adds (kind of) deterministic LRU eviction to the cache module. In my previous approach, eviction was implemented in the cache module directly - the KV module 'didn't know eviction existed'. I realized that my initial approach was problematic because whenever a key was expired (in KV), the LRU (in cache) kept a stale/expired entry around, which required additional cleanup on the cache worker, which was less than ideal. It can become a big issue if the cache has a lot of expirations and little evictions, which seems like a likely scenario for Coyote. 

This moves the eviction code over to the KV module, so it can keep both expirations and evictions in sync (as much as possible) and we reduce this 'stale' entries to a minimum. Unfortunately, this makes things a bit more complex and less elegant than before, since KV now does basically all of the work. 

The LRU eviction is done using `LinkedHashMap` from [hashlink](https://github.com/djc/hashlink?tab=readme-ov-file), which is essentially a combination of a doubly-linked list and a HashMap. This lets us do all the operations that we need to maintain track of the LRU items in O(1) time. I decided to bring in a dependency because it has a good API, it seems well tested and maintained, and I didn't feel like writing `unsafe` code for a non-arena-based implementation of the doubly-linked-list. We can drop it later if we feel like it. 

The LRU tracking is kept in memory, not in fjall, under the assumption that we don't need to persist the LRU, and we can just reset it when the server restarts. 

Are there better cache eviction techniques? [Yes](https://arxiv.org/abs/1512.00727). But I'm trying to keep it simple for now. I think the policy can be changed relatively easily once we have something working.


crates/kv/src/lib.rs
```diff
@@ -43,15 +53,15 @@ pub enum OperationBehavior {
...
+        Self::new("default", EvictionPolicy::NoEviction)
     }
 }
 
 const EXPIRATION_BATCH_SIZE: usize = 100; // FIXME(@svix-lucho): make this configurable?
 
 impl KvStore {
     // FIXME(@svix-lucho): receive the db from the caller
-    pub fn new(namespace: &str) -> Self {
+    pub fn new(namespace: &str, policy: EvictionPolicy) -> Self {
```
svix-james, line 64: Not related to this PR, I guess, but why is this module is using a separate Database per key-value store? Why not use a common Database, as stream is, and use separate Keyspaces per "namespace"?

svix-james: I see the FIXME, we're good.

svix-james (approved):

svix-james:
Separately, it would be _great_ to get some API tests in place for this module. :pray:

crates/kv/src/lib.rs
svix-lucho: this was resolved in https://github.com/svix/coyote-private/pull/31, James beat me to it

PR #31, opened 2026-01-27
raft: wire up the networking

svix-jbrown:
This implements a basic network driver for the server. It still doesn't actually use the raft database as backing for any of the data stores.

To be able to test this, I had to fix the places where KV/Cache/Rate-Limiter were hardcoding paths. And also fix a bunch of other stuff.

### DIFF
--- a/crates/kv/src/lib.rs
+++ b/crates/kv/src/lib.rs
@@ -51,44 +51,11 @@ pub enum OperationBehavior {
     Update,
 }
 
-impl Default for KvStore {
-    fn default() -> Self {
-        Self::new("default", EvictionPolicy::NoEviction)
-    }
-}
-
 const EXPIRATION_BATCH_SIZE: usize = 100; // FIXME(@svix-lucho): make this configurable?
 
 impl KvStore {
-    // FIXME(@svix-lucho): receive the db from the caller
-    pub fn new(namespace: &str, policy: EvictionPolicy) -> Self {
-        let db = Database::builder(format!("db/kv/{namespace}"))
-            .open()
-            .unwrap();
-
+    pub fn new(namespace: &str, db: Database, policy: EvictionPolicy) -> Self {
         let kv_keyspace = format!("_coyote_kv_{namespace}");
-
-        let tables = {
-            let opts = KeyspaceCreateOptions::default();
-            db.keyspace(&kv_keyspace, || opts).unwrap()
-        };
-
-        Self {
-            db,
-            tables,
-            policy,
-            lru: LinkedHashMap::new(),
-        }
-    }
-
-    // FIXME(@svix-lucho): remove when we receive db from the caller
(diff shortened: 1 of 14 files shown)


src/core/cluster/network.rs
```diff
@@ -0,0 +1,171 @@
...
+    ) -> Self::Network {
+        NetworkClient {
+            target,
+            node: node.clone(),
+            client: self.client.clone(),
+        }
+    }
+}
+
+pub(super) fn detect_address(cfg: &Configuration) -> anyhow::Result<SocketAddr> {
```
tasn, line 158: I have an idea that could work and doesn't depend on any external crates:
*Connect to another node in the cluster (each node should have that, right?).
* Use [local_addr](https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.local_addr) to get our local address.

This is nice because it automatically supports ip4 and ipv6, as well as ensuring we choose an IP that is routable to the other node (which won't work with the current approach).

What do we use it for btw, is it for reporting our IP to the cluster once we join?
I actually have zero idea about how node discovery is usually done, but if DNS we'll just poll it periodically (or when another node joins the cluster), what other ways are there? (as I said, I really don't know).

svix-jbrown: This is called from `initialize()`; the first node in the cluster needs to know its own IP to bootstrap the cluster.

Right now, this isn't doing any peer discovery at all (all peers are manually added by calling the `/admin/` RPCs). I agree that if we have a peer discovery mechanism (whether that's SRV records or mDNS or whatever), we can probably ditch this method, as well as ditching the admin RPCs.

crates/kv/src/lib.rs
```diff
@@ -41,21 +41,11 @@ pub enum OperationBehavior {
...
-impl Default for KvStore {
-    fn default() -> Self {
-        Self::new("default")
-    }
-}
-
 const EXPIRATION_BATCH_SIZE: usize = 100; // FIXME(@svix-lucho): make this configurable?
 
 impl KvStore {
     // FIXME(@svix-lucho): receive the db from the caller
```
svix-lucho: you can remove this comment now 😆

crates/kv/src/lib.rs
```diff
@@ -66,12 +56,7 @@ impl KvStore {
         Self { db, tables }
     }
 
-    pub fn new_temporary(namespace: &str) -> Self {
-        let db = Database::builder(format!("db/kv/{namespace}"))
-            .temporary(true)
-            .open()
-            .unwrap();
-
+    pub fn new_temporary(namespace: &str, db: Database) -> Self {
```
svix-lucho: And this can also be removed, it was only used in tests. Since you're creating a temp db in the test, we can just use `new()`

svix-lucho (approved):

PR #41, opened 2026-01-27
Implement stream.fetch-locking and stream.ack

svix-gabriel:
(Sorry for the heckin-chonker PR. It's a lot of tests, so the diff isn't as bad as it looks).

This implements `stream.fetch-locking`, which consumes messages from the Stream using "Stream semantics" (head of line blocking, no concurrent consumers form the same consumer group).

This also implements `stream.ack`, which will work for both `stream.fetch-locking`, as well as the "queue" variation for consuming messages.

How I recommend viewing this PR

1. First, I'd look at the new tests for fetch-locking and ack functionality. [here](https://github.com/svix/coyote-private/pull/41/changes#diff-03b82057a4ff27433313ee9e4ade0876d63e1d3a72a46ee1952445c82bf8e435). This should hopefully be semi-convincing that the core feature works, so you can understand the guarantees made by the new RPC calls.
2. Then, I'd look at the first commit in isolation: [Add helper methods for interacting with Leases](https://github.com/svix/coyote-private/pull/41/changes/f70314da2cb84d5c9dac43908260989fdfadd6d6). This adds the Lease table, and a bunch of helper methods for working with Leases. (These are the same Leases described in the original [Stream RFC](https://github.com/svix/rfc/pull/38/changes)). 
3. Finally, the last commit, which implements the operations and RPC calls on top of the helper methods. This commit is much more straightforward than the others.

The Lease logic here is fairly complicated, but I opted to implement it fully in this PR, because it will make the `queue` implementation much more straightforward, since queue semantics will use Leases as well.

# TODOs
- As shown in the demo, this uses `streamIds`. I'll fix this in a later PR so users can just pass in the stream name, and we don't expose the ID at all.

### DIFF
--- a/crates/stream/src/tables.rs
+++ b/crates/stream/src/tables.rs
@@ -61,7 +61,7 @@ impl TableRow for StreamRow {
 
 /// A lease represents a consumer group's "hold" on a block of messages. This is used to prevent multiple consumers in the same consumer group
 /// from touching the same messages.
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub(crate) struct LeaseRow {
     pub stream_id: StreamId,
     pub cg: ConsumerGroup,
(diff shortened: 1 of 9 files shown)


crates/stream/src/tables.rs
```diff
@@ -49,7 +55,209 @@ impl TableRow for StreamRow {
...
+    pub block_end: MsgId,
+    pub leased_at: Timestamp,
+    /// LeaseRows expire based on the visibility timeout.
+    pub expires_at: Timestamp,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    /// When the block of messages was acknowledged / processed by a consumer in the group.
+    pub acked_at: Option<Timestamp>,
+}
+
+/// For the key, we use a composit of (stream_id, consumer_group, block_end).
```
svix-james, line 80:
```suggestion
/// For the key, we use a composite of (stream_id, consumer_group, block_end).
```

I'll see myself out now.

svix-gabriel: Lol I fixed it in a later commit.

The typo checker in CI caught it

svix-james (approved):

PR #42, opened 2026-01-28
Config groups, first pass

svix-james:
First attempt at using common configuration to get the correct database
and cache/key-value/idempotency configurations, other modules to
follow later. There's not actually an API right now for creating these
configurations, so this is really just scaffolding at this point.

The key thing to note here is that the proper config group for
most of the modules right now (though this may change) comes from
pulling the "group" name out of the first chunk of the `/`-delimited
key that's used. We then find that in the database and, if it doesn't
exist (which is true for everything right now), we return a default
config. Right now this default is hard-coded. Later it will come from
something that's previously been put into the database via a config
file or some bootstrapping process.

Note that I don't love the name "configgroup", "config_group", or
really anything that's been suggested thus far, but this is what's
been mentioned in the RFC, so keeping it for now.

Also, I removed the management DB. All these databases seem like
overkill. I think we can add it back later as needed.

Tests / proper "default" config groups rather than hard-coded defaults
will follow soon after, but I really want to get some of this code
reviewed / merged before I do too much more.

### DIFF
--- a/crates/coyote-configgroup/src/lib.rs
+++ b/crates/coyote-configgroup/src/lib.rs
@@ -1,0 +1,89 @@
+use fjall::{Database, Error, KeyspaceCreateOptions};
+
+use crate::entities::{ModuleConfig, StorageType};
+use coyote_error::Result;
+
+pub mod entities;
+pub mod operations;
+mod tables;
+pub use tables::ConfigGroup;
+
+pub fn group_name(key: &str) -> Option<&str> {
+    if !key.contains('/') {
+        return None;
+    }
+    key.split("/").next()
+}
+
+#[derive(Clone)]
+pub struct State {
+    db: fjall::Database,
+    keyspace: fjall::Keyspace,
+    both_dbs: BothDatabases,
+}
+
+// Yeah this is dumb AF. I don't care
+// right now.
+#[derive(Clone)]
+pub struct BothDatabases {
+    pub ephemeral_db: fjall::Database,
+    pub persistent_db: fjall::Database,
+}
+
+impl State {
+    pub fn db(&self) -> &fjall::Database {
+        &self.db
+    }
+
+    pub fn keyspace(&self) -> &fjall::Keyspace {
+        &self.keyspace
+    }
+
+    pub fn init(both_dbs: BothDatabases) -> Result<Self, Error> {
+        const CONFIGGROUP_KEYSPACE: &str = "_coyote_cfggroup";
+
+        let db = both_dbs.persistent_db.clone();
+        let keyspace = {
+            let opts = KeyspaceCreateOptions::default();
+            db.keyspace(CONFIGGROUP_KEYSPACE, || opts)?
+        };
+
+        Ok(Self {
+            db,
+            both_dbs,
+            keyspace,
+        })
+    }
+
+    pub fn fetch_group<C: ModuleConfig>(
+        &self,
+        group_name: String,
+    ) -> Result<Option<ConfigGroup<C>>> {
+        ConfigGroup::fetch(&self.keyspace, &group_name)
+    }
+
+    pub fn fetch_all_groups<C: ModuleConfig>(
+        &self,
+    ) -> Result<impl Iterator<Item = Result<ConfigGroup<C>>>> {
+        ConfigGroup::fetch_all(&self.keyspace)
+    }
+
+    pub fn give_me_the_right_db<C: ModuleConfig>(&self, configgroup: &ConfigGroup<C>) -> Database {
+        match configgroup.storage_type {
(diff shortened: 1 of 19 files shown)


crates/coyote-configgroup/src/tables.rs
```diff
@@ -0,0 +1,53 @@
...
+pub struct ConfigGroup<C: ModuleConfig> {
+    pub id: ConfigGroupId,
+    pub name: String,
+    pub storage_type: StorageType,
+
+    pub created_at: Timestamp,
+    pub updated_at: Timestamp,
+
+    // Module-specific
+    pub config: C,
```
svix-james, line 23: I had originally gone down the path of using a flattened enum here, similar to how SinkConfig works in svix-server streams, but all the matching I was having to do even in this limited PR was getting cumbersome. But maybe there are reasons why an enum might actually be a better approach?

crates/coyote-configgroup/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+        };
+
+        Ok(Self {
+            db,
+            both_dbs,
+            keyspace,
+        })
+    }
+
+    pub fn fetch_kv_group(&self, group_name: String) -> Result<Option<KeyValueGroup>> {
```
svix-james: These can all be simplified. Will do it in a followup / sooner if no one reviews this.

crates/kv/src/lib.rs
```diff
@@ -227,33 +223,106 @@ const MAX_CAPACITY: u64 = 1024 * 8; // FIXME: make this configurable
...
-            // Eviction
-            if store.disk_space() > MAX_CAPACITY
-                && store.policy == EvictionPolicy::LeastRecentlyUsed
-            {
-                // FIXME(@svix-lucho): we can add smarter eviction instead of just doing one at a time
-                let _ = store.evict_lru(1);
+
+        timer.tick().await;
+
+        let kv_groups = match configgroup_state.fetch_all_groups() {
```
svix-james, line 251: Yep, repetitive. I'm assuming this can be simplified, but just trying to beat the merge conflicts right now.

src/lib.rs
```diff
@@ -132,6 +141,115 @@ async fn run_interserver(cfg: Configuration, state: AppState, listener: Option<T
     state.raft.shutdown().await.unwrap();
 }
 
+impl AppState {
+    // FIXME: Blocking
+    pub fn kv_store(&self, key_name: &str) -> Result<KvStore> {
```
svix-james: Yep, repetitive. I'm assuming these methods can be simplified further, but just trying to beat the merge conflicts right now.

src/lib.rs
```diff
@@ -146,50 +264,33 @@ pub async fn run_with_prefix(
...
 
     // build our application with a route
     let app_state = AppState {
         cfg: cfg.clone(),
-        kv_store,
-        cache_store,
         rate_limiter: crate::v1::modules::rate_limiter::RateLimiter::new(
             "rate_limiter_default",
             persistent_db.clone(),
         ),
```
svix-lucho, line 287: You can ignore the rate limiter for now, I'm thinking of making it use the KV like idempotency does, so it'll be a similar refactor.

src/v1/endpoints/cache.rs
```diff
@@ -119,11 +117,10 @@ async fn cache_get(
 /// Cache Delete
 #[aide_annotate(op_id = "v1.cache.delete")]
 async fn cache_del(
-    State(AppState {
-        mut cache_store, ..
-    }): State<AppState>,
+    State(state): State<AppState>,
     ValidatedJson(data): ValidatedJson<CacheDeleteIn>,
 ) -> Result<Json<CacheDeleteOut>> {
+    let mut cache_store = state.cache_store_by_key(&data.key.0)?;
```
svix-lucho, line 123: This can come from an extractor right? And same for other places. Because I feel like it can be easy to miss or get wrong. 

(not for this PR, asking for later)

svix-james: Yeah I agree but I'm not sure how easy it is. Needing to parse the group name out of the key (which is one reason I don't like embedding the group name this way) could be problematic, since it means we need to parse the body. Someone better w/ Axum knowledge may be able to chime in here. 

But either way, future work potentially.

svix-lucho: Hmm maybe it'll be hard to generalize for all APIs

svix-lucho (approved):

src/v1/endpoints/cache.rs
svix-james: TBH I'm hoping we find a different way to go here entirely long-term.

PR #44, opened 2026-01-28
Add coyote-runtime for "shared nothing" worker tasks

svix-gabriel:
This adds a `coyote-runtime` crate, with a `WorkerPool` for executing tasks. This is to support the architecture outlined in https://github.com/svix/rfc/pull/28.

The WorkerPool is just a standalone library for now. The next PR will integrate it with the actual Coyote server and RPC calls.

## Background

One of the design decisions in Coyote is for operations on the same "resource" to be isolated to a single thread. e.g.  Reads and writes to `Stream A` always executing on the same thread. This makes designing these features much simpler, since we don't need to worry about locking or other concurrency/parallelization/transaction issues when doing a bunch of writes for a single resource.

Another goal is to make sure that we don't accidentally make blocking `fjall` calls on our `async` `axum` server. To pull this off, we need to isolate all of the `fjall` resources away from the axum server.

## WorkerPool design

The idea is that we'll initialize the worker pool as such
```rust
// our beloved shared state with all the `fjall::Database`/`fjall::Keyspace`/etc handles, maybe in-memory caches
// down the line, etc.
let app_state = AppState { ... } 
let pool = WorkerPool::spawn(app_state, num_threads).expect("initializing threadpool");
```

The `app_state` will be cloned and injected onto each thread in the pool.

Then, whenever we want to execute work that needs to interact with `fjall`/`app_state`, it can be in one of two ways.

```rust

let operation = move |app_state: &mut AppState| {
     app_state.db.insert(/* blah blah blah */);
     Ok(())
};

let result = pool
      .run_on_owned_thread("resource_key",  operation)
      .await
      .unwrap();
```
This would execute the `operation` closure on a specific thread, and guarantee that any operation that also uses `"resource_key"` will always run on that same thread.

Example: if `"resource_key"` was a specific Stream identifier, this would ensure that ALL operations for the same stream go to the same thread. This avoids the concurrency issues from before, and also lets us potentially take advantage of thread_local state, lockless caching, etc.

This would effectively replace `tokio::spawn_blocking(operation).await` that we currently have in our axum handlers.

The other way to interact with the threadpool is to run a closure on _any arbitrary thread_.

```rust
let operation = move |app_state: &mut AppState| {
     // A read only operation, that potentially doesn't need to happen on a specific thread
     app_state.db.get(/* blah blah blah */);
     Ok(())
};

let result = pool
      .run_on_any_thread(operation)
      .await
      .unwrap();
```

Some operations may not need to always run on a particular thread, so we can just schedule the work on the soonest available thread.

## Other Features
- The worker pool should gracefully handle panics, so they don't crash the thread.
- Right now, the worker pool will wait for all threads to finish on `Drop`. I think this is the right call, as when all the handles to the WorkerPool are dropped when receiving the shutdown signal, it'll make sure any outstanding work on the threadpool is finished. But maybe there's more granular shutdown signal handling we need here

## What's missing

- "thread per core". This PR doesn't both with trying to bind threads to cores. We can do that later.
- Wiring this up with the axum server. That'll be a followup PR.


svix-gabriel:
Moving this back to draft after discussions w.r.t. raft/consensus stuff.

PR #45, opened 2026-01-29
raft: wire up kv::set to go through consensus

svix-jbrown:
This is a sketch of what it would look like to map high-level operations into the consensus layer.

Obviously it's extremely verbose; I think it would be quite reasonable to write a macro to generate all of the code in `crates/kv/src/operations/mod.rs` and much of the code in `crates/kv/src/operations/set.rs` and `crates/kv/src/operations/delete.rs`. I'm envisioning something like

```rust
// complete contents of mod.rs

mod set;
mod delete;

impl_module_operations!(
  Set => set::SetOperation + set::SetResponse,
  Delete => delete::DeleteOperation + delete::DeleteResponse
)
```

I have that part mostly built and would plan to put it in the `coyote-operations` trait so we can reuse it across all modules.

I also think we can make a macro for the stuff in `handle.rs`. I had initially hoped to use `enum_dispatch` for it, but it doesn't support associated types. I still think we can have something like

```rust
impl_handles!(
  Kv => coyote_kv::operations::KvRequest + coyote_kv::Operations::KvResponse
)
```

and have that generate  the Request type, the Response type, and the various convenience methods on the handle object.

This is not landable as-is for several reasons, including that it breaks rate-limiting and cache (which... internally do cross-module writes, which will be annoying to implement).

Like I said, mostly just a sketch. This does actually work, though.

### DIFF
--- a/crates/coyote-error/src/lib.rs
+++ b/crates/coyote-error/src/lib.rs
@@ -67,6 +67,11 @@ impl Error {
     }
 
     #[track_caller]
+    pub fn operation(code: StatusCode, detail: Option<String>) -> Self {
+        Self::new(ErrorType::Operation { code, detail })
+    }
+
+    #[track_caller]
     pub fn trace(mut self) -> Self {
         self.trace.push(Location::caller());
         self
@@ -93,6 +98,11 @@ impl IntoResponse for Error {
                 tracing::debug!("{:?}, location: {:?}", &s, stringified);
                 s.into_response()
             }
+            ErrorType::Operation {
+                code,
+                detail: Some(detail),
+            } => (code, detail).into_response(),
+            ErrorType::Operation { code, detail: _ } => code.into_response(),
             s => {
                 tracing::error!("type: {:?}, location: {:?}", s, stringified);
                 (StatusCode::INTERNAL_SERVER_ERROR, Json(json!({}))).into_response()
@@ -145,6 +155,11 @@ pub enum ErrorType {
     Http(HttpError),
     /// Timeout error
     Timeout(String),
+    /// An error from an Operation application
+    Operation {
+        code: StatusCode,
+        detail: Option<String>,
+    },
 }
 
 impl fmt::Display for ErrorType {
@@ -152,6 +167,13 @@ impl fmt::Display for ErrorType {
         match self {
             Self::Generic(s) | Self::Validation(s) | Self::Timeout(s) => s.fmt(f),
             Self::Http(s) => s.fmt(f),
+            Self::Operation { detail, code } => {
+                if let Some(detail) = detail {
+                    detail.fmt(f)
+                } else {
+                    write!(f, "code {code}")
+                }
+            }
         }
     }
 }
--- a/crates/kv/src/lib.rs
+++ b/crates/kv/src/lib.rs
@@ -134,12 +134,7 @@ impl KvStore {
         Ok(())
     }
 
-    pub fn set_operation(
-        &self,
-        key: String,
-        model: KvModel,
-        behavior: OperationBehavior,
-    ) -> SetOperation {
+    pub fn set_operation(key: String, model: KvModel, behavior: OperationBehavior) -> SetOperation {
         SetOperation::new(key, model, behavior)
     }
 
--- a/crates/kv/src/lib.rs
+++ b/crates/kv/src/lib.rs
@@ -14,9 +14,14 @@ use jiff::Timestamp;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
-use crate::tables::{ExpirationRow, KvPairRow};
+pub mod operations;
 
-#[derive(Debug, Deserialize, Eq, PartialEq, Serialize)]
+use crate::{
+    operations::SetOperation,
+    tables::{ExpirationRow, KvPairRow},
+};
+
+#[derive(Debug, Deserialize, Eq, PartialEq, Serialize, Clone)]
 pub struct KvModel {
     pub expires_at: Option<Timestamp>,
     pub value: Vec<u8>,
@@ -129,7 +134,22 @@ impl KvStore {
         Ok(())
     }
 
+    pub fn set_operation(
+        &self,
+        key: String,
+        model: KvModel,
+        behavior: OperationBehavior,
+    ) -> SetOperation {
+        SetOperation::new(key, model, behavior)
+    }
+
     pub fn set(&mut self, key: &str, model: &KvModel, behavior: OperationBehavior) -> Result<()> {
+        // TODO: remove this method
+        tracing::error!("unsafe method KvStore::set called!");
+        self.set_(key, model, behavior)
+    }
+
+    fn set_(&mut self, key: &str, model: &KvModel, behavior: OperationBehavior) -> Result<()> {
         match behavior {
             OperationBehavior::Upsert => {
                 self.insert_with_expiration(key, model)?;
--- a/justfile
+++ b/justfile
@@ -1,7 +1,7 @@
 lint:
+    cargo +nightly clippy --workspace --fix --allow-dirty --all-features --all-targets
     cargo +nightly fmt
     cargo sort --workspace -o package,lib,bin,features,dependencies,dev-dependencies,lints,workspace
-    cargo +nightly clippy --workspace --fix --allow-dirty --all-features --all-targets
 
 # Test the backend
 test *args='':
--- a/justfile
+++ b/justfile
@@ -1,7 +1,7 @@
 lint:
-    cargo +nightly clippy --workspace --fix --allow-dirty --all-features --all-targets
     cargo +nightly fmt
     cargo sort --workspace -o package,lib,bin,features,dependencies,dev-dependencies,lints,workspace
+    cargo +nightly clippy --workspace --fix --allow-dirty --all-features --all-targets
 
 # Test the backend
 test *args='':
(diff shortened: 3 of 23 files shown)


src/v1/endpoints/kv.rs
```diff
@@ -111,17 +111,17 @@ pub struct KvDeleteOut {
...
-    let key = data.key.clone();
+    let key = data.key.0.clone();
     let behavior = data.behavior.clone();
     let model = data.into_model();
 
-    kv_store
-        .set(&key.0, &model, behavior)
-        .map_err(|e| crate::error::Error::generic(e))?;
+    let operation = kv_store.set_operation(key, model, behavior);
+    repl.client_write_kv(operation).await.0?;
```
svix-lucho:
```suggestion
    let key = data.key.0.clone();
    let behavior = data.behavior.clone();
    let model = data.into_model();

    let operation = coyote_kv::operations::SetOperation::new(key, model, behavior);
    repl.client_write_kv(operation).await.0?;
```

you can do this here right? So we're not 'using' the config groups before consensus.

crates/kv/src/lib.rs
```diff
@@ -121,7 +126,22 @@ impl KvStore {
...
     }
 
+    pub fn set_operation(
+        &self,
+        key: String,
+        model: KvModel,
+        behavior: OperationBehavior,
+    ) -> SetOperation {
+        SetOperation::new(key, model, behavior)
+    }
```
svix-lucho: why do we need this?

src/core/cluster/state_machine.rs
```diff
@@ -363,18 +369,21 @@ impl RaftStateMachine<TypeConfig> for Store {
             match item.payload {
                 EntryPayload::Blank => {
                     tracing::trace!("heartbeat");
-                    replies.push(raft::Response::blank())
+                    replies.push(Response::Blank)
                 }
                 EntryPayload::Normal(req) => {
                     // TODO actually apply something
```
svix-lucho:
```suggestion
```

src/v1/endpoints/kv.rs
svix-james: I'm not sure I understand the suggestion here. Don't you need to know, e.g., the correct db, which comes from the config group, before sending the request to Raft?

svix-lucho: IIUC that happens in [`applier.rs`](https://github.com/svix/coyote-private/blob/3e554696579e3c7cab6272b473b5a7e3ff7e6e97/src/core/cluster/applier.rs) (after consensus).

```rust
pub(super) fn apply_request(request: Request, state: &AppState) -> anyhow::Result<Response> {
    Ok(match request {
        Request::Kv(req) => {
            // TODO: this shouldn't be mut but KvStore currently requires it
            let mut store = state.kv_store_by_key(req.key_name())?;
            Response::Kv(req.apply(&mut store))
        }
    })
}
```

Why does it need to happen before?

svix-jbrown: I don't think it does for KV, but I think in general modules should probably be responsible for creating their own operations (as a point in which they can do validation, etc)

src/core/cluster/state_machine.rs
svix-jbrown: heh

src/v1/endpoints/kv.rs
svix-lucho: Yeah that's fair. I just mean that it should be an associated function that doesn't take in `self`, since it doesn't depend on it

svix-jbrown: ok, I can do that

justfile
```diff
@@ -1,7 +1,7 @@
 lint:
-    cargo +nightly clippy --workspace --fix --allow-dirty --all-features --all-targets
     cargo +nightly fmt
     cargo sort --workspace -o package,lib,bin,features,dependencies,dev-dependencies,lints,workspace
+    cargo +nightly clippy --workspace --fix --allow-dirty --all-features --all-targets
```
svix-gabriel: Unrelated to the main change, but we should be running `clippy --fix` BEFORE `fmt`, since clippy can sometimes produce changes that fmt then fixes, but not the other way around

src/core/cluster/handle.rs
```diff
@@ -0,0 +1,35 @@
+use super::raft::{NodeId, Raft};
+use coyote_kv::operations::KvRequest;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub enum Request {
+    Kv(coyote_kv::operations::Operation),
```
svix-gabriel, line 31: Is this where we're doing the enum dispatch you talked about?

So I guess the idea is for each operation, we'll add new variants here (to both Request and Response enums?)

src/core/cluster/handle.rs
```diff
@@ -0,0 +1,35 @@
...
+pub struct RaftState {
+    pub raft: Raft,
+    pub node_id: NodeId,
+}
+
+impl RaftState {
+    pub async fn client_write_kv<O: KvRequest>(&self, op: O) -> O::Response {
+        let request = Request::Kv(op.into());
+        // TODO: don't unwrap here!
+        let response = self.raft.client_write(request).await.unwrap();
```
svix-gabriel: I know you marked them with TODO's, but why are we starting with panics here? Seems like using regular Result<...> types here would be simpler

src/core/cluster/handle.rs
```diff
@@ -0,0 +1,35 @@
...
+}
+
+#[derive(Clone)]
+pub struct RaftState {
+    pub raft: Raft,
+    pub node_id: NodeId,
+}
+
+impl RaftState {
+    pub async fn client_write_kv<O: KvRequest>(&self, op: O) -> O::Response {
```
svix-gabriel: Why have a distinct method for each operation type? Why not just a generalized `client_write_op(&self, op: impl Into<Request>)` (or something similar)

justfile
svix-jbrown: The downside is that clippy takes several minutes to run, and `cargo fmt` takes <1 second to run

src/core/cluster/handle.rs
svix-jbrown: Yep

src/core/cluster/handle.rs
svix-jbrown: ah, when I wrote that, this method had more code, and I didn't want to make a new Result type to unify them all. I guess now there's only one failable path, so it's easy enough.

justfile
svix-gabriel: I don't really see why that's beneficial since `just lint` runs all the commands anyway.

My preference would be keep `clippy` before `fmt`. The benefit of running `just lint` is that you can be pretty mindless about it. Run it once, when you're done, and you can be sure that as long as there's no warnings/errors, you're PR will pass linting in CI.

Changing the order breaks that guarantee.

src/core/cluster/handle.rs
svix-jbrown: OK, I did it. It adds some very gross type bounds, but does work!

src/core/cluster/handle.rs
```diff
@@ -0,0 +1,86 @@
...
+        let Ok(module_response) =
+            <<O as OperationRequest>::Response as OperationResponse>::ResponseParent::try_from(
+                response.data,
+            )
+        else {
+            panic!("failed to deserialize module response");
+        };
+        let Ok(resp) = module_response.try_into() else {
+            panic!("failed to deserialize module response");
+        };
```
svix-jbrown: these aren't just unwraps because then I would need to add an extra `<<<O as OperationRequest>::Response as OperationResponse>::ResponseParent as TryFrom<Response>>::Error: Debug` constraint on this method and that makes me too sad

crates/coyote-error/src/lib.rs
```diff
@@ -93,6 +98,11 @@ impl IntoResponse for Error {
                 tracing::debug!("{:?}, location: {:?}", &s, stringified);
                 s.into_response()
             }
+            ErrorType::Operation {
+                code,
+                detail: Some(detail),
+            } => (code, detail).into_response(),
```
svix-jbrown, line 104: obviously this should be changed at some point to format the detail (ideally using `MsgpackOrJson`), but then we have some cyclic dependency issues.

I wonder if we should move the `IntoResponse` impl for `coyote_error` into the server module so that it can format the responses as something other than plaintext?

src/core/cluster/app.rs
```diff
@@ -70,7 +70,7 @@ where
 
 #[tracing::instrument(skip_all)]
 async fn append_entries(
-    State(state): State<AppState>,
```
svix-jbrown, line 73: all of this noise in every file is because the Raft object itself needs to have a handle on the `AppState` to do the application, so it had to be moved out of the AppState. I put it in an Extension; the other option would be to put it in the `State` and have some `SuperAppState` that contains both the current AppState and the Raft module, but that seems excessive?

src/core/cluster/handle.rs
```diff
@@ -0,0 +1,88 @@
...
+pub struct RaftState {
+    pub raft: Raft,
+    pub node_id: NodeId,
+}
+
+impl RaftState {
+    /// Write a single operation into the Raft log and return its response.
+    pub async fn client_write<O>(&self, op: O) -> WriteResult<O::Response>
+    where
+        O: Into<Request> + OperationRequest,
```
svix-gabriel, line 67: "Generic type annotations can't hurt you"

svix-gabriel (approved):

PR #49, opened 2026-01-29
Implement queue semantics with Stream

svix-gabriel:
This adds the `stream.fetch` RPC call, which is equivalent to "Queue semantics" for Stream.

When we say "queue semantics", we mean:
- multiple consumers in the same group can make requests concurrently
- Instead of limiting a single consumer, with head of line blocking, each consumer simply grabs messages "in order", and those messages are locked (for the duration of the visibility timeout).

This was pretty easy to do using the Lease helper APIs added in a prior PR.

TODOs

- There's no locking logic in place to actually protect against parallel fetch calls. We were gonna rely on the "shared nothing" architecture, but that's been put on hold until all of the raft/consensus algorithm stuff is sorted out

- no DLQing yet. That's the next PR

- I imagine we'll hash out API details, but the core functionality for supporting queue semantics is all here.

### DIFF
--- a/crates/stream/src/operations/fetch.rs
+++ b/crates/stream/src/operations/fetch.rs
@@ -1,0 +1,75 @@
+use std::num::NonZeroU16;
+
+use crate::{
+    State,
+    entities::{ConsumerGroup, MsgId, MsgOut, StreamName},
+    tables::{LeaseDiff, LeaseRow, MsgRow, NameToStreamRow},
+};
+use coyote_error::Result;
+use jiff::Timestamp;
+
+pub struct Fetch {
+    lease_diff: LeaseDiff,
+    msgs: Vec<(MsgId, MsgRow)>,
+}
+
+pub struct FetchOutput {
+    pub msgs: Vec<MsgOut>,
+}
+
+impl Fetch {
+    pub fn new(
+        state: &State,
+        name: StreamName,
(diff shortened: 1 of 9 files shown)


crates/stream/src/operations/fetch.rs
```diff
@@ -0,0 +1,75 @@
...
+
+        // Unlike FetchLocking, we don't block on active leases.
+        // Instead, we exclude both acked and active leases from the fetch.
+        let blocked_leases = leases
+            .iter()
+            .filter(|lease| lease.acked_at.is_some() || lease.is_active(now));
+
+        let msgs = MsgRow::fetch_available(state, stream_id, blocked_leases, batch_size.into())?;
+
+        let mut lease_diff = LeaseRow::cull_and_compact(leases, now);
```
svix-james, line 40: I guess there's not too much point in worrying about this too much right now, but scanning all leases constantly and calling `cull_and_compact` on them when both acking and fetching seems a bit scary.

svix-james (approved):

crates/stream/src/operations/fetch.rs
svix-gabriel: I don't think it'll be too bad. Even w/something like 20 concurrent consumers, that's ~20 leases at a given time.

But definitely something we'll keep an eye on

PR #50, opened 2026-01-30
use a nextest profile to ignore the openraft test

svix-jbrown:
I find this to be slightly more discoverable than `#[ignore]` on the test.


svix-jplatte (approved):
nice

PR #55, opened 2026-01-31
discovery for coyote peers

svix-jbrown:
This adds some basic discovery for cluster bootup, based on a configured list of seed nodes.

When a new node comes up, it reaches out to all of the seed nodes. If there is a single cluster containing some subset of them, it attempts to join that cluster. If none of them are in a cluster, it attempts to initialize a new cluster. This is somewhat racy, because if you launch several nodes at the same time they all might try to initialize and you'd end up with 3 one-node clusters; I'll have work on improving that, but I wanted to get //something// out.

This also changes the config system. I like the idea of using a nested config (where cluster config is in its own Table), but this doesn't play super-well with the `config` module. I reconfigured it to support nested config for environment variables (separated by the `:` character), which required fixing up the `dotenvy` crate to also support such keys. The SHA here is https://github.com/allan2/dotenvy/pull/161 rebased on top 0.15.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,6 +1,11 @@
 [profile.default]
 default-filter = 'not test(~openraft_slow)'
+slow-timeout = { period = "10s", terminate-after = 4 }
 
 [profile.ci]
 fail-fast = false
 default-filter = 'all()'
+
+[[profile.ci.overrides]]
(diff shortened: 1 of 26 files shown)


svix-james (approved):
Ok

PR #56, opened 2026-02-02
Support MsgPack for public API request bodies

### LINKED ISSUE #52: Support msgpack in public API request bodies

svix-jplatte:
Closes https://github.com/svix/coyote-private/issues/52.

### DIFF
--- a/crates/coyote-proto/src/lib.rs
+++ b/crates/coyote-proto/src/lib.rs
@@ -1,4 +1,4 @@
 mod msgpack;
-mod validated_json;
+mod msgpack_or_json;
 
-pub use self::{msgpack::MsgPack, validated_json::ValidatedJson};
+pub use self::{msgpack::MsgPack, msgpack_or_json::MsgPackOrJson};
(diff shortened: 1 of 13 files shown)


crates/coyote-proto/src/lib.rs
```diff
@@ -1,4 +1,4 @@
 mod msgpack;
-mod validated_json;
+mod msgpack_or_json;
```
tasn, line 2: A small nitpick, but now is probably the time to do it, can we just rename the models to `ApiRequest` and `ApiResponse` and rename the module to something better?
Even if we keep it just msgpack and json forever, it feels like it's nicer? Especially since the request part also does validation (I assume, or did you remove that part entirely? If you have, we probably want some validation?).

svix-jbrown (approved):

crates/coyote-proto/src/lib.rs
svix-jplatte: I considered it, but this only for the public API, and will likely be used for both request and response bodies. So I suppose I could name it `PublicApiBody`, but it feels a bit weird / not in line with axum conventions.

Re. validation, that's still there. It's just a bit weird to have it in the name if this type is also going to be used for response bodies.

PR #59, opened 2026-02-02
Support MsgPack for public API response bodies

### LINKED ISSUE #53: Support msgpack in public API response bodies
Requires some task-local magic to be able to access (information derived from) the `Accept` header value in the new `IntoResponse` implementation we'll use.

svix-jplatte:
Closes https://github.com/svix/coyote-private/issues/53.

### DIFF
--- a/crates/coyote-proto/src/msgpack_or_json.rs
+++ b/crates/coyote-proto/src/msgpack_or_json.rs
@@ -1,3 +1,5 @@
+#![expect(clippy::disallowed_types)] // FIXME: currently triggers due to input_with / output_with
+
 use aide::OperationIo;
 use axum::{
     RequestExt as _,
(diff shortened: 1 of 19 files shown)


crates/coyote-proto/src/msgpack_or_json.rs
```diff
@@ -1,22 +1,56 @@
...
+            .any(|accept| accept == mime::APPLICATION_MSGPACK.as_ref().as_bytes())
+    {
+        SupportedContentType::MsgPack
+    } else {
+        // Default to JSON if nothing is specified
+        SupportedContentType::Json
+    };
+
+    REQ_ACCEPT.scope(accept_value, next.run(request)).await
+}
```
svix-jbrown: something to consider adding here would be to also look at the `Content-Type` header; I've build some applications where if you don't pass `Accept`, they fall back to the `Content-Type` of your request and that makes it easier for people to do ad-hoc testing

src/core/cluster/app.rs
```diff
@@ -1,16 +1,17 @@
 use std::collections::{BTreeMap, BTreeSet};
 
 use axum::{
-    Json,
+    // Json,
```
svix-jbrown: why the comment out instead of just delete?

svix-jplatte: Development leftover.

PR #61, opened 2026-02-02
KV: some API improvements and more tests

svix-lucho:
- Use `Vec<u8>` instead of `String` for values
- Make `expire_in` an `Option` instead of using `0` as the value when there's no expiration. 
- Add a bunch of e2e tests

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -1241,7 +1241,13 @@
             "pattern": "^[a-zA-Z0-9\\-_.]+$"
           },
           "value": {
-            "type": "string"
+            "type": "array",
+            "items": {
+              "type": "integer",
+              "format": "uint8",
+              "maximum": 255,
+              "minimum": 0
+            }
           }
         },
         "required": [
@@ -1260,7 +1266,8 @@
             "description": "Time to live in milliseconds",
             "type": "integer",
             "format": "uint64",
-            "minimum": 0
+            "minimum": 1,
+            "nullable": true
           },
           "key": {
             "type": "string",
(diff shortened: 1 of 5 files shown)


tests/it/kv.rs
```diff
@@ -1,12 +1,13 @@
...
+use jiff::Timestamp;
 use serde_json::{Value, json};
 use test_utils::{StatusCode, TestClient, TestResult};
 
 async fn kv_set(
     client: &TestClient,
     key: &str,
-    expire_in: u64,
+    expire_in: Option<u64>,
     value: &str,
```
svix-james, line 11: Maybe `value` here should just be a byte slice?

svix-james (approved):

openapi.json
```diff
@@ -1241,7 +1241,13 @@
...
           },
           "value": {
-            "type": "string"
+            "type": "array",
+            "items": {
+              "type": "integer",
+              "format": "uint8",
+              "maximum": 255,
+              "minimum": 0
+            }
```
svix-jplatte, line 1250: This is pretty awkward. I think we should accept strings over JSON. Not sure if base64 or escapes for non-UTF8 is more sensible though.

svix-jplatte: I guess we can keep it like this for this PR but definitely need a follow-up ticket.

tests/it/kv.rs
svix-lucho: It looks cleaner in the tests if this utility method just accepts a string and then we convert it to bytes

openapi.json
svix-lucho: Yeah, it does feel a bit awkward. I'll create a follow-up ticket, since the stream module needs the same thing.

svix-lucho: https://github.com/svix/monorepo-private/issues/12523

PR #62, opened 2026-02-02
Add DLQ functionality to Stream

svix-gabriel:
This implements DLQ functionality for Stream. It works with both "stream" and "queue" semantics.

This is
- The ability to DLQ specific messages for a `(stream, consumer_group)`
- The ability to redrive the entire DLQ for a `(stream, consumer_group)` so those messages are visible again.

Rather than using a separate DLQ table like originally planned, it was simpler to just track DLQ'd messages with more Leases. Maybe we'll change this approach in the future 🤷.

I recommend reviewing commit https://github.com/svix/coyote-private/pull/62/changes/b650bf05df27f9c1304e28879fd2437fad38fc98 first, as it's really the bulk of the changes.

### DIFF
--- a/tests/it/stream.rs
+++ b/tests/it/stream.rs
@@ -783,3 +783,341 @@ async fn queue_fetch_single_ack() -> TestResult {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn queue_dlq_and_redrive() -> TestResult {
+    let TestContext {
+        client,
+        handle: _handle,
+        ..
+    } = super::start_server().await;
+
+    let _stream = client
+        .post("stream/create")
+        .json(json!({
+            "name": "test-stream",
+            "maxByteSize": 1024,
+            "retentionPeriodSeconds": 9999
+        }))
+        .await?
+        .expect(StatusCode::OK)
+        .json();
+
+    client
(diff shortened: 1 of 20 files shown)


tests/it/stream.rs
```diff
@@ -783,3 +783,244 @@ async fn queue_fetch_single_ack() -> TestResult {
...
+        .post("stream/dlq")
+        .json(json!({
+            "name": "test-stream",
+            "consumerGroup": "test-group",
+            "msgId": 1
+        }))
+        .await?
+        .expect(StatusCode::OK);
+
+    // Ack messages A and C (ids 0 and 2)
```
svix-lucho, line 984: what happens when acking a dlq'ed message? I think we should have a test for that

svix-gabriel: Added a test.

The behavior (which I think is correct) is that an ACK'd message gets removed from the DLQ as well, so it isn't redriven (i.e. if the message was successfully processed, why DLQ it?)

svix-lucho: Yeah I agree. It's different from how SQS works I think, I prefer your behavior.

svix-lucho (approved):

PR #85, opened 2026-02-04
Idempotency: use vec<u8> for response types, and add e2e tests

svix-lucho:
Also adds idempotency to the groups bootstrap process

### DIFF
--- a/src/bootstrap.rs
+++ b/src/bootstrap.rs
@@ -6,7 +6,8 @@ use config::ConfigBuilder;
 use coyote_configgroup::{
     BothDatabases,
     entities::{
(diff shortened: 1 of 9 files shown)


src/bootstrap.rs
```diff
@@ -177,6 +190,15 @@ pub fn run(bootstrap_cfg_path: Option<&str>, app_config: AppConfig) {
         }
     }
 
+    if let Some(idempotency) = bootstrap.idempotency {
```
svix-james, line 193: 🤩

svix-james (approved):

PR #87, opened 2026-02-04
Rate limiter: use tagged enum in API

svix-lucho:
This lets us support different rate limiter algorithms in the API


tests/it/server/mod.rs
```diff
@@ -137,6 +148,7 @@ pub struct TestContext {
     pub cfg: Arc<ConfigurationInner>,
     pub handle: IsolatedServerHandle,
     pub token: String,
+    pub sdk_client: CoyoteClient,
```
svix-james: Why use this instead of the web client?

It's fine w/ me, but I assume there was a reason why we otherwise wanted to use a raw web client?

src/v1/endpoints/rate_limiter.rs
```diff
@@ -15,10 +15,66 @@ use validator::Validate;
...
+    pub window_size_seconds: u64,
+
+    /// Maximum number of requests allowed within the window
+    #[validate(range(min = 1))]
+    pub max_requests: u64,
+}
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Validate, JsonSchema)]
-pub struct RateLimiterConfig {
+pub struct RateLimiterTokenBucketConfig {
```
svix-james, line 77: Is this data that should potentially go into a config group / have defaults associated w/ it? I'm not really clear how / if config-groups are relevant for rate-limiting.

tests/it/server/mod.rs
svix-jplatte: Yeah I think we shouldn't , as discussed on Slack recently.

src/v1/endpoints/rate_limiter.rs
svix-lucho: In the way we designed the rate limiter API (up until now), a config group is not necessary, because the 'config' is sent on every rate-limit call. I think we can ignore them for now.

tests/it/server/mod.rs
svix-lucho: I just wanted a sanity check to test that the tagged enums worked as expected **somewhere**, since we had issues with them before.

I can remove this, but I think we need this sort of test somewhere else.

svix-jplatte: I see what you mean. I think the SDK itself can have a serialization test against the generated model, using plain serde (w/o making an API request).

svix-lucho: removed this part

svix-james (approved):

PR #89, opened 2026-02-04
Add builder for test server

svix-james:
I need to use this for benchmarking so would like
to organize this a bit better upfront.


tests/it/server/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+/// TestContext without a running server. Since there's no background task for a server,
+/// you need to ensure to keep this context object alive for your whole test to prevent
+/// the workdir from being dropped and cleaned up
+pub(crate) struct ServerlessTestContext {
+    pub cfg: Arc<ConfigurationInner>,
+    workdir: TempDir,
+}
+
+pub(crate) fn build_config_without_server(
+    cfg: Option<ConfigurationInner>,
```
svix-james, line 190: Ok, this is a bit awkward, but hoping we won't need `build_config_without_server` at all once the bootstrap stuff is using proper APIs.

svix-jplatte (approved):

PR #92, opened 2026-02-05
Add criterion-based benchmarking

svix-james:
Just some kv-benchmarks for now. Also refactors more of the common
test server setup code.

### DIFF
--- a/benches/kv.rs
+++ b/benches/kv.rs
@@ -1,0 +1,141 @@
+use coyote_client::models::{KvDeleteIn, KvGetIn, KvSetIn};
+use criterion::{BatchSize, Criterion, criterion_group, criterion_main};
+use rand::{
+    Rng,
+    distr::{Alphanumeric, SampleString},
+};
+#[path = "utils.rs"]
+mod utils;
+
+use crate::utils::setup_server_simple;
+
+fn bench_kv(c: &mut Criterion) {
+    let ctx = setup_server_simple();
+    let client = ctx.client;
+
+    let mut rng = rand::rng();
+    let rt = tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()
+        .unwrap();
+
+    let test_key = "test-key".to_string();
+    let mut test_val = vec![0u8; 256];
+    rng.fill(&mut test_val[..]);
+
+    c.bench_function("kv_set_fixed", |b| {
+        b.iter_batched(
+            || (test_key.clone(), test_val.clone()),
+            |(key, val)| {
+                rt.block_on(async {
+                    std::hint::black_box(client.kv().set(KvSetIn::new(key, val), None))
+                        .await
+                        .unwrap();
+                })
+            },
+            BatchSize::SmallInput,
+        )
+    });
+
+    c.bench_function("kv_set_random_small", |b| {
+        b.iter_batched(
+            || {
+                let k = Alphanumeric.sample_string(&mut rng, 16);
+                let mut v = vec![0u8; 256];
+                rng.fill(&mut v[..]);
+                (k, v)
+            },
+            |(k, v)| {
+                rt.block_on(async {
+                    std::hint::black_box(client.kv().set(KvSetIn::new(k, v), None))
+                        .await
+                        .unwrap();
+                })
+            },
+            criterion::BatchSize::SmallInput,
+        )
+    });
+
+    c.bench_function("kv_set_random_medium_value", |b| {
+        b.iter_batched(
+            || {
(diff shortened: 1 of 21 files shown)


src/core/otel_spans.rs
```diff
@@ -92,9 +92,8 @@ impl<B> MakeSpan<B> for AxumOtelSpanCreator {
             app_id = tracing::field::Empty,
         );
 
-        if let Err(e) = span.set_parent(remote_context) {
+        if span.set_parent(remote_context).is_err() {
```
svix-james, line 95: This is blowing up the `bench` runs 🤷

crates/test-utils/src/server.rs
```diff
@@ -186,22 +198,17 @@ pub(crate) struct ServerlessTestContext {
     workdir: TempDir,
 }
 
-pub(crate) fn build_config_without_server(
-    cfg: Option<ConfigurationInner>,
-) -> ServerlessTestContext {
-    setup_tracing_for_tests();
```
svix-james, line 195: I'm not clear on why this was needed. But directly calling `setup_tracing_for_tests` breaks things locally for me.

benches/utils.rs
```diff
@@ -0,0 +1,51 @@
+#![allow(unused)]
+
+use coyote::cfg::LogLevel;
+use coyote_client::{CoyoteClient, CoyoteOptions};
+use tempfile::TempDir;
+use tokio::runtime::Runtime;
+
+#[path = "../tests/it/server/mod.rs"]
+pub(crate) mod test_server;
```
svix-jplatte: We should not be doubly compiling modules to share them. If the test server is useful both in the `coyote` crates' integration tests and here, move it to the test-utils crate please.

benches/kv.rs
```diff
@@ -0,0 +1,141 @@
+use coyote_client::models::{KvDeleteIn, KvGetIn, KvSetIn};
+use criterion::{BatchSize, Criterion, criterion_group, criterion_main};
+use rand::{
+    Rng,
+    distr::{Alphanumeric, SampleString},
+};
+#[path = "utils.rs"]
+mod utils;
```
svix-jplatte: If we need this bench to consist of multiple modules, let's move it to `benches/kv/main.rs`, then the `#[path]` won't be necessary.

svix-jplatte (changes requested):

benches/utils.rs
svix-james: Ok, for some reason I thought I'd hit circular dependency issues if I did so, so didn't even try it. But let's do it.

benches/kv.rs
svix-james: Ok, so this is actually common bench code. It could potentially also be move to `test-utils`, but this is not code specific just to `kv` benchmarks.

svix-jplatte: Do we want separate benchmarks to live in different executables? Maybe we should have a separate package for benches then and use the library crate part of it for helpers then?

svix-james: I honestly don't know the best practices here, but this is how I've seen things organized before. Ideally we can run benchmarks for individual modules separately w/out too much trouble. That's my only requirement. Tokio organizes it into what I guess become separate executables after compilation(?): https://github.com/tokio-rs/tokio/tree/master/benches

svix-jplatte: Yeah, those are different executables in tokio.
I'd say let's do a `benchmarks/Cargo.toml` package where we can add bench-only dependencies, put actual benchmark code in `benches` subdirectory (a bit awkward, but works) and bench utility code in `src` subdirectory.

svix-james: Ok, took a stab at it. Not sure how the benches are supposed to reference the lib code, but this works. 

This also has the benefit of not spamming stdout w/ ignored tests, which is great.

benchmarks/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "benchmarks"
```
svix-jplatte: I'd name this `coyote-benchmarks`

benchmarks/benches/kv.rs
```diff
@@ -4,7 +4,8 @@ use rand::{
     Rng,
     distr::{Alphanumeric, SampleString},
 };
-#[path = "utils.rs"]
+
+#[path = "../src/lib.rs"]
 mod utils;
```
svix-jplatte: No need for this, there's an implicit dependency on the library crate from the same package (currently `benchmarks`, or `coyote_benchmarks` with the change proposed above).

svix-jplatte (approved):
Good to merge with the last `#[path] mod` removed

PR #101, opened 2026-02-06
Wire up rate-limiter to Raft

svix-lucho:
Same as https://github.com/svix/coyote-private/pull/45 but for the rate limiter. 

Added operations for `limit` and `reset`. I didn't add one for `get_remaining` since it's only a read. 

~There's a bunch of repeated code inside `operations`, but I think it's good to have a few more examples before replacing it with a macro. I'll see if my little brain can come up with a good one later.~ I added the macro

Part of https://github.com/svix/coyote-private/issues/80


svix-lucho:
the macro to reduce boilerplate is in https://github.com/svix/coyote-private/pull/114

svix-james:
Sorry this is just an unhelpful side thought.... but damn this module is so latency-critical, I'm just wondering why we even want to delay it by requiring consensus :sweat_smile:

svix-james (approved):
Looks good to me, above comment notwithstanding.

svix-lucho:
> Sorry this is just an unhelpful side thought.... but damn this module is so latency-critical, I'm just wondering why we even want to delay it by requiring consensus 😅

I 100% agree with your concern. But I think that falls under optimizations we have to do later. I'm just trying to give all modules a similar structure here. If/when we decide to make tweaks in specific areas, we can do it.

One way of optimizing this would be to do consensus in batches (with the tradeoff of having inconsistent data during failovers, etc) and/or run the consensus out of band from the main request. That seems way out of scope for what I'm doing here.

PR #102, opened 2026-02-06
example configs to boot a three-node cluster

svix-jbrown:
This is pretty janky but it does work most of the time.

Sometimes the nodes don't quite bootstrap right; the discovery code is very bad and needs to be improved.

### DIFF
--- a/example-configs/three-node-cluster/prefix-output.sh
+++ b/example-configs/three-node-cluster/prefix-output.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+TAG="$1"
+shift
+
+if [[ -z "$TAG" ]]; then
+    echo >&2 "Usage: tag command"
+    exit 2
+fi
+
+exec > >(
+    trap "" INT TERM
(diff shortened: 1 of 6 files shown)


example-configs/three-node-cluster/prefix-output.sh
```diff
@@ -0,0 +1,20 @@
...
+exec > >(
+    trap "" INT TERM
+    sed -u "s/^/${TAG} /"
+)
+exec 2> >(
+    trap "" INT TERM
+    sed -u "s/^/${TAG} /" >&2
+)
+
+exec "$@"
```
svix-jbrown, line 20: this file exists so that we can prefix our log lines with which instance they came from. I couldn't find anything in `tracing_subscriber` to customize the formatting by adding a tag without replacing the formatter wholesale, which is annoying.

We have to do this dance with `exec >` to ensure that we can forward signals properly so that a sigint kills the whole thing instead of just killing the sed process.

svix-james (approved):

PR #110, opened 2026-02-07
Stream: Use configgroups

svix-james:
Instead of creating a separate `StreamRow`, just create a `ConfigGroup`. This also removes the need for `NameToStreamRow`, since config groups are already indexed by name (and module).

Also, streams AFAIK shouldn't have a "default" config group like the others, so removed that from the bootstrap file for now.


src/lib.rs
```diff
@@ -211,6 +211,12 @@ impl AppState {
         let kv_store = self.get_store_by_key::<IdempotencyConfig>(key_name)?;
         Ok(IdempotencyStore::new(kv_store))
     }
+
+    pub fn get_stream(&self, name: &str) -> Result<ConfigGroup<StreamConfig>> {
+        self.configgroup_state
+            .fetch_group::<StreamConfig>(name)?
+            .ok_or_else(|| Error::generic(format!("Stream {name} not found")))
```
svix-gabriel: I realize this is following the patterns established by the other `get_*` methods, but why doesn't this return `Result<Option<ConfigGroup<...>>>`? Are these really only ever called if we expect the Stream to exist?

src/v1/endpoints/stream.rs
```diff
@@ -343,11 +357,13 @@ struct RedriveOut {}
 /// Redrives messages from the dead letter queue back to the stream.
 #[aide_annotate(op_id = "v1.stream.redrive")]
 async fn redrive(
-    State(AppState { stream_state, .. }): State<AppState>,
+    State(state): State<AppState>,
     MsgPackOrJson(data): MsgPackOrJson<RedriveIn>,
 ) -> Result<MsgPackOrJson<RedriveOut>> {
+    let group = state.get_stream(&data.name)?;
```
svix-gabriel, line 363: Ok, this makes me think `get_stream` really should return `Result<Option<...>>`. Right now it returns `Error::generic(...)` for a non-extant stream, which doesn't seem appropriate here

crates/stream/src/tables.rs
```diff
@@ -15,67 +11,6 @@ use crate::{
     entities::{ConsumerGroup, MsgHeaders, MsgId, StreamId},
 };
 
-// IMPORTANT. Since these are all shared in the same fjall::Keyspace, the table prefixes must be unique.
-static_assertions::const_assert!(fjall_utils::are_all_unique(&[
```
svix-gabriel, line 19: Why remove this assertion, rather than just removing the `StreamRow::TABLE_PREFIX` from the array? It's still important that the other prefixes are definitively unique.

crates/stream/src/entities.rs
```diff
@@ -9,7 +10,7 @@ use validator::Validate;
 // FIXME(@svix-gabriel) - I opted for type aliases here just for expediency.
 // We absolutely can (and should) use more robust types.
 pub type StreamId = Uuid;
```
svix-gabriel: Do we even need the `StreamId` type anymore? Seems like they're all now `ConfigGroupId`?

crates/stream/src/tables.rs
svix-james: So the only thing that would be here is `LeaseRow`, right? ... we can keep it, just seemed silly to check for the uniqueness of one thing.

src/v1/endpoints/stream.rs
svix-james: I just figured this particular method shouldn't return an `Option` given that we don't want to do anything without there already being a stream? 

I don't care that much, just seemed easier than mapping a `None` everywhere.

crates/stream/src/tables.rs
svix-gabriel: Ok, fair.

I was thinking if we add more tables in the future, I like that we explicitly call out the table prefixes need to be unique.

But IDK what other tables we'd add here, so 🤷

src/lib.rs
svix-james: Yes, that was the idea.

svix-gabriel: Ok, I guess at the very least I'd just change this so it isn't Error::generic, and instead Error::user_facing or some such, since AFAIK we always use Error::generic for internal servery errors

crates/stream/src/entities.rs
svix-james: Yep, removed it.

svix-gabriel (approved):

PR #111, opened 2026-02-09
Clean up scalar docs

svix-jplatte:
We had lots of Svix copy-paste stuff in there.


src/static/docs.html
```diff
@@ -125,19 +22,6 @@
...
-      // Ensure old webhook urls keep working
-      // old link: https://api.svix.com/docs#tag/Webhook/operation/endpoint.updated
-      // new link: https://api.svix.com/docs#tag/Webhook/webhook/endpoint.updated
-      // Scalar (we can't modify this) prefixes each webhook slug with /webhook, that code allows us to keep old urls working
-      redirect: (hash) => {
-        if (hash.startsWith("#tag/Webhook/operation/")) {
-          return hash.replace("#tag/Webhook/operation/", "#tag/Webhook/webhook/")
-        }
-        return hash;
-      },
```
tasn, line 140: This can go, but the rest we should keep.

src/static/docs.html
```diff
@@ -2,116 +2,13 @@
 <html>
 
 <head>
-  <title>Svix API Reference</title>
+  <title>Coyote API Reference</title>
   <!-- needed for adaptive design -->
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
-
```
tasn, line 9: Let's keep all of this, even if only so we remember to change it. :)

src/openapi.rs
```diff
@@ -41,54 +41,47 @@ pub fn initialize_openapi() -> OpenApi {
...
-            title: "Svix API".to_owned(),
+            title: "Coyote API".to_owned(),
             version: VERSION.to_owned(),
-            extensions: indexmap::indexmap! {
-                "x-logo".to_string() => serde_json::json!({
-                    "url": "https://www.svix.com/static/img/brand-padded.svg",
-                    "altText": "Svix Logo",
-                }),
-            },
+            // extensions: indexmap::indexmap! {
```
tasn: We can keep this temporarily.

src/static/docs.html
```diff
@@ -125,19 +126,6 @@
       // FIXME hiddenClients: true,
       darkMode: false,
       showToolbar: "never",
-      generateOperationSlug: (operation) => `operation/${operation.operationId}`,
-      generateTagSlug: (tag) => `${tag.name.replace(" ", "-")}`,
-      generateWebhookSlug: (webhook) => webhook.name,
```
tasn, line 130: Let's keep these as well.

svix-jplatte: why?

tasn: We don't have to, I'll approve the PR. The thinking was to match what we already do elsewhere (which has been working for us), so why change.

tasn (approved):

src/static/docs.html
svix-jplatte: I think we only have this in Svix such that `#anchors` that worked with Redoc continue to work.

PR #114, opened 2026-02-09
Macro to declare raft operations in modules

svix-lucho:
Add a `raft_module_operations` macro to help reduce most of the boilerplate around declaring raft operations. 

The idea is that we only need to define the `struct` for the operation input, the output (`FooOperation` and `FooResponse`), and the `apply` method.

For example, for a module M1, we'd use it like this

```rust
raft_module_operations!(
    M1Request,
    M1Operation {
        Foo(FooOperation) -> FooResponse,
        ...
    },
    state = &M1State,
);

struct FooOperation {}

struct FooResponse {}

impl M1Request for FooOperation {
  fn apply(s: &M1State) {
    // apply the operation in state s
  }
}
```

I don't know if I went too crazy with the macro. I like that it eliminates all the boilerplate, but I'm worried it might be doing a lot of things behind the scenes. I want to get your opinions. There's also a big chance that this can be simplified in some way that I didn't see - there's a lot of indirection between all the types for Raft so it's easy to get lost in the complexity.


svix-jbrown (approved):
This looks very similar to what I'd prototyped last week.

PR #115, opened 2026-02-09
Fix bug in rate limiter and add more tests

svix-lucho:
The token bucket implementation had a very stupid bug 🤦  because it was updating the `last_refill` time on every `limit` call, so the bucket never got refilled. I didn't catch it in the existing tests because I tested it with `get_remaining`, which didn't repro the bug.

This adds a test for this case, and a few more tests for the module.


svix-lucho:
This is trivial, I'll just merge it

Issue #106, opened 2026-02-04
Figure out the best way to serialize/deserialize bytes when using JSON


svix-lucho:
When we use `Vec<u8>` as the type in the API, the OpenAPI spec looks like this
```
"type": "array",
  "items": {
    "type": "integer",
    "format": "uint8",
    "maximum": 255,
    "minimum": 0
  }
```
which is very inefficient and awkward. We should have a special type that transforms into base64 or something more suitable for byte arrays. 

Use the new type for values in KV, cache, idempotency and stream/queue

tasn:
Idea: have a wrapper value.
So:
```rust
struct KvIn {
  key: String,
  value: bytes,
}
```
Would translate to:
```js
{
  "key": "foo",
  "value": { "format": "base64", "data": "RI0dlG0aGIXoi1Xfo8khICy2" }
}
```
Or
```js
{
  "key": "foo",
  "value": { "format": "text" /* Default */ , "data": "hello" }
}
```
Or we can even accept (input only, not output, and only with curl, so not actually exposed through the API):
```js
{
  "key": "foo",
  "value": "hello",
}
```

When serializing (returning to user) we'll always use the long form:
```js
{
  "key": "foo",
  "value": { "format": "text", "data": "hello" }
}
```
and we always try to serialize to string ("text"), but if that fails, we fallback to base64.

Goes without saying: this doesn't affect msgpack, just json.

svix-jplatte:
Grooming
- We can launch MVP with JSON support disabled

PR #116, opened 2026-02-09
Module API feedback 2026-02-09

tasn:
See the markdown file in the PR.


tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+- [ ] Docs viewer:
+  - [ ] Should remove all the "POST" on the sidebar, as everything is "POST" given how we do RPC.
+  - [ ] Figure out if there's a way to de-emphasize the HTTP nature of it (maybe we need a different viewer) - as we don't care about 200, almost everything should be 200, but we want to show the outputs.
+  - [ ] Want to show that body can be msgpack too (but same kind of data).
+  - [ ] Start with the response schema open?
+- [ ] Need to do the same `enum` label fixes we did for Svix.
+- [ ] I haven't reviewed the `v1.cache.set` "operation naming" - need to do that.
+- [ ] `key` should also support `/` and tbh probably `#`, `|` and potentially other common delimiters.
+- [ ] Default values should be better for properties (in the docs).
+- [ ] SDKs: do we generate enums for enums? We should if not.
```
svix-jplatte, line 17: Yes, same thing we have in Svix (though Lucho had some issues with codegen at one point, not sure if fixed)

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
+# Review of the Coyote API 2026-02-09
+
+Note: I reviewed the docs (localhost:8050/docs), so maybe some things are correct but just not documented correctly?
+
+IMPORTANT: I only commented on the things that need to change, not all the great things there. So sorry that it's a long list of "let's change this" and not a lot of "this is awesome", but it's mostly awesome, just comments for API consistency and other improvement ideas.
+
+## General
```
svix-jplatte, line 7: One thing I noticed is that we use the "naive" encoding of `Vec<u8>`, which is great for msgpack but horrible for JSON. We should decide on how to handle binary fields in JSON in a less awful way.

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+  - [ ] All the comments from cache.
+- [ ] Cache delete:
+  - [ ] All the comments from cache.
+- [ ] Add a "scan" operation to list keys. I think we should probably support * and maybe in the future ** like mentioned in the ACL RFC. Though for now, maybe let's start with just prefix, so just foo/bar/* or foo/* (so always needs to have a /* at the end, and that's the only place a * is allowed for now.
+
+## Rate-limiter
+- [ ] I mentioned it somewhere (don't remember where and what came out of it): Token bucket is superior to fixed window and is a superset of it - you can implement fixed window with token bucket when refill time is the same as window size.
+  - [ ] This means that having both is needlessly confusing (let's remove fixed window).
+  - [ ] Though I'd also say: do we even want multiple strategies, or just one? E.g. maybe we want leaky bucket? Though maybe that can be a future thing and we shouldn't soil the API yet? (as long as we can support it later with `type` added optionally and changes the config).
+  - [ ] Either way, `method` feels like an add name for the type.
```
svix-jplatte: Did you mean *bad* name?

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+  - [ ] Retention:
+    - [ ] `maxByteSize` and `retentionPeriodSeconds` should both sit under a nested `retention` and be called `size` and `time` or `size` and `age`.
+    - [ ] `time/age` should be in milliseconds (not seconds). `size` should remain in bytes.
+  - [ ] Also mentioned in General above: returning entities upon creation.
+  - [ ] `created_at` and `updated_at` - I mentioned it in kv, let's remove the `_at`.
+  - [ ] Should there be a stream-level default visibility timeout? Can override on receive, but at least have a default? Feels easier, especially with multi-language codebases.
+- [ ] Append to Stream:
+  - [ ] Should rename it to `publish` or whatever else is common. Append is unusual for a stream product (kafka, pulsar, rabbitmq, all don't use it).
+  - [ ] `payload` -> `value` to be more consistent with the rest of the API (and Kafka and the likes).
+  - [ ] I don't know how it'll look like in the SDK, maybe ask Jonas, but I'd love it to be able to send `{"msgs": ["msg1", "msg2"]}` instead of the full decorated record every time (want to support both).
```
svix-jplatte, line 95: I think the only realistic way of supporting this is hand-coding that particular method for each SDK.

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+  - [ ] Returning data:
+    - [ ] `retry_after`: isn't that just `refill_interval` always? Unless we support multiple `token` fetches and then it's a bit more complicated potentially. Should this be in seconds or ms if we change refill interval too?
+    - [ ] Here we called it `result` is idempotency we called it `status`. In idempotency value is all lowercase (`completed`) here it's upper case (`BLOCK`). Need to be consistent.
+    - [ ] `remaining` -> `available`? Not sure, just bouncing ideas. Maybe it's fine.
+- [ ] Get remaining:
+  - [ ] Not sure: but do we want `retry_after` here? Given that it's not really a try? It also lets us not pass the config which I thin kis a big win for this function in terms of simplicity?
+- [ ] How do you envision "multi rate-limit" looking like in the future? So letting us limit on a few in parallel?
+- [ ] How do we pass the config around for the rate-limiter? Do we want to init something before, or just pass it live?
+  - [ ] I'm just concerned it may be (1) an implementation detail, and (2) confusing.
+- [ ] Should we add a command to reset the amount of tokens to a specific number? "Set remaining" or something?
```
svix-lucho, line 81: There is a reset route

svix-jplatte (commented):
oops, didn't actually submit the comments I wrote earlier

tmp/api-feedback-2026-02-09.md
tasn: There's https://github.com/svix/coyote-private/issues/106 for that.

tmp/api-feedback-2026-02-09.md
tasn: It didn't show up in the docs, can you please fix that and lmk so that I can review it?

tmp/api-feedback-2026-02-09.md
tasn: That's fine too.

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Stream
+- [ ] Create stream:
+  - [ ] `name` - let's talk about that later. If we change this to be `namespace` maybe it should be name, though need to decide.
+  - [ ] Retention:
+    - [ ] `maxByteSize` and `retentionPeriodSeconds` should both sit under a nested `retention` and be called `size` and `time` or `size` and `age`.
```
svix-gabriel, line 87: I'm on board, though I think we should keep some specificity of units in the size and time since it avoids any ambiguity. e.g.
```json
{
    "name": "the stream",
    "retention": {
         "sizeBytes": 1024,
         "ageSeconds": 12
     }
}
```

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+    - [ ] `time/age` should be in milliseconds (not seconds). `size` should remain in bytes.
+  - [ ] Also mentioned in General above: returning entities upon creation.
+  - [ ] `created_at` and `updated_at` - I mentioned it in kv, let's remove the `_at`.
+  - [ ] Should there be a stream-level default visibility timeout? Can override on receive, but at least have a default? Feels easier, especially with multi-language codebases.
+- [ ] Append to Stream:
+  - [ ] Should rename it to `publish` or whatever else is common. Append is unusual for a stream product (kafka, pulsar, rabbitmq, all don't use it).
+  - [ ] `payload` -> `value` to be more consistent with the rest of the API (and Kafka and the likes).
+  - [ ] I don't know how it'll look like in the SDK, maybe ask Jonas, but I'd love it to be able to send `{"msgs": ["msg1", "msg2"]}` instead of the full decorated record every time (want to support both).
```
svix-gabriel, line 95: I don't quite get this one. Is the suggestion here that

```
{
   "msgs": ["msg1", "msg2"]
}
```

in lieu of

```
{

   "msgs": [{"value": "msg1" }, {"value": "msg2"}]
}

File: tmp/api-feedback-2026-02-09.md [hunk 8]
```diff
@@ -0,0 +1,125 @@
...
+    - [ ] In rust it's easy with `.into()` but I wonder about other languages.
+  - [ ] I'm more of a `msgs` guy myself (so not sure I even want to make this comment), but I wonder if we should call it `messages` as we don't use shortened versions elsewhere?
+  - [ ] Return info:
+    - [ ] Should change `msgIds` to `offset` and return just an integer (last offset iirc, see what Kafka does). The way stream appending works means that the messages are always going to go in together, so always sequential IDs. No need to send them separately (that's also what Kafka does).
+- [ ] Fetch from stream:
+  - [ ] `fetch` -> `receive` or maybe `consume`? Fetch feels weird to me.
```
svix-gabriel, line 101: I'm on board with `receive`

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+  - [ ] Where does it start fetching from? Is this something we can control? How do we control it? Should we have an optional property to allow for that initial seek?
+  - [ ] Again snake_case over camelCase.
```
svix-gabriel, line 103: We want snake_case over camelCase?

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+  - [ ] `payload` -> `value` to be more consistent with the rest of the API (and Kafka and the likes).
+  - [ ] I don't know how it'll look like in the SDK, maybe ask Jonas, but I'd love it to be able to send `{"msgs": ["msg1", "msg2"]}` instead of the full decorated record every time (want to support both).
+    - [ ] In rust it's easy with `.into()` but I wonder about other languages.
+  - [ ] I'm more of a `msgs` guy myself (so not sure I even want to make this comment), but I wonder if we should call it `messages` as we don't use shortened versions elsewhere?
+  - [ ] Return info:
+    - [ ] Should change `msgIds` to `offset` and return just an integer (last offset iirc, see what Kafka does). The way stream appending works means that the messages are always going to go in together, so always sequential IDs. No need to send them separately (that's also what Kafka does).
+- [ ] Fetch from stream:
+  - [ ] `fetch` -> `receive` or maybe `consume`? Fetch feels weird to me.
+  - [ ] Where does it start fetching from? Is this something we can control? How do we control it? Should we have an optional property to allow for that initial seek?
```
svix-gabriel, line 102: It always fetches from the last position seen by that Consumer Group (the first message in the stream if the consumer group hasn't been registered)

> Should we have an optional property to allow for that initial seek?

We can. It'd be easy to support. I would just have it be something like

```
POST /stream/receive

{
    "streamName": "...."
    "consumer_group": "...",
    "from_offset": 12
}
```

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+  - [ ] Again snake_case over camelCase.
+  - [ ] `consumerGroup` -> `consumer`? I don't think we lose any information, and it makes it more concise and potentially even clearer?
```
svix-gabriel, line 104: I disagree. I think `consumerGroup` and `consumer` mean functionally different things, and I think it would introduce confusion with people coming from Kafka if we changed the meaning of the terms.

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+  - [ ] `visibilityTimeoutSeconds` -> `visibilityTimeout`: no need for the word seconds. Also, maybe let's make it milliseconds since that's what we use elsewhere in the API?
+  - [ ] visibilityTimeout: can make it optional if we set it on the stream level itself.
```
svix-gabriel, line 106: My 2c - I don't think it should be a stream level setting. Having the visibilityTimeout only be configurable from on the receive call makes things really explicit, and avoids the trap of "multiple ways to do things"

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+  - [ ] In addition to `batchSize` need `batchWait`, which is the amount of time to wait/block before returning if you don't have `batchSize` messages available?
+- [ ] Locking fetch from stream:
+  - [ ] I think this should be a `behavior` or `consumer_type` or something on the consumer group, not a separate API (so the same call as `fetch from stream`).
```
svix-gabriel, line 109: Something like `consumer_type: "queue"` or `consumer_type: "stream"`?

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+    - [ ] This should also fail if there is already a consumer group with a different behavior assigned to it.
+- [ ] Ack range:
+  - [ ] This should accept `offset` instead of `min/maxMsgId`. If you're processing a stream, then you want to commit an offset. If you're processing messages individually then to use it you'll need to review and verify they are sequential which is not fun (if we want batch ack, see next section).
+  - [ ] This API should only work with behavior: stream.
+  - [ ] This should have a different name. In kafka it's called "commit offset" iirc.
+- [ ] Ack
+  - [ ] This API should only work with behavior: queue. Maybe we should name it accordingly? E.g. prefix all the queue semantics one to have queue in the name?
+  - [ ] Allow passing a list of msg ids for batch ack?
+- [ ] Dlq
+  - [ ] I think we should have `nack` instead, and then a configuration on the stream for what to do with nack (e.g. after X failures, move to DLQ, before then, follow this visibility/retry schedule).
```
svix-gabriel, line 119: Are you envisioning something like

```
POST /stream/nack
{
     stream_name: "...",
     offset: "...",
     retry_after_millis: 199932
}
```

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+  - [ ] `consumerGroup` -> `consumer`? I don't think we lose any information, and it makes it more concise and potentially even clearer?
+  - [ ] `visibilityTimeoutSeconds` -> `visibilityTimeout`: no need for the word seconds. Also, maybe let's make it milliseconds since that's what we use elsewhere in the API?
+  - [ ] visibilityTimeout: can make it optional if we set it on the stream level itself.
+  - [ ] In addition to `batchSize` need `batchWait`, which is the amount of time to wait/block before returning if you don't have `batchSize` messages available?
+- [ ] Locking fetch from stream:
+  - [ ] I think this should be a `behavior` or `consumer_type` or something on the consumer group, not a separate API (so the same call as `fetch from stream`).
+    - [ ] This should also fail if there is already a consumer group with a different behavior assigned to it.
+- [ ] Ack range:
+  - [ ] This should accept `offset` instead of `min/maxMsgId`. If you're processing a stream, then you want to commit an offset. If you're processing messages individually then to use it you'll need to review and verify they are sequential which is not fun (if we want batch ack, see next section).
```
svix-gabriel, line 112: I'm fine w/this, but my worry with offset is

1. we should be consistent with the API
2. "offset" is gonna be inconsistent with `nack`/`ack`, since in one context (acking for Stream, it acks an entire message range), and in another context (queue), it only acks a single message.

I would be fine with just delineating the Stream/Queue into separate API paths `/stream` vs `/queue`, and then always using `msgId` for queue, and offset for stream.

svix-gabriel (commented):
Anything w.r.t. to Stream that I didn't comment on is something I implicitly agree with

tmp/api-feedback-2026-02-09.md
tasn: With size: I've never seen size be represented in anything other than bytes, so I don't think it's ambiguous.
age: we use milliseconds throughout the API (I made comments on kv/cache too), so I think it's also consistent and clear. Though yeah, age could be ambiguous without a unit in the name.

I never liked APIs that add the unit, it feels like noise. I think for "seconds" it maybe works, but we probably want it to be milliseconds, and then it's extra ugly (`ageMs`). Curious to hear what other people think, but that's why I was pushing back on including the seconds (and similarly the `_at` for timestamps below).

Note: please use snake_case (not camelCase) - I made the comment elsewhere, but Coyote is all camelCase. Even in the JSON/msgpack.

tmp/api-feedback-2026-02-09.md
tasn: Support both, yes. I care less about the wire protocol (though for that we should probably support both as well!), but more about the API, e.g:

```python
coyote.stream.publish(PublishIn(msgs=["msg1", "msg2"]))
# vs.
coyote.stream.publish(PublishIn(msgs=[MessageIn(value="msg1"), MessageIn(value="msg2")]))
```

Though as I said, I think we should support both, just the shortened version as well.

tmp/api-feedback-2026-02-09.md
tasn: Yup. I think this is one of our Svix regrets.

tmp/api-feedback-2026-02-09.md
tasn: OK, no strong feelings there.

tmp/api-feedback-2026-02-09.md
tasn: OK, I'm convinced.

tmp/api-feedback-2026-02-09.md
tasn: Yeah, sounds good to me.

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+- [ ] Get remaining:
+  - [ ] Not sure: but do we want `retry_after` here? Given that it's not really a try? It also lets us not pass the config which I thin kis a big win for this function in terms of simplicity?
+- [ ] How do you envision "multi rate-limit" looking like in the future? So letting us limit on a few in parallel?
+- [ ] How do we pass the config around for the rate-limiter? Do we want to init something before, or just pass it live?
+  - [ ] I'm just concerned it may be (1) an implementation detail, and (2) confusing.
+- [ ] Should we add a command to reset the amount of tokens to a specific number? "Set remaining" or something?
+
+## Stream
+- [ ] Create stream:
+  - [ ] `name` - let's talk about that later. If we change this to be `namespace` maybe it should be name, though need to decide.
```
tasn, line 85: I realized I haven't made the second part of the comment: I think it should be called `id` or `topic` or something for a specific stream.

tmp/api-feedback-2026-02-09.md
tasn: > It always fetches from the last position seen by that Consumer Group (the first message in the stream if the consumer group hasn't been registered)

Kafka starts from the most recent message if it's the first time (and what we did for Svix too for that reason), and I think that makes sense for Coyote (maybe less so for Svix fwiw). Even if only because going back in time so far could be expensive because old data may be in cold storage, etc. Also, the start of the stream is kind of arbitrary because of retention.

> We can. It'd be easy to support. I would just have it be something like

I think `offset` is preferred to `from_offset` (see all the rest of my comments about removing redundant verbs/adjectives/etc). Though we also want to be able to "reset" based on timestamp, not just offset.

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+- [ ] Append to Stream:
+  - [ ] Should rename it to `publish` or whatever else is common. Append is unusual for a stream product (kafka, pulsar, rabbitmq, all don't use it).
+  - [ ] `payload` -> `value` to be more consistent with the rest of the API (and Kafka and the likes).
+  - [ ] I don't know how it'll look like in the SDK, maybe ask Jonas, but I'd love it to be able to send `{"msgs": ["msg1", "msg2"]}` instead of the full decorated record every time (want to support both).
+    - [ ] In rust it's easy with `.into()` but I wonder about other languages.
+  - [ ] I'm more of a `msgs` guy myself (so not sure I even want to make this comment), but I wonder if we should call it `messages` as we don't use shortened versions elsewhere?
+  - [ ] Return info:
+    - [ ] Should change `msgIds` to `offset` and return just an integer (last offset iirc, see what Kafka does). The way stream appending works means that the messages are always going to go in together, so always sequential IDs. No need to send them separately (that's also what Kafka does).
+- [ ] Fetch from stream:
```
tasn, line 100: Need to have a "seek" operation that seeks to a specific offset/timestamp (and that works similarly to the auto-initial-offset seek we talked about below).

tmp/api-feedback-2026-02-09.md
tasn: There are two kinds of nacks (stream and queue), but I was actually thinking it could be the same API for both. For both of them it works on only ONE message at a time.

```
POST /stream/nack
{
     name: "...",
     msgId: "..."
}
```

Please note that I removed the `retry_after_millis`, as that is configured on the stream itself (the retry schedule for `nacks`).

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+  - [ ] `fetch` -> `receive` or maybe `consume`? Fetch feels weird to me.
+  - [ ] Where does it start fetching from? Is this something we can control? How do we control it? Should we have an optional property to allow for that initial seek?
+  - [ ] Again snake_case over camelCase.
+  - [ ] `consumerGroup` -> `consumer`? I don't think we lose any information, and it makes it more concise and potentially even clearer?
+  - [ ] `visibilityTimeoutSeconds` -> `visibilityTimeout`: no need for the word seconds. Also, maybe let's make it milliseconds since that's what we use elsewhere in the API?
```
tasn, line 105: Make sure it's optional, and we have a sane default (e.g. 10s or 30s).

tmp/api-feedback-2026-02-09.md
tasn: It's not inconsistent with ack/nack. They work on a specific message, this works on a range. That's why it's named differently: offset denotes range, msgId denotes item.

Though I agree that we can do better (see next paragraph):

> I would be fine with just delineating the Stream/Queue into separate API paths /stream vs /queue, and then always using msgId for queue, and offset for stream.

Can you show a suggestion that makes it clear that queue calls work on streams, but only on queue consumer groups? I'm with you that it's better, just don't have a good idea.

tmp/api-feedback-2026-02-09.md
svix-gabriel: I guess I'll see what others say. My take is that, for someone first interacting with the API, if they see

```javascript
coyote.stream.create({
    "name": "the stream",
    "retention": {
       "age_ms": 90000,
       "size_bytes": 65536
    }
})
```

that's much more self documenting than 
```javascript
coyote.stream.create({
    "name": "the stream",
    "retention": {
       "age": 90000,
       "size": 65536
    }
})
```

tasn: I guess `age_ms` is much less offensive than `ageMs`, so that's good (now that I saw it). :)

For this specific API, I wonder if this makes more sense (not sure):

```
coyote.stream.create({
    "name": "the stream",
    "retention": {
       "milliseconds": 90000,
       "bytes": 65536
    }
})
```

Another question for us, if we decide to add `_ms` here, do we add it everywhere else? E.g. `ttl_ms`, which feels weird to me.

tmp/api-feedback-2026-02-09.md
svix-gabriel: > Kafka starts from the most recent message if it's the first time

It's the default behavior, but it is configurable, and my own experience was that it's a confusing default.

I would be fine w/making this configurable. Kafka does it with `auto.offset.reset=latest|none|earliest`, I think something like that would be fine here. e.g.

```
POST /stream/receive

{
    "streamName": "...."
    "consumer_group": "...",
    "starting_offset": earliest | latest | <specific id>
}
```

or (probably better)
```
POST /stream/receive

{
    "streamName": "...."
    "consumer_group": "...",
    "starting_position": "earliest" | "latest"
    "offset": <optional specific id>
}
```

tmp/api-feedback-2026-02-09.md
svix-gabriel: Yeah I think that's fine. I'm not attached to `name`, and `id` or `topic` both make sense here

tmp/api-feedback-2026-02-09.md
svix-gabriel: If we let `/receive/` start a specific offset, that effectively covers `seek`

tmp/api-feedback-2026-02-09.md
svix-gabriel: I like
```
    "retention": {
       "milliseconds": 90000,
       "bytes": 65536
    }
```

I do think if we go with `age_ms` here, using `*_ms` everywhere is the right move.

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+- [ ] Should we return entities upon creation? E.g. with `create stream` below? I think probably no?
+- [ ] We use `snake_case`, not `camelCase`.
+- [ ] I noticed a lot of APIs don't have error conditions. E.g. what does `stream ack` return when there's no such message id on the stream? Or a missing consumer group? Or the likes?
+
+## Cache
+- [ ] Cache Set:
+  - [ ] `expire_in` is confusing - `expiry` would be better, though `ttl` feels much clearer.
+  - [ ] Currently doesn't return an object, should at least return {}
+- [ ] Cache get:
+  - [ ] `expires_at` is a bit verbose and unnatural (we don't want to add at/in/etc. everywhere). `expiry` or `expires` would probably be better?
```
svix-lucho, line 31: Ok, I'll update

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+  - [ ] How will schemas work later? (e.g. if we want to enforce schemas on keys in kv, etc) Just adding a schema ID to kv create and the likes?
+- [ ] Need to support request timeouts, and having a consistent way of reporting it from commands.
+  - [ ] One area where it matters (and we maybe want to override it on calls) is `limit` of rate-limiter. E.g. we may want to have a low timeout there and just give up quickly as to not block requests with that?
+- [ ] Should we return entities upon creation? E.g. with `create stream` below? I think probably no?
+- [ ] We use `snake_case`, not `camelCase`.
+- [ ] I noticed a lot of APIs don't have error conditions. E.g. what does `stream ack` return when there's no such message id on the stream? Or a missing consumer group? Or the likes?
+
+## Cache
+- [ ] Cache Set:
+  - [ ] `expire_in` is confusing - `expiry` would be better, though `ttl` feels much clearer.
```
svix-lucho, line 28: I think `ttl` makes more sense since it's a time duration, not a specific date

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+  - [ ] `ttl_seconds` -> `ttl` - we don't put units in names.
+- [ ] Idempotency Abandon:
+  - [ ] `abandon` -> `abort`?
+  - [ ] What happens if there's no key already? (never called start) Do we want to return an error here in the body?
+
+## KV
+- [ ] KV Set:
+  - [ ] All the comments from cache.
+  - [ ] `behavior`: I think it's good the way it is, just curious if we had other ideas we were considering for passing this?
+  - [ ] How do I update just the TTL? Do we even want that?
```
svix-lucho, line 53: Maybe we can have a `set-ttl` operation. But idk, I'm worried it might be a bit confusing. Would you expect `set-ttl` to update the TTL starting now, or starting when the element was created?

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+## Rate-limiter
+- [ ] I mentioned it somewhere (don't remember where and what came out of it): Token bucket is superior to fixed window and is a superset of it - you can implement fixed window with token bucket when refill time is the same as window size.
+  - [ ] This means that having both is needlessly confusing (let's remove fixed window).
+  - [ ] Though I'd also say: do we even want multiple strategies, or just one? E.g. maybe we want leaky bucket? Though maybe that can be a future thing and we shouldn't soil the API yet? (as long as we can support it later with `type` added optionally and changes the config).
+  - [ ] Either way, `method` feels like a bad name for the type.
+  - [ ] NOTE: I'm not commenting on fixed window because of it.
+- [ ] Rate limiter limit:
+  - [ ] `refill_interval_seconds` -> `refill_interval`.
+  - [ ] `refill_interval` should have 1 second as the default.
+  - [ ] Should `refill_interval` be in seconds or milliseconds? I feel like milliseconds?
```
svix-lucho, line 70: Is it realistic to rate limit by the millisecond? I don't see anyone using that

tmp/api-feedback-2026-02-09.md
svix-lucho: Ah sorry I got it confused with the idempotency `abandon`. I'll add it

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+  - [ ] Should we allow optionally taking multiple tokens rather than just 1? OK, I just see it's not in the docs, but yes in the example and is called `units`. Let's call it `tokens` to match what we call it elsewhere?
+  - [ ] Returning data:
+    - [ ] `retry_after`: isn't that just `refill_interval` always? Unless we support multiple `token` fetches and then it's a bit more complicated potentially. Should this be in seconds or ms if we change refill interval too?
+    - [ ] Here we called it `result` is idempotency we called it `status`. In idempotency value is all lowercase (`completed`) here it's upper case (`BLOCK`). Need to be consistent.
+    - [ ] `remaining` -> `available`? Not sure, just bouncing ideas. Maybe it's fine.
+- [ ] Get remaining:
+  - [ ] Not sure: but do we want `retry_after` here? Given that it's not really a try? It also lets us not pass the config which I thin kis a big win for this function in terms of simplicity?
+- [ ] How do you envision "multi rate-limit" looking like in the future? So letting us limit on a few in parallel?
```
svix-lucho, line 78: I think we can add a `limit-all` op that receives the same thing as `limit`, but a list, and does the limit in parallel. So, if one of the keys is blocked, the others don't get consumed. 

If the user wants to consume even if one is blocked, they can just call `limit` sequentially. Or we could even add an option/behavior to `limit-all`, not sure what's more clear.

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+- [ ] Cache delete:
+  - [ ] returns `deleted: true`. I'm thinking out loud, but maybe a more generic (and consistent everywhere) `status: success` or `status: ok` or even `success: true` (maybe not the last one) would be better?
+    - [ ] Or is this to indicate that there was an item? If so, maybe a count of deleted items would be better? Because that will also fit wit doing things like `DELETE item/*` and the likes?
+- [ ] Fix cache get/return bytes (atm it's string).
+
+## Idempotency
+- [ ] Idempotency Start:
+  - [ ] `ttl_seconds` -> `lock_ttl`. This is a bit confusing calling it just ttl. I think my suggestion is terrible btw, but we need something that indicates what it is (I don't think it's a timeout either). Has to have a different name than the other ttl in complete.
+  - [ ] `status: completed` - I know "completed" is because idempotency previously completed. I wonder if that's the right name though, or if there's a clearer name like `found`?
+  - [ ] We are missing a `status`? There should be one for "this is already locked", one for "we got the lock", and one for "we found it!". We are missing one of the locked ones.
```
svix-lucho, line 41: Yeah it throws an HTTP error atm (that's how the previous impl worked and I didn't change it). I'll add the additional status

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+- [ ] Idempotency Complete:
+  - [ ] `complete` -> `finish` to match `start`? Or potentially rename `start` to `init`? Or some other pair?
+  - [ ] `ttl_seconds` -> `ttl` - we don't put units in names.
+- [ ] Idempotency Abandon:
+  - [ ] `abandon` -> `abort`?
+  - [ ] What happens if there's no key already? (never called start) Do we want to return an error here in the body?
```
svix-lucho, line 47: It allows calling `complete` without calling `start`. I think that's a good idea here, why return an error? I feel like it makes the user experience worse with no particular benefit

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+  - [ ] Either way, `method` feels like a bad name for the type.
+  - [ ] NOTE: I'm not commenting on fixed window because of it.
+- [ ] Rate limiter limit:
+  - [ ] `refill_interval_seconds` -> `refill_interval`.
+  - [ ] `refill_interval` should have 1 second as the default.
+  - [ ] Should `refill_interval` be in seconds or milliseconds? I feel like milliseconds?
+  - [ ] Should we allow optionally taking multiple tokens rather than just 1? OK, I just see it's not in the docs, but yes in the example and is called `units`. Let's call it `tokens` to match what we call it elsewhere?
+  - [ ] Returning data:
+    - [ ] `retry_after`: isn't that just `refill_interval` always? Unless we support multiple `token` fetches and then it's a bit more complicated potentially. Should this be in seconds or ms if we change refill interval too?
+    - [ ] Here we called it `result` is idempotency we called it `status`. In idempotency value is all lowercase (`completed`) here it's upper case (`BLOCK`). Need to be consistent.
```
svix-lucho, line 74: I prefer `status`

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+- [ ] KV Get:
+  - [ ] All the comments from cache.
+- [ ] Cache delete:
+  - [ ] All the comments from cache.
+- [ ] Add a "scan" operation to list keys. I think we should probably support * and maybe in the future ** like mentioned in the ACL RFC. Though for now, maybe let's start with just prefix, so just foo/bar/* or foo/* (so always needs to have a /* at the end, and that's the only place a * is allowed for now.
+
+## Rate-limiter
+- [ ] I mentioned it somewhere (don't remember where and what came out of it): Token bucket is superior to fixed window and is a superset of it - you can implement fixed window with token bucket when refill time is the same as window size.
+  - [ ] This means that having both is needlessly confusing (let's remove fixed window).
+  - [ ] Though I'd also say: do we even want multiple strategies, or just one? E.g. maybe we want leaky bucket? Though maybe that can be a future thing and we shouldn't soil the API yet? (as long as we can support it later with `type` added optionally and changes the config).
```
svix-lucho, line 64: IMO having a way to customize the algorithm is a good thing. I think it doesn't significantly soil the API, and makes it much easier to extend in the future.

This is the difference (not a fan of `method` and `config` as the names, but still)

```
{ method: "token_bucket", config: { refill_interval: .., refill_tokens: }
```

```
{ refill_interval: .., refill_tokens: }
```

The way I think about it is: you're using Coyote because you need a rate limiter. You'd expect the rate limiter to be powerful and adaptable to your needs, a limiter that only supports a specific algorithm feels unnecessarily limited. We're adding one small thing in the API that makes it a lot more flexible.

svix-lucho: I don't know, let's say you're using Redis and want to migrate to this. The most convenient thing would be to mimic the same rate limiter behavior you had in Redis here, in case you don't want any functional changes to your API. Why stop ourselves from giving that option?

svix-lucho: And this can also be simplified in the SDK with some small manually written code, where you instantiate the rate limiter with a specific config once and call `limit` without having to pass the config every time.

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+    - [ ] `retry_after`: isn't that just `refill_interval` always? Unless we support multiple `token` fetches and then it's a bit more complicated potentially. Should this be in seconds or ms if we change refill interval too?
+    - [ ] Here we called it `result` is idempotency we called it `status`. In idempotency value is all lowercase (`completed`) here it's upper case (`BLOCK`). Need to be consistent.
+    - [ ] `remaining` -> `available`? Not sure, just bouncing ideas. Maybe it's fine.
+- [ ] Get remaining:
+  - [ ] Not sure: but do we want `retry_after` here? Given that it's not really a try? It also lets us not pass the config which I thin kis a big win for this function in terms of simplicity?
+- [ ] How do you envision "multi rate-limit" looking like in the future? So letting us limit on a few in parallel?
+- [ ] How do we pass the config around for the rate-limiter? Do we want to init something before, or just pass it live?
+  - [ ] I'm just concerned it may be (1) an implementation detail, and (2) confusing.
+- [ ] Should we add a command to reset the amount of tokens to a specific number? "Set remaining" or something?
+
```
svix-lucho, line 82: Sent a PR that fixes most of the issues brought up here https://github.com/svix/coyote-private/pull/135

svix-gabriel:
@tasn okay I wrote https://github.com/svix/rfc/pull/44 to capture all of the proposed changes for the Stream API.

With all the requested configuration changes, I think it's going to be _really_ hard not to have a separate `stream` and `queue` API, but this is what I came up that keeps them unified.

tmp/api-feedback-2026-02-09.md
svix-lucho: Done

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+- [ ] Cache get:
+  - [ ] `expires_at` is a bit verbose and unnatural (we don't want to add at/in/etc. everywhere). `expiry` or `expires` would probably be better?
+- [ ] Cache delete:
+  - [ ] returns `deleted: true`. I'm thinking out loud, but maybe a more generic (and consistent everywhere) `status: success` or `status: ok` or even `success: true` (maybe not the last one) would be better?
+    - [ ] Or is this to indicate that there was an item? If so, maybe a count of deleted items would be better? Because that will also fit wit doing things like `DELETE item/*` and the likes?
+- [ ] Fix cache get/return bytes (atm it's string).
+
+## Idempotency
+- [ ] Idempotency Start:
+  - [ ] `ttl_seconds` -> `lock_ttl`. This is a bit confusing calling it just ttl. I think my suggestion is terrible btw, but we need something that indicates what it is (I don't think it's a timeout either). Has to have a different name than the other ttl in complete.
```
svix-lucho, line 39: Trying to think of a better name

tmp/api-feedback-2026-02-09.md
svix-lucho: Well I guess if `retry_after` is in millis then this should also be in millis

tmp/api-feedback-2026-02-09.md
svix-lucho: Updated for consistent casing

tmp/api-feedback-2026-02-09.md
```diff
@@ -0,0 +1,125 @@
...
+  - [ ] Returning data:
+    - [ ] `retry_after`: isn't that just `refill_interval` always? Unless we support multiple `token` fetches and then it's a bit more complicated potentially. Should this be in seconds or ms if we change refill interval too?
+    - [ ] Here we called it `result` is idempotency we called it `status`. In idempotency value is all lowercase (`completed`) here it's upper case (`BLOCK`). Need to be consistent.
+    - [ ] `remaining` -> `available`? Not sure, just bouncing ideas. Maybe it's fine.
+- [ ] Get remaining:
+  - [ ] Not sure: but do we want `retry_after` here? Given that it's not really a try? It also lets us not pass the config which I thin kis a big win for this function in terms of simplicity?
+- [ ] How do you envision "multi rate-limit" looking like in the future? So letting us limit on a few in parallel?
+- [ ] How do we pass the config around for the rate-limiter? Do we want to init something before, or just pass it live?
+  - [ ] I'm just concerned it may be (1) an implementation detail, and (2) confusing.
```
svix-lucho, line 80: The benefit of not having to create something in advance outweighs this confusion IMO. And as I said above, we can make the API much nicer in the SDK.

Issue #35, opened 2026-01-27
time-to-id log index


svix-jbrown:
There should be a secondary index in the logs database from datetime to log index. This can be relatively sparse (say, written every 1-5 minutes).

PR #117, opened 2026-02-10
raft: port log storage to use TableRow

svix-jbrown:
This ports the `CoyoteLogs` system to use `fjall-utils::TableRow` for storage, as a precursor for storing multiple logical "tables" in this keyspace (for #35).

To do this, I made pretty substantial refactors for TableRow. Instead of the key being `AsRef<u8>`, there's a dedicated `TableKey` trait with provided impls for `String`, `Vec<u8>`, `Uuid`, and `u64`. There's also a related `MonotonicTableKey` trait for types whose serialized byte format is monotonic with their raw format (currently only implemented for big-endian-serialized u64; might be safe for UUID but I need to confirm how the bytes are ordered in its `.as_bytes()` method); implementing this for your `Key` unlocks some new `range`-related methods. Notably, this trait is not implemented for `String`, since the sorting of strings is not the same as the sorting of their utf8-encoded bytestreams.

### DIFF
--- a/crates/fjall-utils/src/table_row.rs
+++ b/crates/fjall-utils/src/table_row.rs
@@ -56,7 +56,7 @@ impl TableKey for u64 {
 
     fn try_from_bytes(bytes: &[u8]) -> Result<Self> {
         if bytes.len() != 8 {
-            return Err(Error::generic("invalid byte length in key key"));
+            return Err(Error::generic("invalid byte length in key"));
         }
         Ok(BigEndian::read_u64(bytes))
     }
--- a/crates/fjall-utils/src/table_row.rs
+++ b/crates/fjall-utils/src/table_row.rs
@@ -1,7 +1,72 @@
-use std::borrow::Cow;
+use byteorder::{BigEndian, ByteOrder};
+use fjall::{Guard, UserKey};
+use std::{
+    borrow::Cow,
+    ops::{Bound, RangeBounds},
+};
+use uuid::Uuid;
 
-use coyote_error::{Error, Result};
+use coyote_error::{Error, Result, ResultExt};
 use serde::{Serialize, de::DeserializeOwned};
+
+/// A trait for primary keys in fjall.
+pub trait TableKey: Clone {
+    fn as_bytes(&self) -> Cow<'_, [u8]>;
+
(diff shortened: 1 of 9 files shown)


crates/fjall-utils/src/table_row.rs
```diff
@@ -1,31 +1,97 @@
...
+impl TableKey for u64 {
+    fn as_bytes(&self) -> std::borrow::Cow<'_, [u8]> {
+        let mut buf = vec![0u8; 8];
+        BigEndian::write_u64(&mut buf, *self);
+        Cow::Owned(buf)
+    }
+
+    fn try_from_bytes(bytes: &[u8]) -> Result<Self> {
+        if bytes.len() != 8 {
+            return Err(Error::generic("invalid byte length in key key"));
```
svix-jplatte: "key key"?

svix-jplatte (commented):
Should probably also be reviewed by sbd. who has worked on the raft logic. Just leaving a small nitpick, can't see anything else that looks off.

crates/fjall-utils/src/table_row.rs
svix-jbrown: the most important key, obviously. :-)

crates/kv/src/tables.rs
```diff
@@ -59,9 +60,15 @@ impl ExpirationKey {
...
+    fn as_bytes(&self) -> Cow<'_, [u8]> {
+        Cow::Borrowed(self.0.as_bytes())
+    }
+
+    fn try_from_bytes(bytes: &[u8]) -> coyote_error::Result<Self> {
+        String::from_utf8(bytes.to_owned())
+            .map(Self)
+            .map_err_generic()
     }
 }
```
svix-lucho, line 73: This one can also be a Monotonic Key, right? I can do it later

svix-lucho (approved):

crates/kv/src/tables.rs
svix-jbrown: You'd need to make sure that `hex::encode(u64.to_be_bytes())` preserves leading zeros so that it's actually lexographically-sortable; I don't know offhand

PR #123, opened 2026-02-10
Add cluster testing to benchmarks

svix-james:


benchmarks/src/lib.rs
```diff
@@ -23,7 +25,7 @@ pub fn setup_server_simple() -> BenchmarkContext {
     let (server, workdir) = rt.block_on(async {
         let workdir = tempfile::tempdir().unwrap();
         let mut cfg = default_server_config(workdir.path());
-        cfg.log_level = LogLevel::Info;
+        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```
svix-james, line 28: Single-node benchmarking improves dramatically w/ this.

crates/test-utils/src/server.rs
```diff
@@ -86,7 +86,7 @@ impl TestServerBuilder {
         let repl_listener = if let Some(listener) = self.repl_listener {
             listener
         } else {
-            TcpListener::bind("127.0.0.1:0").await.unwrap()
+            TcpListener::bind("0.0.0.0:0").await.unwrap()
```
svix-james, line 89: Due to how nodes are discovered, listening on 127.0.0.1 doesn't work for the clustering listener.

benchmarks/src/lib.rs
svix-lucho: Any idea why? I think this needs a comment

svix-lucho (approved):

PR #135, opened 2026-02-13
API changes in rate limiter, kv and idempotency modules

svix-lucho:
Some improvements based on Tom's feedback on https://github.com/svix/coyote-private/pull/116

Summary:
- Rename `expire_in` to `ttl`
- Remove units in field names
- Rename idempotency/abandon to idempotency/abort
- Use consistent casing in enums (snake_case)
- Return `Vec<u8>` instead of `String` in cache
- Add one more status to idempotency instead of returning an HTTP error when the key is locked
- A few other misc field renames

### DIFF
--- a/src/v1/endpoints/cache.rs
+++ b/src/v1/endpoints/cache.rs
@@ -28,19 +28,18 @@ pub struct CacheSetIn {
     pub key: EntityKey,
 
     /// Time to live in milliseconds
-    pub expire_in: u64,
+    pub ttl: u64,
 
-    // FIXME: change to Bytes
-    pub value: String,
+    pub value: Vec<u8>,
 }
 
 impl CacheSetIn {
     fn into_model(self) -> CacheModel {
-        let expires_at = Timestamp::now() + Duration::from_millis(self.expire_in);
+        let expires = Timestamp::now() + Duration::from_millis(self.ttl);
 
         CacheModel {
-            expires_at: Some(expires_at),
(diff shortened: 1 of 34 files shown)


src/v1/endpoints/cache.rs
```diff
@@ -28,19 +28,18 @@ pub struct CacheSetIn {
     pub key: EntityKey,
 
     /// Time to live in milliseconds
-    pub expire_in: u64,
+    pub ttl: u64,
 
-    // FIXME: change to Bytes
-    pub value: String,
+    pub value: Vec<u8>,
```
svix-jplatte, line 33: Still need to figure out how to make this not totally awkward in JSON.

src/v1/endpoints/kv.rs
```diff
@@ -73,7 +73,7 @@ pub struct KvGetOut {
     pub key: Arc<EntityKey>,
 
     /// Time of expiry
-    pub expires_at: Option<Timestamp>,
+    pub expires: Option<Timestamp>,
```
svix-jplatte: Do you strongly prefer this of `expiry`? The verb on its own looks weird to me.

svix-jplatte (approved):
Mostly looks good to me, two small comments.

src/v1/endpoints/kv.rs
svix-lucho: I don't have strong opinions on this one. I changed it to `expiry`

Issue #124, opened 2026-02-10
Don't use colons in env vars


svix-james:
I think having env vars that shell interpreters don't like is bad UX. Let's have a `ConfigurationIn` or something that's got a flattened config, and we can map that to our internal config on startup.

svix-jplatte:
Grooming:

- Rip out config crate, roll our own env var parsing without serde

PR #151, opened 2026-02-19
Remove unused JWT code

svix-jplatte:
We may reintroduce it later, but not using the jwt-simple crate.

Resolves the `FIXME` in `security.rs` and simplifies the config work a little bit (https://github.com/svix/coyote-private/issues/124).


svix-gabriel (approved):
Waow

PR #154, opened 2026-02-19
Add JS SDK

svix-jplatte:
Have verified only that it builds, for now. Needs CI + some basic (manual) testing soon, but I think there's no harm in having this reviewed and merged as-is.

Best reviewed commit by commit.

### DIFF
--- a/clients/javascript/src/apis/kv.ts
+++ b/clients/javascript/src/apis/kv.ts
@@ -1,0 +1,111 @@
+// this file is @generated
+
+import {
+    type KvDeleteIn,
+    KvDeleteInSerializer,
+} from '../models/kvDeleteIn';
+import {
+    type KvDeleteOut,
+    KvDeleteOutSerializer,
+} from '../models/kvDeleteOut';
+import {
+    type KvGetIn,
+    KvGetInSerializer,
+} from '../models/kvGetIn';
+import {
+    type KvGetOut,
+    KvGetOutSerializer,
+} from '../models/kvGetOut';
+import {
+    type KvSetIn,
+    KvSetInSerializer,
+} from '../models/kvSetIn';
+import {
+    type KvSetOut,
+    KvSetOutSerializer,
+} from '../models/kvSetOut';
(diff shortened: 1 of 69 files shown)


clients/javascript/src/apis/kv.ts
```diff
@@ -0,0 +1,111 @@
...
+        idempotencyKey?: string;
+        }
+
+    export class Kv {
+    public constructor(private readonly requestCtx: CoyoteRequestContext) {}
+
+    /** KV Set */
+        public set(
+            kvSetIn: KvSetIn,
+            options?: KvSetOptions,
```
tasn, line 47: Is this what we decided with options? I remember we had a discussion, but I felt like this was not necessarily the option we were gunning for?

clients/javascript/src/models/cacheGetOut.ts
```diff
@@ -0,0 +1,30 @@
...
+
+
+
+
+
+export interface CacheGetOut {
+    /** Time of expiry */
+    expiry?: Date | null;
+key: string;
+value: number[];
```
tasn, line 11: Needs some formatting? I think we run a separate formatter in Svix?

clients/javascript/src/apis/kv.ts
svix-jplatte: There hasn't been a decision, this is just what the Svix templates do. I presented multiple options in https://github.com/svix/rfc/pull/43 and nobody other than you commented, I left a few responses but that was it as of right now.

Updating the codegen templates will be easy though, once we make a decision.

clients/javascript/src/models/cacheGetOut.ts
svix-jplatte: Yeah, we run biome for formatting and linting in Svix. It's something I want to add here too, just focusing on having something that works first. (it's not a lot of work but it works a little differently here compared to svix, as we use openapi-codegen as a Cargo dependency, not the containerized thing)

svix-james (approved):

PR #157, opened 2026-02-19
Change `ConsumerGroup` from type alias to validated type

svix-gabriel:
Planned to do this for a while.

The validations we run on the ConsumerGroup are completely arbitrary right now. I just picked something that seemed reasonable at first glance. It's trivial to change in the future.


svix-gabriel:
CI is failing on autogenerated files... I am going to become the joker

svix-jbrown (approved):

PR #159, opened 2026-02-19
Add directive to install podman (needed for cargo codegen

svix-gabriel:
I couldn't run `cargo codegen` locally. Apparently I needed to install podman.

Then I ran `cargo codegen`.

Also spotted what I assumed was a bug where an error in cargo codegen wasn't getting reported correctly.


codegen/src/formatting.rs
```diff
@@ -91,7 +91,7 @@ async fn exec(cmd: &str, args: impl IntoIterator<Item = &str>) -> io::Result<()>
         .await
         .map_err(|e| {
             let kind = e.kind();
-            let msg = format!("failed to run {cmd}: e");
+            let msg = format!("failed to run {cmd}: {e:?}");
```
svix-jplatte, line 94: So this improved the error message? What's the before / after?

PR #163, opened 2026-02-20
Replace config crate with custom env var parsing

svix-jplatte:
… for the main config file.

Part of https://github.com/svix/coyote-private/issues/124.


svix-james (approved):
LGTM 😎

Issue #77, opened 2026-02-03
Figure out how to do a joined snapshot of fjall DBs


svix-jplatte:
We have two fjall DBs, persisted and ephemeral. For the clustering protocol, we need to be able to create a snapshot that covers both.

tasn:
A couple of ideas:
1. Have a separate Raft for each (JB mentioned this is potentially complex, so maybe not).
2. ~We only have two DB threads, one for persisted and one for ephemeral. We can send them both a message / flip a bool to tell them to stop what they are doing for a sec and do the snapshot.~ ACTUALLY: we only have one thread writing the rafts to both ephemeral and persisted; as we go over the Raft log sequentially so we can't really split this work. This makes it trivial to send the thread a message to pause the work, snapshot both, and resume the work.

svix-jbrown:
I think this works well enough, although none of the modules except Stream actually use it; see #260 for that

PR #168, opened 2026-02-21
first pass at multi-db cluster support

svix-jbrown:
This is the first (and less likely to conflict) PR related to supporting both the ephemeral and the persistent databases in replication.

It stores references to both of these behind an RwLock inside the StateMachine, and when creating/restoring snapshots, holds a write lock on that RwLock.

This also builds out the infrastructure for providing read-only views across the databases.

In a future set of PRs, I'm going to be expanding on this to make it so that instead of the DBs living in `coyote_configgroup`, they live in the state machine and get _borred_ by coyote_configgroup; for `ReadonlyDatabases` we don't need to hold the rwlock at all (which is why that object impls Clone), but for `Databases` we do.

Part of #77

### DIFF
--- a/src/core/cluster/network.rs
+++ b/src/core/cluster/network.rs
@@ -69,7 +69,7 @@ impl NetworkClient {
         let start = Instant::now();
         // TODO(jbrown|2026-02-20) handle multiple addresses
         let addrs = self.node.addrs();
-        let Some(addr) = addrs.get(0) else {
+        let Some(addr) = addrs.first() else {
             tracing::warn!(node_id=?self.target, node=?self.node, "node has no valid addresses, cannot send rpc");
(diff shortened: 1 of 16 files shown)


src/core/cluster/network.rs
```diff
@@ -69,7 +69,7 @@ impl NetworkClient {
         let start = Instant::now();
         // TODO(jbrown|2026-02-20) handle multiple addresses
         let addrs = self.node.addrs();
-        let Some(addr) = addrs.get(0) else {
+        let Some(addr) = addrs.first() else {
```
svix-james, line 72: Praise be to Clippy, for He is wise and all-knowing

svix-james (approved):

PR #173, opened 2026-02-23
Prefactor: rename stream -> stream-deprecated

svix-gabriel:
Split off from a larger change.

We have to make a bunch of API changes as outlined in https://github.com/svix/rfc/pull/47 + https://github.com/svix/rfc/pull/44.

A decent chunk of the core implementation is going to stay the same, but I think it'll end up being easier to implement and review if we can separate the old stuff from the new, and then delete the old stuff.

There's a (for now) empty `stream-internals` crate. We're gonna have `queue`, `stream`, and eventually something called `fifo` (name tbd). The API and semantics will be slightly different, but they'll share a lot of bones so that's what `stream-internals` is for.


svix-jbrown (approved):

svix-james:
Does this mean I shouldn't be trying to write benchmarks against the APIs? :melting_face:

svix-gabriel:
@svix-james I'd recommend prioritizing non-stream APIs for now, but if Stream is all that's left I can port the benchmarks to the new endpoints.

I won't delete the old endpoints until the new ones are in place, so hopefully there won't be any merge conflicts

PR #174, opened 2026-02-23
server: use non-colon form of env vars in development.env

svix-jbrown:
these have been broken since commit bfecef7a0d5204f0326da54d8a6ff0f616d311f7


svix-jplatte (approved):
Sorry!

Issue #176, opened 2026-02-24
Annotate input struct fields as positional, for SDKs


svix-jplatte:
For the SDKs, we want to have some body fields passed as positional parameters, rather than as fields of the main parameter object.
To be able to do this, we first have to encode this information the OpenAPI spec somehow.

svix-jplatte:
Not everything is annotated, but https://github.com/svix/coyote-private/pull/181 shows how to do it (it's quite trivial).

PR #181, opened 2026-02-24
cache: Mark some input parameters as positional

svix-jplatte:
Only does a single module, but this is so trivial to add for others once decided which fields should be positional, I consider this as closing https://github.com/svix/coyote-private/issues/176.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -1322,6 +1322,9 @@
         },
         "required": [
           "key"
+        ],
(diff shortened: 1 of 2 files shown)


tasn (approved):

svix-jplatte:
This is incomplete, but I'll merge now to not have to switch branches / rebase. Shouldn't hurt to have this in.

PR #192, opened 2026-02-24
Remove some unwraps and such

svix-james:
RE #144 

There's still some, but I don't think they're egregious. 
Mostly startup-type things.

### DIFF
--- a/src/core/cluster/logs.rs
+++ b/src/core/cluster/logs.rs
@@ -278,14 +278,17 @@ impl CoyoteLogs {
         let last_purged_log_id = LAST_PURGED_LOG_ID.get(&self.meta_keyspace);
         let first_id = Log::range(&self.log_keyspace, ..)
             .next()
-            .map(|l| l.unwrap().0);
+            .and_then(|l| l.ok())
+            .map(|(k, _)| k);
         let last_id = Log::range(&self.log_keyspace, ..)
             .next_back()
-            .map(|l| l.unwrap().0);
+            .and_then(|l| l.ok())
+            .map(|(k, _)| k);
         tracing::trace!(?last_purged_log_id, first_id, last_id, "log metadata");
         for row in Log::range(&self.log_keyspace, ..) {
-            let (index, value) = row.unwrap();
-            tracing::trace!(?index, ?value, "log");
+            if let Ok((index, value)) = row {
+                tracing::trace!(?index, ?value, "log");
+            }
         }
         tracing::trace!("END LOG TRACE");
     }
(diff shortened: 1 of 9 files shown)


src/core/cluster/logs.rs
```diff
@@ -278,18 +278,22 @@ impl CoyoteLogs {
...
-            .map(|l| l.unwrap().0);
+            .and_then(|l| l.ok())
+            .map(|(k, _)| k);
         tracing::trace!(?last_purged_log_id, first_id, last_id, "log metadata");
         for row in Log::range(&self.log_keyspace, ..) {
-            let (index, value) = row.unwrap();
-            tracing::trace!(?index, ?value, "log");
+            if let Ok((index, value)) = row {
+                tracing::trace!(?index, ?value, "log");
+            }
```
svix-jbrown: you can actually delete this whole method; it was only used for debugging

src/core/cluster/state_machine.rs
```diff
@@ -87,18 +87,15 @@ pub struct Store {
...
-    fn snapshot_idx(&self) -> u64 {
-        // TODO: fix these unwraps
+    fn snapshot_idx(&self) -> Option<u64> {
         self.snapshot_id
             .split('-')
             .next_back()
-            .unwrap()
-            .parse()
-            .unwrap()
+            .and_then(|s| s.parse().ok())
```
svix-jbrown: this one I think should stay as an unwrap or expect, because if we hit it we need to shit down the server because it means something is corrupt in our snapshots. If you would prefer to make it return `anyhow::Result<u64>` and make the caller propagate/panic it, I'm fine with that too.

svix-james: I assume that `TODO` was written w/ noble intentions, so let's not have this panic.

svix-jplatte (dismissed):

src/core/cluster/state_machine.rs
```diff
@@ -212,8 +212,7 @@ impl Store {
         self.last_membership = last_membership;
         self.snapshot_idx = last_snapshot
             .as_ref()
-            .map(|s| s.meta.snapshot_idx())
-            .unwrap_or(0);
+            .map_or(Ok(0), |s| s.meta.snapshot_idx())?;
```
svix-jbrown, line 204: sorry, what I had been suggesting was propagating a parse failure here; something like

```
self.snapshot_idx = last_snapshot
  .as_ref()
  .map(|s| s.meta.snapshot_idx())
  .transpose()?
  .unwrap_or(0)
```

svix-james: I'm confused -- is there a difference here? The `snapshot_idx` function will propagate a parse failure, yes?

svix-jbrown: oh, you're right, I misread the `map_or` docs.

svix-jbrown (approved):

src/core/cluster/state_machine.rs
svix-james: I was trying hard to avoid using a `transpose`.... which always confuse my idiot brain.

PR #204, opened 2026-02-25
Cluster: support having hostnames in seed_nodes.

### LINKED ISSUE #199: Cluster: support bootstrapping from hostnames, not just addreses
This is required for setups like docker compose, and potentially k8s where it's easier to just pass the hostnames than specific IPs.

Open questions:
- Should we support peers to share hostnames at discovery, or at that point should we only share IPs?
- Should we ever resolve them *ahead of time* or just always use DNS when those are passed (as we make HTTP calls anyway). Extra relevant because there may be a load balancer or something in front of it that has TLS or expects a hostname for a different reason.

tasn:
This PR only resolves hostnames from the original config, but doesn't resolve hostnames when shared, nor does it allow it.

Partially fixes #199, though there are still open questions.

TBH, I feel like the other solutions suggested in #199 are better.

### DIFF
--- a/benchmarks/src/lib.rs
+++ b/benchmarks/src/lib.rs
@@ -1,6 +1,6 @@
 #![allow(unused)]
 
-use coyote::{cfg::LogLevel, core::cluster::proto::HealthResponse};
+use coyote::{cfg::{LogLevel, SeedNode}, core::cluster::proto::HealthResponse};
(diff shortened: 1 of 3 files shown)


tasn:
Closing this as @svix-jbrown is taking a different approach.

Issue #128, opened 2026-02-12
Stream/Queue API: implement changes from review


svix-gabriel:
Implement all of the API changes described here https://github.com/svix/coyote-private/pull/116/changes#diff-a4337f70b22371ac15e35d5883ed36faec47388f1860f986f7e4833ced76359eR83

(Taking into account the comments)

tasn:
Is this done?

PR #205, opened 2026-02-25
Change `stream.create` -> `msgs.topic.create`

svix-gabriel:
Part of https://github.com/svix/coyote-private/issues/128

One of the api changes we are making is changing `stream.create` -> `msgs.topic.create`, with a slightly different input format

```
{
   "persistence": "ephemeral", // Similar to other modules, need the naming there.
   "retention": {
      "bytes": 4_194_304,
      "millis": 604_800
   }
 }
```

As outlined [here](https://github.com/svix/rfc/pull/47/changes#diff-2af67d20315233a3c2bb2fbb26acff6ffbb954a1b61e32dd78e3f9039ae50c51R41).

The other changes are outlined [here](https://github.com/svix/rfc/pull/44/changes#diff-f460d68a82fb139a4c3c165c91ccc41c23d9fccadcd1289ff7f96c0b12f8033cR6), but in summary

* Routes are now organized under `msgs`/`topics`
* `created_at` / `updated_at` -> `created` / `updated`
* Retention is in a single object
* Persistence is configurable

More PRs for more endpoints to come.


crates/fjall-utils/src/duration_millis.rs
```diff
@@ -0,0 +1,12 @@
+use std::time::Duration;
```
svix-jbrown, line 1: I... guess? This doesn't really feel very fjall-related and seems like it could just as easily go somewhere else.

svix-jbrown (dismissed):

svix-james (approved):

crates/fjall-utils/src/duration_millis.rs
svix-gabriel: Possibly. I originally put it here because its used when serializing the struct into fjall, as oppose to our public API (though I guess it could be used there too).

I think about where else this can go

PR #207, opened 2026-02-25
cluster: allow adding peers with hostnames instead of SocketAddrs

### LINKED ISSUE #199: Cluster: support bootstrapping from hostnames, not just addreses

svix-jbrown:
This adds hostname support throughout the stack (in the config, at discovery time, and at actual replication time). In a production config, I'd want to use SocketAddrs to avoid having DNS as a piece that could break, but if you need hostnames, you need hostnames.

I intentionally did not go the route of resolving the hostname at discovery time and storing the SocketAddr in the Raft membership because I think that violates the principle of least surprise, and would be bad for users when DNS mappings change (which can even happen in docker-compose around a `down` + `up`).

I'm keeping the `Node` object (which gets serialized into logs) as an explicitly-tagged enum, to minimize potential confusion, but adding a new `PeerAddr` type that's used in config and discovery which is just greedily parsed from a string. We use the popular [addr](https://crates.io/crates/addr) crate to validate that the name portion is a valid domain name.

Fixes #199

### DIFF
--- a/src/cfg/mod.rs
+++ b/src/cfg/mod.rs
@@ -22,6 +22,78 @@ mod defaults;
 
 pub type Configuration = Arc<ConfigurationInner>;
 
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub enum PeerAddr {
+    SocketAddr(SocketAddr),
+    HostnameAndPort { hostname: String, port: u16 },
+}
+
+impl PeerAddr {
+    pub fn as_base_url(&self) -> url::Url {
+        let base = self.to_string();
+        format!("http://{base}")
+            .parse()
+            .expect("we validated this already")
+    }
+}
+
+impl std::fmt::Display for PeerAddr {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::SocketAddr(s) => s.fmt(f),
+            Self::HostnameAndPort { hostname, port } => write!(f, "{hostname}:{port}"),
+        }
+    }
+}
+
+impl From<SocketAddr> for PeerAddr {
+    fn from(value: SocketAddr) -> Self {
+        Self::SocketAddr(value)
+    }
+}
+
+impl FromStr for PeerAddr {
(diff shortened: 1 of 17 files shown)


src/cfg/mod.rs
```diff
@@ -98,8 +170,15 @@ pub struct ClusterConfiguration {
     #[serde(default = "defaults::cluster_election_timeout_max_ms")]
     pub election_timeout_max_ms: u64,
 
+    /// Address that other nodes should use to communicate with this one. If not passed, we'll
+    /// attempt to discover it at boot time. This cannot currently be changed after cluster
+    /// initialization.
     #[serde(default)]
-    pub seed_nodes: Vec<SocketAddr>,
+    pub my_address: Option<PeerAddr>,
```
svix-james: Looks good overall, but I don't love `my_address` as a config param name. This is effectively the same as "advertised" name in Kafka parlance, right? Maybe we should use something like that instead.

svix-jbrown: I like that idea

src/cfg/mod.rs
```diff
@@ -371,6 +451,9 @@ fn load_toml(config_toml: Option<&str>) -> anyhow::Result<Arc<ConfigurationInner
     if let Some(value) = env_var("COYOTE_BOOTSTRAP_CFG_PATH")? {
         *bootstrap_cfg_path = Some(value);
     }
+    if let Some(value) = env_var("COYOTE_MY_ADDRESS")? {
```
svix-james:
```suggestion
    if let Some(value) = env_var("COYOTE_ADVERTISED_ADDRESS")? {
```

svix-james (approved):

PR #223, opened 2026-02-26
cli: Fix config loading

svix-jplatte:


clients/cli/src/main.rs
```diff
@@ -149,7 +149,7 @@ fn get_client(cfg: &Config) -> Result<CoyoteClient> {
 fn get_client_options(cfg: &Config) -> Result<CoyoteOptions> {
     Ok(CoyoteOptions {
         debug: false,
-        server_url: cfg.server_url().map(Into::into),
+        server_url: dbg!(cfg.server_url().map(Into::into)),
```
svix-gabriel: Is this deliberate?

svix-jplatte: Nope, did not intend to keep this in. Thanks for catching.

svix-gabriel (approved):

PR #225, opened 2026-02-26
cluster: support periodic snapshotting and log retention

### LINKED ISSUE #36: time-based log retention
We should be making sure to call [`purge_log`](https://docs.rs/openraft/latest/openraft/raft/trigger/struct.Trigger.html#method.purge_log) periodically to make sure that we're not retaining logs indefinitely even if we aren't doing leadership changes.

This depends on #35

svix-jbrown:
This allows the user to configure the database to take snapshots periodically (either after a certain number of writes or a certain time period), and to truncate logs that are included by a more recent snapshot.

Fixes #36


svix-james (dismissed):

svix-james:
but ze conflicts

svix-james (approved):

PR #231, opened 2026-02-26
more codegen lint stuff

svix-jbrown:
This runs `just codegen lint` on main, fixes all the issues, and changes `cargo codegen` to also run clippy with the same args as `just lint` so they won't fight any more. This is very dumb.


svix-gabriel (dismissed):
:sigh: This is gonna cause more merge conflicts for me, but yeah let's just do it

svix-gabriel (approved):

svix-jplatte:
I strongly disagree about codegen running clippy. We should either make the templates not emit problematic code, or if too annoying silence the relevant lints for the generated modules.

PR #241, opened 2026-02-27
dev: do not run clippy on coyote-cli or coyote-client

svix-jbrown:
This is ugly, but it changes both CI and the `justfile` to only run clippy on the non-codegen'd packages, so that a future codegen issue won't result in inconsistent behavior for everyone.

This depends on [cargo-workspaces](https://github.com/pksunkara/cargo-workspaces), because there doesn't appear to be a first-party way to get a list of all crates in a workspace, which is kind of weird.


svix-jplatte:
As I said before, I don't think we should. If clippy causes issues for generated code, we can either fix the templates or add allow attributes. Really no different to non-generated code.

Issue #122, opened 2026-02-10
Add first batch of Coyote SDKs


svix-jplatte:
For now we only have Rust + CLI. Add all the languages we have in Svix.

svix-jplatte:
Next up: (after Rust, Go, JS) - Java, Python [both sync, async]

svix-jplatte:
Stopping for now and working on higher priority stuff. Will create new tickets for remaining SDKs.

PR #244, opened 2026-02-27
java: Add Coyote class

svix-jplatte:
Missed this originally because it's hand-written for Svix.

Part of https://github.com/svix/coyote-private/issues/122.

### DIFF
--- a/codegen/templates/java/api_summary.java.jinja
+++ b/codegen/templates/java/api_summary.java.jinja
@@ -1,0 +1,45 @@
+{% do set_summary_filename("Coyote.java") -%}
+// this file is @generated
+package com.svix.coyote;
+
+import java.util.Map;
+import java.util.TreeMap;
+
+{% for resource in api.resources -%}
+import com.svix.coyote.apis.{{ resource.name | to_upper_camel_case }};
+{% endfor %}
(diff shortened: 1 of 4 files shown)


codegen/templates/java/api_summary.java.jinja
```diff
@@ -0,0 +1,45 @@
...
+            options.setServerUrl(CoyoteOptions.DEFAULT_URL);
+        }
+
+        HttpUrl parsedUrl = HttpUrl.parse(options.getServerUrl());
+        if (parsedUrl == null) {
+            throw new IllegalArgumentException("Invalid base url");
+        }
+
+        Map<String, String> defaultHeaders = new TreeMap<>();
+        defaultHeaders.put("user-agent", "coyote-libs/0.1.0/java");
```
svix-gabriel, line 31: Should we start parameterizing the version we put here?

svix-gabriel (approved):

codegen/templates/java/api_summary.java.jinja
svix-jplatte: Yeah, the version should also not be hardcoded. I'll make some tickets.

PR #247, opened 2026-02-27
python: Switch from JSON to msgpack

svix-jplatte:
Part of https://github.com/svix/coyote-private/issues/122.


clients/python/coyote/internal/api_common.py
```diff
@@ -198,15 +143,20 @@ def _request_sync(
...
-def _filter_response_for_errors_response(response: httpx.Response) -> httpx.Response:
+def _parse_response[T: BaseModel](
+    response: httpx.Response,
+    response_type: t.Optional[type[T]],
+) -> t.Optional[T]:
     if 200 <= response.status_code <= 299:
-        return response
+        if response_type is not None:
+            response_decoded = msgpack.unpackb(response.content)
+            return response_type.model_validate(response_decoded)
```
svix-jbrown, line 156: future improvement: we should actually check the content-type of the response; right now, the user might get a garbage msgpack crash if they get back, say, a `text/plain` response from a load balancer.

svix-jbrown (approved):

PR #248, opened 2026-02-27
Add `msgs.publish` 

svix-gabriel:
This is the implementation of `coyote.msgs.publish`.

It's very analogous to the current stream.append implementation, with the caveat of managing partitions.

We'll clean up the old `stream.append` implementation later. Trying to DRY things up now would be a hassle when the old code is going to get nuked anyway.


svix-jplatte:
Sorry, just merged a codegen template change :melting_face:

crates/msgs/src/operations/publish.rs
```diff
@@ -0,0 +1,99 @@
...
+
+use super::{MsgsRaftState, MsgsRequest, PublishResponse};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct PublishOperation {
+    namespace_id: NamespaceId,
+    topic: String,
+    msgs: Vec<MsgIn>,
+    /// Partition assigned to messages without a key. Chosen randomly by the
+    /// caller so the Raft state machine stays deterministic.
```
svix-jplatte:
```suggestion
    /// Partition assigned to messages without a key.
    ///
    /// Chosen randomly by the caller so the Raft state machine stays deterministic.
```

svix-jplatte (dismissed):
I know basically nothing about the whole fjall + raft side of things, so this is only a code style review. If you think that's good enough, go ahead. (I've left a small nitpick so this PR doesn't auto-merge before this is read :smile:)

svix-gabriel:
I think that's good enough. The raft/integration here isn't any different than the extant stream code, which folks who were more familiar with that stuff previously reviewed

svix-jplatte (approved):

PR #252, opened 2026-02-27
Add `coyote.msgs.stream.receive(...)`

svix-gabriel:
The logic here isn't perfect, but it'll be easier to correct (and show correct behavior in the tests) once `commit-offset` is in (coming in a followup PR).

### DIFF
--- a/crates/msgs/src/operations/stream_receive.rs
+++ b/crates/msgs/src/operations/stream_receive.rs
@@ -1,0 +1,180 @@
+use std::num::NonZeroU16;
+
+use coyote_namespace::entities::NamespaceId;
+use jiff::Timestamp;
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    State,
+    entities::{
+        ConsumerGroup, DEFAULT_PARTITION_COUNT, Offset, PartitionIndex, partition_topic_name,
+    },
+    tables::{LeaseDiff, LeaseRow, MsgRow, OffsetRow},
+};
+
+use super::{MsgsRaftState, MsgsRequest, StreamReceiveResponse};
+
+fn group_into_contiguous_ranges(offsets: &[Offset]) -> Vec<(Offset, Offset)> {
+    if offsets.is_empty() {
+        return Vec::new();
+    }
+
+    let mut ranges = Vec::new();
+    let mut range_start = offsets[0];
+    let mut range_end = offsets[0];
+
+    for &offset in &offsets[1..] {
+        if offset == range_end + 1 {
+            range_end = offset;
+        } else {
+            ranges.push((range_start, range_end));
+            range_start = offset;
+            range_end = offset;
+        }
+    }
+    ranges.push((range_start, range_end));
+
+    ranges
+}
+
+fn create_leases_for_msgs(
+    offsets: &[Offset],
+    namespace_id: NamespaceId,
+    partition: PartitionIndex,
+    cg: ConsumerGroup,
+    now: Timestamp,
+    lease_duration: std::time::Duration,
+    lease_diff: &mut LeaseDiff,
+) {
+    let ranges = group_into_contiguous_ranges(offsets);
+    for (block_start, block_end) in ranges {
+        lease_diff.to_insert.push(LeaseRow {
+            namespace_id,
+            partition,
+            cg: cg.clone(),
+            block_start,
+            block_end,
+            leased_at: now,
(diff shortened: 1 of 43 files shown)


crates/msgs/src/operations/stream_receive.rs
```diff
@@ -0,0 +1,179 @@
...
+        let lease_duration = std::time::Duration::from_millis(self.lease_duration_millis);
+        let mut all_msgs = Vec::new();
+        let mut all_lease_diffs = Vec::new();
+        let mut remaining = usize::from(self.batch_size.get());
+
+        for p in 0..DEFAULT_PARTITION_COUNT {
+            if remaining == 0 {
+                break;
+            }
+            let partition = PartitionIndex::new(p).unwrap();
```
svix-jplatte: We just got rid of most unwraps a day or two ago. Should probably not introduce new ones?

crates/msgs/src/operations/stream_receive.rs
```diff
@@ -0,0 +1,179 @@
...
+            let start_offset = OffsetRow::fetch(state, self.namespace_id, partition, &self.cg)?
+                .unwrap_or(Offset::MIN);
+
+            let leases = LeaseRow::fetch_all(state, self.namespace_id, partition, &self.cg)?;
+
+            let blocked_leases = leases
+                .iter()
+                .filter(|l| l.acked_at.is_some() || l.is_dlq() || l.is_active(now));
+
+            let batch_size = std::num::NonZeroUsize::new(remaining).unwrap();
```
svix-jplatte: Here too.

crates/msgs/src/operations/stream_receive.rs
```diff
@@ -0,0 +1,179 @@
...
+            }
+
+            all_lease_diffs.push(lease_diff);
+        }
+
+        let mut batch = state.db.batch();
+        for diff in all_lease_diffs {
+            diff.apply_diff(state, &mut batch)?;
+        }
+        batch.commit().map_err(coyote_error::Error::from)?;
```
svix-jplatte, line 156: Uselsss `map_err`.

src/v1/endpoints/msgs.rs
```diff
@@ -152,6 +154,73 @@ async fn publish(
...
+#[aide_annotate(op_id = "v1.msgs.stream.receive")]
+async fn stream_receive(
+    State(state): State<AppState>,
+    Extension(repl): Extension<RaftState>,
+    MsgPackOrJson(data): MsgPackOrJson<StreamReceiveIn>,
+) -> Result<MsgPackOrJson<StreamReceiveOut>> {
+    let namespace = state
+        .namespace_state
+        .fetch_stream_namespace(&data.name)?
+        .ok_or_else(|| Error::http(HttpError::not_found(None, None)))?;
```
svix-jplatte, line 198: We have an `ok_or_not_found` helper here like in webhooks, no?

svix-gabriel: I don't think so...

crates/msgs/src/operations/stream_receive.rs
svix-gabriel: That's what I thought, but this is the error I get locally if I remove it
```
`?` couldn't convert the error to `OperationError`
the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
the trait `From<fjall::Error>` is not implemented for `OperationError`
but trait `From<coyote_error::Error>` is implemented for it
for that trait implementation, expected `coyote_error::Error`, found `fjall::Error`
```

which admittedly doesn't make much sense to me

crates/msgs/src/operations/stream_receive.rs
svix-gabriel: Replaced this one with an `.expect(...)`. The input's been validated already so I think it's more appropriate here than typical `?` error handling.

crates/msgs/src/operations/stream_receive.rs
svix-jplatte: Oh the function returns `coyote_operations::Result`, not `coyote_error::Result`. It does make sense then.

src/v1/endpoints/msgs.rs
svix-jplatte: Okay, should add it (not in this PR of course).

svix-jplatte (approved):

PR #256, opened 2026-02-27
Add simple cli `benchmark` command

svix-james:
It can be run with configurable number of iterations and modules to be run against. Right now idempotency/ rate-limiting are not supported because there are sdk issues with them.

### DIFF
--- a/clients/cli/src/cmds/benchmark.rs
+++ b/clients/cli/src/cmds/benchmark.rs
@@ -247,17 +247,17 @@ trait BenchShard {
 // KV module
 
 #[derive(Clone)]
-struct TomBenchKvSet {
+struct BenchKvSet {
     keys: Arc<Vec<String>>,
 }
 
-impl TomBenchKvSet {
+impl BenchKvSet {
     fn setup(keys: Arc<Vec<String>>) -> Self {
         Self { keys }
     }
 }
 
-impl BenchShard for TomBenchKvSet {
+impl BenchShard for BenchKvSet {
     fn iterations(&self) -> u64 {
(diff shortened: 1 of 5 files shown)


clients/cli/src/cmds/benchmark.rs
```diff
@@ -0,0 +1,537 @@
...
+
+        // Start of real code
+        let t = quanta::Instant::now();
+        let mut recv = StreamReceiveIn::new(
+            consumer_group.to_owned(),
+            ns_name.to_owned(),
+            topic.to_owned(),
+        );
+        recv.batch_size = Some(1);
+        let out = client.msgs().stream().receive(recv).await?;
```
tasn, line 510: Returns a partition locked error when running:
```
cargo run -- benchmark --modules msgs --concurrency 2 --iterations 1000
```

svix-gabriel: What do you want the behavior here to be? Return an empty set of messages, or something else?

tasn: Not sure I understand the question, I'm consuming from two different *UNRELATED* topics, nothing to do with one another, I don't know why it was even blocking?

Each concurrent test has its own topic.

tasn:
I'm closing this as I'll create a new one.

clients/cli/src/cmds/benchmark.rs
svix-gabriel: Oh if they're consuming from different topics, that's definitely a bug then.

I'll look into it

Issue #187, opened 2026-02-24
Start w/ some basic infra for Coyote


svix-james:
* New AWS account(s)?
* EKS from start
* Terraform

ashayshub:
Discussion [thread](https://svixhq.slack.com/archives/C01U9M7H9DF/p1772208911570209)

ashayshub:
We have infra for deploying to svix-cloud (dev+staging) in place.

PR #262, opened 2026-02-27
feat(tf): base infra for coyote

ashayshub:
This PR creates the base infra for coyote eks.

- Creates the coyote networking stack
- eks cluster
Relates to: #187


terraform/infra/modules/eks/cluster.tf
```diff
@@ -0,0 +1,209 @@
...
+#tfsec:ignore:aws-eks-no-public-cluster-access
+#tfsec:ignore:aws-eks-no-public-cluster-access-to-cidr
+module "eks" {
+  source  = "terraform-aws-modules/eks/aws"
+  version = "~> 21.15"
+
+  name                    = local.eks_cluster_name
+  kubernetes_version      = var.k8s_version
+  endpoint_private_access = true
+  endpoint_public_access  = true
```
svix-james, line 21: Can we avoid this? What do we do in the clickhouse clusters?

terraform/infra/modules/observability/metrics-server.tf
```diff
@@ -0,0 +1,39 @@
+resource "helm_release" "metrics_server" {
```
svix-james: I haven't looked into this closely, but can we just do this?

https://github.com/svix/monorepo-private/blob/main/terraform-eks/modules/eks_core_addons/aws_addons.tf#L100C11-L100C24

terraform/infra/outputs.tf
```diff
@@ -0,0 +1,131 @@
...
+
+output "database_subnet_ids" {
+  value = module.net.database_subnet_ids
+}
+
+output "eks_subnet_ids" {
+  value = module.net.eks_subnet_ids
+}
+
+# eks
```
svix-james, line 52: Why commented out?

terraform/infra/providers.tf
```diff
@@ -0,0 +1,38 @@
...
+
+provider "helm" {
+  kubernetes = {
+    host                   = module.eks.cluster_endpoint
+    cluster_ca_certificate = base64decode(module.eks.cluster_certificate_authority_data)
+    token                  = data.aws_eks_cluster_auth.this.token
+  }
+}
+
+# provider "datadog" {
```
svix-james: Same

terraform/infra/modules/eks/cluster.tf
ashayshub: We can avoid it. Should I enable just the private access in this PR itself or a separate one that includes the tailscale config ? Fyi: It can be disabled later.

terraform/infra/providers.tf
ashayshub: I don't have the valid datadog creds for dev.

terraform/infra/modules/eks/cluster.tf
svix-james: Ok, we can do it later. Let's just make sure it's on the list.

svix-james:
Ok, some things from actually reviewing the infra in AWS
* Let's ditch the pet names -- we already have an animal name for the project, adding another random animal name on there is confusing :laughing:. If we really need this, can we just add a few random [a-z] characters instead?
* There are 12 subnets.... I guess maybe we do need that many. It just seems insane.
* What is the difference between `svix` node group and `coyote` node group?
* Do we need modules nested under `infra` folder? Can there not just be a top-level `modules` folder under `terraform`? TBH I'm not clear if that would make more sense or not.
* I honestly don't like the name `edge` for the module. I'm not exactly sure what it means. Maybe `lb` or `ingress` instead?

terraform/infra/modules/eks/cluster.tf
ashayshub: I checked this. In order to achieve this we need to have the terraform runner able to connect to the private endpoint of the eks. This way the terraform runner running the terraform code will be able to connect to the EKS. For this we may have to use self hosted terraform runner with the tailscale configured. I don't think this is required but we can do this at a later stage if at all needed for more security.

terraform/infra/modules/observability/metrics-server.tf
ashayshub: I have implemented this in the EKS module itself along with a [few other addons](https://github.com/svix/coyote-private/blob/ashay/base-tf-structure/terraform/infra/modules/eks/cluster.tf#L27-L109).

ashayshub:
> * Let's ditch the pet names -- we already have an animal name for the project, adding another random animal name on there is confusing 😆. If we really need this, can we just add a few random [a-z] characters instead?
--> will do

> * There are 12 subnets.... I guess maybe we do need that many. It just seems insane.
--> So
   - 1 this does not cost us.
   - 2 It is impossible later to add availabilty zones to the EKS cluster
   - 3 It would help us in testing the cluster later without needing to make changes  
  
> * What is the difference between `svix` node group and `coyote` node group?
---> One is in the database subnet group and another is in the general subnet group
 
> * Do we need modules nested under `infra` folder? Can there not just be a top-level `modules` folder under `terraform`? TBH I'm not clear if that would make more sense or not.
--> We need to segregate the infra into core infrastructure vs application infra structure. We name the terraform workspace likewise so that it is faster to deploy application which are not intertwined directly with the core infra. Apart from that it makes sense to sub-modularize the infra into network, eks etc.. 
 
> * I honestly don't like the name `edge` for the module. I'm not exactly sure what it means. Maybe `lb` or `ingress` instead?
---> I can name it ingress. The resource that resides on the brink of the infra is usually known as "edge" resource in the infra parlance. Later on adding a new service/ feature unrelated to ingress would help in calling it edge.

terraform/infra/modules/ingress/alb-ic.tf
```diff
@@ -0,0 +1,69 @@
+resource "kubernetes_namespace_v1" "alb_ctr_ns" {
+  metadata {
+    name = local.ns_alb_ctr
+  }
+}
+
+module "eks_sa_role_lb_ctrl" {
+  source  = "terraform-aws-modules/iam/aws//modules/iam-role-for-service-accounts"
```
svix-jbrown, line 8: spurious slash

```suggestion
  source  = "terraform-aws-modules/iam/aws/modules/iam-role-for-service-accounts"
```

terraform/infra/modules/eks/cluster.tf
```diff
@@ -0,0 +1,257 @@
...
+    metrics-server = {
+      most_recent = true
+    }
+
+    kube-state-metrics = {
+      most_recent = true
+    }
+
+    external-dns = {
+      most_recent = true
```
svix-jbrown, line 46: setting `most_recent` on these seems like a footgun; it means that *sometimes* a random terraform apply will have a side-effect of bumping a component. I kind of think we should be pinning all of the versions (which I think we do for the other EKS cluster).

terraform/infra/modules/eks/cluster.tf
```diff
@@ -0,0 +1,257 @@
...
+      use_latest_ami_release_version = true,
+      subnet_ids                     = var.eks_subnet_ids
+      taints = {
+        critical_addons = {
+          key    = "CriticalAddonsOnly",
+          value  = "true",
+          effect = "NO_SCHEDULE"
+        }
+      }
+    },
```
svix-jbrown, line 113: node groups with multiple AZs in them are dangerous because they'll break anything with an attached PVC, because the autoscaling controller doesn't know how to handle them. We should start this out right, with one node group of each class in each subnet.

terraform/infra/modules/eks/cluster.tf
```diff
@@ -0,0 +1,257 @@
...
+  eks_managed_node_groups = {
+    "system-node-group-0" = {
+      instance_types                 = var.system_instance_types,
+      min_size                       = var.system_min_node_count,
+      max_size                       = var.system_max_node_count,
+      desired_size                   = var.system_desired_node_count,
+      kubernetes_version             = var.k8s_version
+      ami_type                       = var.system_ami_type,
+      enable_efa_only                = false, # this is enabled by default and might not allow common instance types?
+      use_latest_ami_release_version = true,
```
svix-jbrown: same comment as before; setting this means that a random terraform apply will do the (downtime-inducing!) AMI upgrade on the node group. We should be pinning AMI releases and only upgrading by hand (and with staging/prod rollouts).

terraform/infra/modules/eks/cluster.tf
```diff
@@ -0,0 +1,257 @@
...
+    # create_placement_group         = true
+    # placement = {
+    #   availability_zone = data.aws_subnet.cayote_subnet.availability_zone
+    #   group_name        = "${local.eks_cluster_name}-coyote-pg-${data.aws_subnet.cayote_subnet.availability_zone}"
+    # }
+    # }
+  }
+
+  # auth
+  authentication_mode                      = "API_AND_CONFIG_MAP"
```
svix-jbrown, line 147: do we actually need the config map? I know it was deprecated a long time ago and I am optimistic that we can just use IAM principals for new stuff

terraform/infra/modules/eks/cluster.tf
```diff
@@ -0,0 +1,257 @@
...
+
+  access_entries = merge(
+    local.admin_users,
+    local.admin_roles,
+  )
+}
+
+#tfsec:ignore:aws-s3-enable-versioning
+#tfsec:ignore:aws-s3-enable-bucket-logging
+resource "aws_s3_bucket" "alb_log_bucket" {
```
svix-jbrown, line 158: what's all this ALB stuff doing in the `eks/cluster` module?

terraform/infra/modules/eks/iam.tf
```diff
@@ -0,0 +1,52 @@
+
+module "ebs_csi_irsa" {
+  source  = "terraform-aws-modules/iam/aws//modules/iam-role-for-service-accounts"
```
svix-jbrown: there are lots of spurious double slashes in this file

```suggestion
  source  = "terraform-aws-modules/iam/aws/modules/iam-role-for-service-accounts"
```

terraform/infra/modules/networking/subnet-eks.tf
```diff
@@ -0,0 +1,64 @@
+# only ipv4
+# for eks subnet, mimicking private subnet from
+# https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/cf18c37591f860908e2223b4f488787e8a5f74f3/main.tf
+
```
svix-jbrown, line 4: do we also want to create network endpoints for S3 (and any other services we use that have free endpoints) in here?

terraform/infra/modules/observability/kube-prometheus-stack.tf.later
```diff
@@ -0,0 +1,96 @@
...
+
+  data = {
+    username = "admin"
+    password = random_password.grafana_admin.result
+  }
+}
+
+#ToDo: Dedicated AZ for monitoring/node affinity based on availability zone
+#    : for ebs vicinity
+resource "helm_release" "kube_prometheus_stack" {
```
svix-jbrown: is there a reason this is set up totally differently than the prometheus that we use for everything else? It's not using the same repository, it's not on the same version, and it doesn't have the same config.

terraform/infra/modules/observability/obs-storageclass.tf.later
```diff
@@ -0,0 +1,16 @@
+resource "kubernetes_storage_class_v1" "gp3_ebs" {
```
svix-jbrown: why is this in the observability section instead of in the main EKS cluster section?

also, in the other clusters we intentionally disable gp2 so it doesn't get picked; is that no longer necessary?

terraform/infra/modules/observability/obs-storageclass.tf.later
```diff
@@ -0,0 +1,16 @@
...
+  metadata {
+    name = "gp3-ebs"
+  }
+
+  storage_provisioner = "ebs.csi.aws.com"
+
+  parameters = {
+    type = "gp3"
+  }
```
svix-jbrown: we should be enabling default encryption on ebs volumes

```suggestion
  parameters = {
    type = "gp3"
    encrypted = "true"
```

terraform/infra/variables.tf
```diff
@@ -0,0 +1,190 @@
...
+}
+
+variable "system_desired_node_count" {
+  type    = number
+  default = 2
+}
+
+variable "app_instance_types" {
+  type    = list(string)
+  default = ["t3.micro"]
```
svix-jbrown: I don't think anything will fit on a t3.micro.... I don't even think kubelet will fit on a t3.micro.

svix-jbrown (commented):
Wow, this is a lot of config that's totally distinct from and not compatible with the other two EKS setups we have. What's the long-term plan here? Are we committing to maintaining a totally-separate stack of terraform for this project? Is this just to set up for testing the helm chart/operator/whatever and tear down later?

On an architectural note, I'm skeptical of having the same terraform config create the EKS cluster and also install resources on it; this caused an endless number of problems for clickhouse (the official recommendation is that the config for a provider should not depend on the outcome of something else inside the same terraform state), which is why we have two totally independent terraform instances there. Ideally we would be using Terraform Stacks to get independent execution for the lower level that creates the network/EKS cluster, and the higher level that installs resources onto the cluster. If we don't want to use Stacks, we could also do it by having two different terraform states like we do for clickhouse (or we can just accept that sometimes this will fail in obscure ways).

terraform/infra/modules/ingress/alb-ic.tf
ashayshub: Not spurious, it is scoped for the `iam-role-for-service-accounts` submodule.

terraform/infra/modules/eks/cluster.tf
ashayshub: Most of these will work well for the cluster_version they have released it against. I wasn't in favour of using addons for cert-manager and external-dns but gave in considering the ease of use. I agree any major version change and I'll be in cold waters. I am taking out cert-manager and external-dns out of these addons section and am going to use a plain helm_release. For one reason: the lack of documentation on what goes behind it. I tried looking it up and it is not easy.

terraform/infra/modules/eks/cluster.tf
ashayshub: - Would the coyote application place a criteria on users to deploy the db in a nodegroup without multi-az? Most clusters using this would have multi-az already into it? Shouldn't we architect the k8s operator in such a way that the operator chooses a primary zone for the db app (depending on the cloud) and sticks with it for the PVC with headless service that allows one to connect to a specific zone?

- Will this db app need multi-az clustering at some point in time (I don't know when that will be if at all) say read replicas across AZ for better perf(?

terraform/infra/modules/eks/cluster.tf
ashayshub: True. Will fix it.

terraform/infra/modules/eks/cluster.tf
ashayshub: I have just explicitly set the default value here. I can remove it but let's go with the defaults unless aws changes it?

terraform/infra/variables.tf
ashayshub: Agree. will fix it.

terraform/infra/modules/observability/obs-storageclass.tf.later
ashayshub: So this SC was only scoped for observability but i'll change it and use a general one since I guess it isn't a norm to scope it for particular module. One SC would be enough for all the cluster requirements.

svix-james:
Can we move this under the `infra` folder, just to avoid cluttering top-level any further?

PR #263, opened 2026-02-27
Implement commit-offsets

svix-gabriel:
Next step is performance optimizations

### DIFF
--- a/crates/msgs/src/operations/stream_commit.rs
+++ b/crates/msgs/src/operations/stream_commit.rs
@@ -1,0 +1,76 @@
+use coyote_namespace::entities::NamespaceId;
+use jiff::Timestamp;
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    State,
+    entities::{ConsumerGroup, Offset, PartitionIndex},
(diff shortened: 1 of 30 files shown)


crates/msgs/src/operations/stream_commit.rs
```diff
@@ -0,0 +1,76 @@
...
+        Self {
+            namespace_id,
+            partition,
+            cg,
+            offset,
+        }
+    }
+
+    fn apply_real(self, state: &State) -> coyote_operations::Result<StreamCommitResponseData> {
+        let now = Timestamp::now();
```
svix-jbrown: the timestamp needs to come as part of the commit operation so that it's the same on every node

svix-gabriel: ah fuck I forgot about that. LM fix

svix-gabriel: Fixed. I messed this up in a few places so I went back and fixed it everywhere

svix-jbrown (approved):

PR #269, opened 2026-02-27
Fix default partition count, allowing configuring partitions

svix-gabriel:
Mistakenly earlier had topics start with a default of 16 partitions, rather than 1.

So we start with 1, but need to support configurable number of partitions for together.ai.

I've capped it at 64 for now, but this is arbitrary.

### DIFF
--- a/crates/msgs/src/tables.rs
+++ b/crates/msgs/src/tables.rs
@@ -527,8 +527,70 @@ impl OffsetRow {
     }
 }
 
+// ---------------------------------------------------------------------------
+// TopicConfigRow — per (namespace, topic) configuration (e.g. partition count)
+// ---------------------------------------------------------------------------
+
+const TOPIC_CONFIG_PREFIX: &[u8] = b"_MSGTOPICCONF_\0";
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub(crate) struct TopicConfig {
+    pub partition_count: u16,
+}
+
+fn topic_config_key(ns_id: NamespaceId, topic: &str) -> Vec<u8> {
+    let mut key = Vec::with_capacity(TOPIC_CONFIG_PREFIX.len() + 16 + 1 + topic.len());
+    key.extend_from_slice(TOPIC_CONFIG_PREFIX);
+    key.extend_from_slice(&ns_id.as_u128().to_be_bytes());
+    key.extend_from_slice(b"\0");
+    key.extend_from_slice(topic.as_bytes());
+    key
+}
+
+pub(crate) struct TopicConfigRow;
+
+impl TopicConfigRow {
+    pub(crate) fn fetch(
+        state: &State,
+        ns_id: NamespaceId,
+        topic: &str,
+    ) -> Result<Option<TopicConfig>> {
+        let key = topic_config_key(ns_id, topic);
+        match state.metadata_tables.get(&key)? {
+            Some(val) => {
+                let config: TopicConfig = rmp_serde::from_slice(&val).map_err(Error::generic)?;
+                Ok(Some(config))
+            }
+            None => Ok(None),
+        }
+    }
+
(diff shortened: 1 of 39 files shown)


crates/msgs/src/tables.rs
```diff
@@ -527,8 +527,70 @@ impl OffsetRow {
...
+}
+
+fn topic_config_key(ns_id: NamespaceId, topic: &str) -> Vec<u8> {
+    let mut key = Vec::with_capacity(TOPIC_CONFIG_PREFIX.len() + 16 + 1 + topic.len());
+    key.extend_from_slice(TOPIC_CONFIG_PREFIX);
+    key.extend_from_slice(&ns_id.as_u128().to_be_bytes());
+    key.extend_from_slice(b"\0");
+    key.extend_from_slice(topic.as_bytes());
+    key
+}
```
svix-jbrown, line 548: I feel like we do this in enough places, and it's catastrophic enough if someone makes an error like forgetting a nul byte that we should introduce some safe wrapper around a prefixed key (which does things like guarantee that the nul bytes are in place).

crates/msgs/src/tables.rs
```diff
@@ -527,8 +527,70 @@ impl OffsetRow {
...
+
+    pub(crate) fn store(
+        batch: &mut OwnedWriteBatch,
+        state: &State,
+        ns_id: NamespaceId,
+        topic: &str,
+        config: &TopicConfig,
+    ) -> Result<()> {
+        let key = topic_config_key(ns_id, topic);
+        let val: Vec<u8> = rmp_serde::to_vec(config).map_err(Error::generic)?;
```
svix-jbrown: you probably want to use `.to_vec_named` so you can add new fields in a backwards-compatible way in the future.

```suggestion
        let val: Vec<u8> = rmp_serde::to_vec_named(config).map_err(Error::generic)?;
```

src/v1/endpoints/msgs.rs
```diff
@@ -270,6 +270,64 @@ async fn stream_commit(
...
+    State(state): State<AppState>,
+    Extension(repl): Extension<RaftState>,
+    MsgPackOrJson(data): MsgPackOrJson<TopicConfigureIn>,
+) -> Result<MsgPackOrJson<TopicConfigureOut>> {
+    if data.topic.contains('~') {
+        return Err(Error::http(HttpError::bad_request(
+            Some("invalid_topic".to_owned()),
+            Some("Topic name must not contain '~'.".to_owned()),
+        )));
+    }
```
svix-jbrown, line 303: can be a cleanup in a latear PR, but I'd love to see us using validator attributes (or, better yet, having a newtype wrapper around `String` for `TopicName` that has the validator attached to it) so we don't have this code copy-pasted.

src/v1/endpoints/msgs.rs
```diff
@@ -270,6 +270,64 @@ async fn stream_commit(
...
+
+    if data.partitions == 0 || data.partitions > MAX_PARTITION_COUNT {
+        return Err(Error::http(HttpError::bad_request(
+            Some("invalid_partition_count".to_owned()),
+            Some(format!(
+                "Partition count must be between 1 and {MAX_PARTITION_COUNT}."
+            )),
+        )));
```
svix-jbrown, line 311: ditto; this should probably be a validator on the struct

src/v1/endpoints/msgs.rs
svix-gabriel: Yup, agreed (though I'd like to save it for a later PR).

crates/msgs/src/tables.rs
svix-gabriel: Also reasonable, I'll do that in a follow up PR (it's late in the day for me)

svix-jbrown (approved):

PR #270, opened 2026-02-27
cluster: do not block every write on fsync

svix-jbrown:
This drops the median response time for a write on my system from 4033µs (and much more when the write rate exceeds the max flush rate) to 111µs (and remaining stable up to local CPU saturation).

Part of #271

### DIFF
--- a/src/cfg/mod.rs
+++ b/src/cfg/mod.rs
@@ -142,6 +142,14 @@ impl From<Dir> for PathBuf {
     }
 }
 
+#[derive(Debug, Deserialize, PartialEq, Eq, Clone, Copy, Default)]
+#[serde(rename_all = "kebab-case")]
+pub enum FsyncMode {
+    EveryCommit,
+    #[default]
+    EverySecond,
+}
+
 impl DatabaseConfig {
     fn database(dir: &Path, file: &str) -> Result<fjall::Database> {
         let dir = Dir::new(dir)?;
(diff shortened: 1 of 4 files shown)


src/cfg/mod.rs
```diff
@@ -279,6 +287,13 @@ pub struct ClusterConfiguration {
         default = "defaults::log_index_interval"
     )]
     pub log_index_interval: Duration,
+
```
tasn, line 290: Not a blocker for this PR, but we should probably allow setting it on the namespace rather than the config so that people can change the durability mode (was already on our roadmap anyway).

src/cfg/mod.rs
```diff
@@ -279,6 +287,13 @@ pub struct ClusterConfiguration {
...
+    /// How often to sync logs. If this is set to "every second" and you experience a catastrophic
```
tasn, line 291: Do we have a way to auto-detect this? (doesn't need to be part of the PR)

So that we recover ourselves rather than telling people to do it.

tasn (approved):

src/cfg/mod.rs
svix-jbrown: Logs don't have namespaces.

src/cfg/mod.rs
svix-jbrown: I don't think so; I'm not sure how we would tell if the OS had crashed and thrown away all in-memory buffers.

PR #280, opened 2026-02-28
Add simple cli benchmark command#256

tasn:
It can be run with configurable number of iterations and modules to be run against. Right now idempotency/ rate-limiting are not supported because there are sdk issues with them.

Based on #256


clients/cli/src/cmds/benchmark.rs
```diff
@@ -0,0 +1,565 @@
...
+    iterations: u64,
+    all_stats: &mut Vec<Stats>,
+) -> Result<()> {
+    let test_name = bench.name();
+    let pb = new_bar(test_name.to_string(), iterations);
+    let handles = (0..concurrency).map(|shard_id| {
+        let client = Arc::clone(&client);
+        let pb = pb.clone();
+        bench.clone().bench_shard(client, iterations, pb, shard_id)
+    });
```
svix-jbrown, line 212: you need to either use a barrier before starting the the `bench_shard` function or substantially increase your iterations, otherwise you're going to get statistical oddities at high concurrency (where there's a substantial time during which not all tasks have been spawned).

clients/cli/Cargo.toml
```diff
@@ -19,6 +19,10 @@ doctest = false
 
 [dependencies]
 anyhow.workspace = true
+comfy-table = "7"
+governor = "0.7"
+indicatif = "0.17"
+quanta = "0.12"
```
svix-jbrown: I have never used this library and am always a little skeptical of anything that uses RDTSC directly (given the issues with frequency scaling); I think it's kind of silly to imagine that we're going to get into the nanosecond range where a vDSO call is too expensive. Whatever, it's just cruft.

clients/cli/Cargo.toml
```diff
@@ -27,9 +31,12 @@ config.workspace = true
...
 idna_adapter.workspace = true
+rand.workspace = true
 rustls.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 tokio = { workspace = true, features = ["macros", "time", "rt-multi-thread"] }
 tracing.workspace = true
 tracing-subscriber.workspace = true
+hdrhistogram = "7.5.4"
+futures = "0.3.32"
```
svix-jbrown: you could just use futures-util from the workspace instead

tasn: Fixed.

clients/cli/Cargo.toml
tasn: Good call, I'll get rid of it.

clients/cli/src/cmds/benchmark.rs
tasn: Can you make a code suggestion? If not, I'll figure out how to take it form your example.

tasn: NVM, on it.

svix-gabriel (approved):

PR #288, opened 2026-02-28
expose HTTP/2

svix-jbrown:
This makes a few changes:

- enables the `http2` feature in axum so it supports both
- exposes `http2` as a feature in coyote-cli and enables it by default
- adds a `http1` option to both the rust client and the cli which enables HTTP1; by default, we force HTTP/2 even over plaintext (RFC be damned).

Tested that this makes HTTP/2 requests by default but falls back to HTTP/1.1 if you pass `COYOTE_HTTP1=true`

### DIFF
--- a/clients/cli/src/cmds/benchmark.rs
+++ b/clients/cli/src/cmds/benchmark.rs
@@ -751,13 +751,13 @@ impl BenchShard for BenchMsgsStreamReceive {
             .unzip();
         BenchResult::finalize_result_stats(
             Arc::clone(&cfg),
-            rcv_results.into_iter(),
+            rcv_results,
             format!("{} - receive", self.test_name()),
             all_stats,
         )?;
         BenchResult::finalize_result_stats(
             Arc::clone(&cfg),
-            commit_results.into_iter(),
+            commit_results,
             format!("{} - commit", self.test_name()),
             all_stats,
         )?;
--- a/clients/cli/src/config.rs
+++ b/clients/cli/src/config.rs
@@ -10,9 +10,9 @@ pub struct Config {
     pub auth_token: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     server_url: Option<String>,
-    #[cfg(feature = "http2")]
+    #[cfg(feature = "http1")]
     #[serde(default)]
-    pub force_http2: bool,
+    pub http1: bool,
     #[serde(skip_serializing_if = "Option::is_none")]
     debug_url: Option<String>,
 }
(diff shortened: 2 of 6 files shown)


tasn (dismissed):

tasn (dismissed):

clients/cli/src/cmds/benchmark.rs
```diff
@@ -710,13 +710,13 @@ impl BenchShard for BenchMsgsStreamReceive {
             .unzip();
         BenchResult::finalize_result_stats(
             Arc::clone(&cfg),
-            rcv_results.into_iter(),
+            rcv_results,
```
tasn, line 754: Weird that it was needed here.

clients/cli/src/config.rs
```diff
@@ -10,6 +10,9 @@ pub struct Config {
     pub auth_token: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     server_url: Option<String>,
+    #[cfg(feature = "http2")]
+    #[serde(default)]
```
tasn, line 14: I'd say: shouldn't be a setting, just always on. We also don't need the separate feature, we should always turn http2. It's our wire protocol.

clients/rust/src/client.rs
```diff
@@ -41,6 +41,10 @@ pub struct CoyoteOptions {
     /// between the last two is that DNS resolution also goes through the proxy
     /// for `socks5h`, but not for `socks5`.
     pub proxy_address: Option<String>,
+
+    /// Force the use of HTTP/2, even on plaintext connections
+    #[cfg(feature = "http2")]
+    pub http2_only: bool,
```
tasn: Same comments as the CLI.

clients/cli/src/cmds/benchmark.rs
svix-jbrown: `clippy` did it when I ran it.

clients/cli/src/config.rs
svix-jbrown: We had a whole conversation about supporting people in environments that cannot use HTTP/2, such as behind traditional load balancers... The RFC says that we support both HTTP/1.1 and HTTP/2.

clients/cli/src/cmds/benchmark.rs
tasn: Yeah, I run `just lint` and it didn't do it, and anyway CI worked for my PR...

clients/cli/src/config.rs
tasn: Yeah, but the default should be http2.0 and you'd need to force it go down to 1.1, not the other way around.

clients/cli/src/cmds/benchmark.rs
svix-jbrown: what nightly are you on?

tasn: Fairly recent:
```
cargo 1.95.0-nightly (ce69df6f7 2026-02-12)
```

tasn (approved):

PR #289, opened 2026-02-28
Grab bag of performance optimizations

svix-gabriel:
This is a grab bag of performance optimizations, all of them suggested by Claude.

Each optimization run was tested separately. I'd:
- Spin up the coyote server (using `--release`)
- run the benchmarks with a hardcoded batch_size of 1000
- implement the change
- nuke the `db` folder, rerun the server
- rerun the benchmarks
- Compare.

Some of these optimizations may be, erm, controversial, so we can just throw each one out individually if need be.

Also all of them were pretty underwhelming.

One change Claude suggested, which did seem reasonable but also much more non-trivial to implement, is optimizing the Raft path so we aren't serializing entire message payloads in the raft layer. I'm not sure if that's even possible, I need to think about it more.

I'm gonna think more on suggestions from https://svixhq.slack.com/archives/C0A72988962/p1772216430053149?thread_ts=1772215766.329019&cid=C0A72988962, but not all of these necessarily apply since it was testing against the old implementation would have more concurrent leases active at any given time.

### DIFF
--- a/crates/msgs/src/lib.rs
+++ b/crates/msgs/src/lib.rs
@@ -1,6 +1,8 @@
+use std::sync::Arc;
+
 use coyote_error::Error;
 use coyote_namespace::entities::NamespaceId;
-use fjall::KeyspaceCreateOptions;
+use fjall::{KeyspaceCreateOptions, compaction::Fifo};
 use fjall_utils::{ReadableDatabase, ReadableKeyspace};
 
 pub mod entities;
@@ -26,7 +28,12 @@ impl State {
         };
 
         let msg_table = {
-            let opts = KeyspaceCreateOptions::default().expect_point_read_hits(true);
+            let opts = KeyspaceCreateOptions::default()
+                .expect_point_read_hits(true)
+                .compaction_strategy(Arc::new(Fifo {
+                    limit: u64::MAX,
+                    ttl_seconds: None,
+                }));
             db.keyspace(MSG_TABLE_KEYSPACE, || opts)?
         };
 
--- a/crates/msgs/src/lib.rs
+++ b/crates/msgs/src/lib.rs
@@ -21,12 +21,12 @@ impl State {
         const MSG_TABLE_KEYSPACE: &str = "_coyote_msgs";
 
         let metadata_tables = {
-            let opts = KeyspaceCreateOptions::default();
+            let opts = KeyspaceCreateOptions::default().expect_point_read_hits(true);
             db.keyspace(METADATA_KEYSPACE, || opts)?
         };
 
         let msg_table = {
-            let opts = KeyspaceCreateOptions::default();
+            let opts = KeyspaceCreateOptions::default().expect_point_read_hits(true);
             db.keyspace(MSG_TABLE_KEYSPACE, || opts)?
         };
 
--- a/crates/msgs/src/operations/publish.rs
+++ b/crates/msgs/src/operations/publish.rs
@@ -11,7 +11,7 @@ use crate::{
     entities::{
         MsgIn, Offset, Partition, RawTopic, Topic, TopicIn, partition_for_key, random_partition,
     },
-    tables::{MsgRow, msg_row_key},
+    tables::{MsgRow, NextWriteOffsetRow, msg_row_key},
 };
 
 use super::{MsgsRaftState, MsgsRequest, PublishResponse};
@@ -83,6 +83,7 @@ impl PublishOperation {
 
         for (partition, msgs) in by_partition {
             let start_offset = MsgRow::next_offset(state, self.namespace_id, partition)?;
+            let msg_count = msgs.len();
 
             for (i, (original_idx, msg)) in msgs.into_iter().enumerate() {
                 let offset = start_offset + i as Offset;
@@ -101,6 +102,15 @@ impl PublishOperation {
                     },
                 ));
             }
+
+            let next_offset = start_offset + msg_count as Offset;
+            NextWriteOffsetRow::store(
+                &mut batch,
+                state,
+                self.namespace_id,
+                partition,
+                next_offset,
+            )?;
         }
 
         batch.commit().map_err(coyote_error::Error::from)?;
--- a/crates/msgs/src/tables.rs
+++ b/crates/msgs/src/tables.rs
@@ -72,6 +72,15 @@ impl MsgRow {
         ns_id: NamespaceId,
         partition: Partition,
     ) -> Result<Offset> {
+        NextWriteOffsetRow::fetch_or_scan(state, ns_id, partition)
+    }
+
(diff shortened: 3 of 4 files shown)


crates/msgs/src/tables.rs
```diff
@@ -602,9 +615,59 @@ pub(crate) fn topic_partition_count(state: &State, ns_id: NamespaceId, topic: &s
     }
 }
 
+// ---------------------------------------------------------------------------
+// NextWriteOffsetRow — tracks the next write offset per (namespace, partition)
+// Eliminates the reverse range scan in max_offset() on every publish.
+// ---------------------------------------------------------------------------
```
svix-jbrown, line 621: do we actually need to store this in the database? if it's purely for performance reasons, couldn't we just keep it in memory on each node? I don't really see the benefit to doing the write amplification of persisting this key on every publish

tasn:
> One change Claude suggested, which did seem reasonable but also much more non-trivial to implement, is optimizing the Raft path so we aren't serializing entire message payloads in the raft layer. I'm not sure if that's even possible, I need to think about it more.

Yes, it's on my todo (maybe there's even a ticket). At the moment we have 2x+ write amplification as we store values in both the raft log and disk.

crates/msgs/src/tables.rs
tasn: Yeah, 100% agree on this one.

I'd also say though: the performance issues weren't in publish and this shouldn't matter for receive/commit which are the problematic ones? I can review this further in a bit, but that's one thought.

svix-jbrown:
> > One change Claude suggested, which did seem reasonable but also much more non-trivial to implement, is optimizing the Raft path so we aren't serializing entire message payloads in the raft layer. I'm not sure if that's even possible, I need to think about it more.
> 
> Yes, it's on my todo (maybe there's even a ticket). At the moment we have 2x+ write amplification as we store values in both the raft log and disk.

There are two parts here:

1. The actual time to serialize/deserialize. We could probably try to refactor things to only serialize once and move the physical bytes around without deserializing again until apply time, but that would be a large and invasive change. We can add some specific tracers around the `rmp_serde` calls, but I would be surprised if they were substantial.
2. The actual storage. I do not believe there's any acceptable way to avoid writing the actual data twice because that's how write-ahead logs work, and we need to have the actual value persisted indefinitely in the log to be able to replay it to other nodes later.

crates/msgs/src/lib.rs
```diff
@@ -21,12 +23,17 @@ impl State {
...
         };
 
         let msg_table = {
-            let opts = KeyspaceCreateOptions::default();
+            let opts = KeyspaceCreateOptions::default()
+                .expect_point_read_hits(true)
+                .compaction_strategy(Arc::new(Fifo {
+                    limit: u64::MAX,
+                    ttl_seconds: None,
+                }));
```
svix-jbrown, line 36: did you figure out why fifo was making it panic yet?

crates/msgs/src/lib.rs
```diff
@@ -21,12 +23,17 @@ impl State {
...
-            let opts = KeyspaceCreateOptions::default();
+            let opts = KeyspaceCreateOptions::default().expect_point_read_hits(true);
             db.keyspace(METADATA_KEYSPACE, || opts)?
         };
 
         let msg_table = {
-            let opts = KeyspaceCreateOptions::default();
+            let opts = KeyspaceCreateOptions::default()
+                .expect_point_read_hits(true)
+                .compaction_strategy(Arc::new(Fifo {
```
tasn, line 33: This seems wrong. Reading the docs, none of the reasons to use FIFO are met.
https://docs.rs/lsm-tree/3.0.3/lsm_tree/compaction/struct.Fifo.html

From the page:
> 1. You only want to store recent data (unimportant logs, …)
> 2. Your keyspace grows monotonically (e.g. time series)
> 3. You only insert new data (no updates)


1: not the case, we don't want to delete old data I know you set it to MAX but still.
2. Not the case, we can have multiple topics per keyspace, so data is inserted anywhere...
3. We trim topics after X bytes/time so data is definitely deleted.

Though anyway, judging by the name it essentially disables compaction, which we don't want...

crates/msgs/src/operations/publish.rs
```diff
@@ -11,7 +11,7 @@ use crate::{
     entities::{
```
tasn, line 11: In addition to the previous comments: it didn't seem to make any difference.
The current performance is an order of magnitude worse than what it should be:
```

╭──────────────────────────────┬─────────┬─────────┬────────┬─────────┬─────────┬─────────┬─────────┬───────────┬──────────────╮
│ op                           ┆ ops/sec ┆ mean    ┆ ±      ┆ p50     ┆ p99     ┆ p99.9   ┆ max     ┆ bytes/sec ┆ entities/sec │
╞══════════════════════════════╪═════════╪═════════╪════════╪═════════╪═════════╪═════════╪═════════╪═══════════╪══════════════╡
│ publish (batch=1)            ┆ 2092    ┆ 956µs   ┆ 321µs  ┆ 922µs   ┆ 1.13ms  ┆ 3.83ms  ┆ 10.13ms ┆ 5.93 MB   ┆ 2092         │
│ receive (batch=1) - receive  ┆ 445     ┆ 4.49ms  ┆ 1.11ms ┆ 4.93ms  ┆ 6.94ms  ┆ 9.18ms  ┆ 9.70ms  ┆ 52.88 KB  ┆ 445          │
│ receive (batch=1) - commit   ┆ 1490    ┆ 1.34ms  ┆ 1.48ms ┆ 0µs     ┆ 4.32ms  ┆ 5.03ms  ┆ 5.87ms  ┆ --        ┆ 1490         │
│ publish (batch=10)           ┆ 153     ┆ 13.06ms ┆ 2.30ms ┆ 13.79ms ┆ 17.89ms ┆ 25.50ms ┆ 29.01ms ┆ 4.34 MB   ┆ 1530         │
│ receive (batch=10) - receive ┆ 199     ┆ 10.02ms ┆ 3.14ms ┆ 10.09ms ┆ 17.23ms ┆ 20.57ms ┆ 20.80ms ┆ 2.02 MB   ┆ 1990         │
│ receive (batch=10) - commit  ┆ 831     ┆ 2.41ms  ┆ 2.90ms ┆ 1µs     ┆ 8.81ms  ┆ 11.78ms ┆ 26.45ms ┆ --        ┆ 8310         │
╰──────────────────────────────┴─────────┴─────────┴────────┴─────────┴─────────┴─────────┴─────────┴───────────┴──────────────╯
```
There's no way receive with 1 item should take 5ms at the P50, and p99 of 20ms. It just doesn't make any sense.

I'm looking into this now (I was doing general improvements for the last few hours) and will report back.

PR #294, opened 2026-03-01
Use the mimalloc memory allocator.

tasn:


svix-jbrown (approved):
oh yeah, good point. I keep forgetting that Rust removed the jemalloc-by-default behavior a few years back.

PR #295, opened 2026-03-01
Fjall: configure wanted cache_size to fjall recommended values.

### LINKED ISSUE #137: Verify we like the fjall configs (compaction, compression, etc)
Assuming they are a pain to change later.

KV separation for some (all?) of the modules, compaction, compression, etc.

**MAIN MISSING*:
- [ ] Setting keyspace options sensibly for each module: E.g. kv separation (and threshold), etc. https://docs.rs/fjall/latest/fjall/struct.KeyspaceCreateOptions.html
  - [ ] Should we enable `expect_point_read_hits` for everything? I feel like we should have mostly just successful lookups? **NOTE:** well, not the case in kv/cache/idempotency/rate-limit, so probably not for everything.
- [ ] The Raft databases are potentially created separately, we need to make sure the values there are good too. (both the db and the keyspaces).
- [ ] Potentially reduce the ephemeral cache size to much lower as it should be on faster disk anyway.

tasn:
Partially fixes #137

### DIFF
--- a/src/cfg/mod.rs
+++ b/src/cfg/mod.rs
@@ -151,10 +151,29 @@ pub enum FsyncMode {
 }
 
 impl DatabaseConfig {
+    fn default_cache_size() -> u64 {
+        let mut sys = sysinfo::System::new_all();
+        sys.refresh_all();
+
+        let ret = sys.total_memory();
+        if ret == 0 {
+            // Default to fjall's current default (32 MiB)
+            32 * 1024 * 1024
+        } else {
+            // Fjall recommends 20-25% of the memory for cache.
+            // We can probably do more, but let's start with that.
+            ret / 5
+        }
+    }
+
     fn database(dir: &Path, file: &str) -> Result<fjall::Database> {
         let dir = Dir::new(dir)?;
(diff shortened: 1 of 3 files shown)


src/cfg/mod.rs
```diff
@@ -151,10 +151,29 @@ pub enum FsyncMode {
...
+        sys.refresh_all();
+
+        let ret = sys.total_memory();
+        if ret == 0 {
+            // Default to fjall's current default (32 MiB)
+            32 * 1024 * 1024
+        } else {
+            // Fjall recommends 20-25% of the memory for cache.
+            // We can probably do more, but let's start with that.
+            ret / 5
```
svix-jbrown, line 165: so we'll be using 40%, since we have two database instances?

svix-jbrown (approved):

src/cfg/mod.rs
tasn: Yes, that's why I went with the lower bound of this.

Though thinking about it a bit more, we can probably reduce it for the ephemeral one. I added a note here: https://github.com/svix/coyote-private/issues/137

PR #299, opened 2026-03-01
dev: add local harness for benchmarking (jaeger, etc.)

tasn:


testing-docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+    command:
+      - "--config.file=/etc/prometheus/prometheus.yml"
+      - "--storage.tsdb.path=/prometheus"
+      - "--web.enable-otlp-receiver"
+  jaeger:
+    image: jaegertracing/all-in-one:1.76.0
+    ports:
+      - "16686:16686"
+      - "4317:4317"
+      - "4318:4318"
```
svix-jbrown, line 17: what do you think of making these run on different ports than the ones in svix-webhooks-private so that people don't have conflicts if they start both?

testing-docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+    ports: ["9090:9090"]
+    volumes:
+      - prometheus-data:/prometheus
+      - "./docker/prometheus:/etc/prometheus/"
+    command:
+      - "--config.file=/etc/prometheus/prometheus.yml"
+      - "--storage.tsdb.path=/prometheus"
+      - "--web.enable-otlp-receiver"
+  jaeger:
+    image: jaegertracing/all-in-one:1.76.0
```
svix-jbrown, line 13: if we're starting anew, we should use jaeger 2.x instead of this old one; svix-webhooks-private just uses the old one because we have otel tests that use the undocumented internals APIs to read data back out from jaeger and those tests are broken on jaeger 2

testing-docker-compose.yml
tasn: Sure, any suggestions? Or just whatever I want?

testing-docker-compose.yml
tasn: I looked into that (saw the FIXME), but there doesn't seem to be a 2.0 all-in-one image, so I went with 1.0. :(
Any idea what's the solution here?

Maybe it's named differently? I don't know, I just stopped looking after 5-10mins. :)

svix-jbrown: I can try to look at it tomorrow; IIRC there isn't an all-in-one any more but the regular docker container has all the functionality.

testing-docker-compose.yml
svix-jbrown: just grep through the other codebases I guess

svix-gabriel (approved):

testing-docker-compose.yml
tasn: Let's do it in a follow up PR, feel free to change it to whatever you want. I just want to clean the way for #303.

testing-docker-compose.yml
tasn: Let's do it in a follow up PR, feel free to change it to whatever you want. I just want to clean the way for https://github.com/svix/coyote-private/pull/303.

PR #300, opened 2026-03-02
Instrument stream publish/receive/commit

svix-gabriel:
I tried to capture all the methods that do any sort of IO. There's a bunch of helpers that are "pure" (cpu only). I didn't bother instrumenting those.

Or at least that was the goal.

### DIFF
--- a/crates/msgs/src/operations/stream_receive.rs
+++ b/crates/msgs/src/operations/stream_receive.rs
@@ -101,9 +101,9 @@ impl StreamReceiveOperation {
             }
         }
 
-        let span = Span::current();
-        span.record("partition_count", partition_count);
-        span.record("partitions_eligible", partitions.len());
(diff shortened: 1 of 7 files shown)


crates/msgs/src/operations/stream_receive.rs
```diff
@@ -99,6 +101,10 @@ impl StreamReceiveOperation {
             }
         }
 
+        let span = Span::current();
+        span.record("partition_count", partition_count);
+        span.record("partitions_eligible", partitions.len());
```
svix-jplatte:
```suggestion
        Span::current()
            .record("partition_count", partition_count)
            .record("partitions_eligible", partitions.len());
```

svix-jplatte (dismissed):

svix-jplatte (approved):

PR #314, opened 2026-03-02
cluster: refactor log sync worker

svix-jbrown:
This removes the hardcoded "every commit" and "every second" options in favor of more flexible configuration. It also adds configuration validation since there are more complicated invariants here now.

### DIFF
--- a/src/core/cluster/logs.rs
+++ b/src/core/cluster/logs.rs
@@ -4,6 +4,7 @@ use std::{
     fmt::Debug,
     ops::{Bound, RangeBounds},
     sync::Arc,
+    time::Duration,
 };
 
 use anyhow::Context;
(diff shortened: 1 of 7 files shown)


svix-gabriel (dismissed):

src/core/cluster/logs.rs
```diff
@@ -238,63 +239,67 @@ impl std::error::Error for BackgroundFsyncFailedError {
...
                 .map_err(|err| {
                     tracing::error!(?err, "error flushing fjall in background");
                     BackgroundFsyncFailedError(err.to_string())
-                });
-            for callback in new_buf.drain(..) {
-                callback.log_io_completed(result.clone().map_err(std::io::Error::other));
-            }
+                })
         })
+        .instrument(tracing::info_span!("logs:flush_worker:every"))
```
svix-jplatte: You're undoing parts of #304 here and below. I'm guessing that's not intentional?

svix-jplatte (changes requested):

src/core/cluster/logs.rs
svix-jbrown: Probably rebase artifacts since this PR is originally from Sunday. I'll rebase it again.

svix-jplatte (approved):

PR #315, opened 2026-03-02
Namespace: some cleanup for fetching namespaces across modules

tasn:


crates/coyote-namespace/src/lib.rs
```diff
@@ -102,27 +98,15 @@ impl State {
...
-
-    pub fn fetch_idempotency_namespace(
-        &self,
-        namespace_name: &str,
-    ) -> Result<Option<Namespace<KeyValueConfig>>> {
+    ) -> Result<Option<Namespace<C>>> {
         Namespace::fetch(&self.keyspace, namespace_name)
     }
 
+    // FIXME: this module is being removed.
```
svix-gabriel: FWIW I already went ahead and removed the module if you want to do it here as well.

crates/coyote-namespace/src/lib.rs
```diff
@@ -152,14 +136,14 @@ mod tests {
     use super::*;
 
     #[test]
-    fn test_namespace_name() {
-        assert_eq!(namespace_name("tom:bar"), "tom");
-        assert_eq!(namespace_name("tom:bar/baz"), "tom");
-        assert_eq!(namespace_name("bill"), DEFAULT_NAMESPACE_NAME);
-        assert_eq!(namespace_name(":bar"), DEFAULT_NAMESPACE_NAME);
+    fn test_namespace_parse_key() {
+        assert_eq!(namespace_parse_key("tom:bar"), (Some("tom"), "bar"));
```
svix-gabriel: Not that I was a huge fan of `namespace_name` name before, but I think `namespace_parse_key` can potentially be more confusing since `key` already means something different in different contexts (kv vs msgs, etc.).

Maybe `parse_namespace`?

tasn: Fixing, thanks! Will also rebase.

svix-gabriel (approved):

PR #317, opened 2026-03-02
Propagate tracing info across raft requests

svix-james:
This will cause the raft requests to show up as child span of the original HTTP request trace.

### DIFF
--- a/crates/coyote-operations/src/lib.rs
+++ b/crates/coyote-operations/src/lib.rs
@@ -1,5 +1,6 @@
 use axum::response::IntoResponse;
-use std::fmt::Debug;
+use opentelemetry::{Context, trace::TraceContextExt};
+use std::{collections::HashMap, fmt::Debug};
 
 use serde::{Deserialize, Serialize, de::DeserializeOwned};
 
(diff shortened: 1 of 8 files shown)


src/core/cluster/handle.rs
```diff
@@ -37,7 +37,7 @@ impl std::error::Error for ResponseParseError {
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug)]
-pub enum Request {
```
svix-jbrown, line 37: I really don't want to rename this core type, and would prefer to instead make the wrapper (which is only used in two places) the thing that has a differnt name

src/core/cluster/handle.rs
```diff
@@ -49,71 +49,86 @@ pub enum Request {
     Msgs(coyote_msgs::operations::MsgsOperation),
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct Request {
+    pub inner: OpRequest,
+    pub context: Option<OperationRequestMetadata>,
```
svix-jbrown, line 53: this should have `skip_serializing_if`, and probably also rename to smaller fields

crates/coyote-operations/src/lib.rs
```diff
@@ -28,6 +29,21 @@ pub trait ModuleRequest: Serialize + DeserializeOwned + Clone + Debug {
...
+
+impl From<Context> for OperationRequestMetadata {
+    fn from(ctx: Context) -> Self {
+        let mut map = HashMap::new();
+        opentelemetry::global::get_text_map_propagator(|p| p.inject_context(&ctx, &mut map));
+        OperationRequestMetadata {
+            trace_context: if map.is_empty() { None } else { Some(map) },
+        }
+    }
+}
```
svix-jbrown, line 51: this seems like potentially a lot of verbose data to put on every single request. do we need the full trace context as a stringly typed map or can we just propagate the u64 trace ID?

svix-gabriel (approved):

PR #335, opened 2026-03-03
Github action to publish docker image

svix-james:
Mostly copy-pasted from svix-webhooks-private.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,8 +1,7 @@
 name: Coyote Docker Build
 
 on:
-  push:
-    branches: ['main']
+  workflow_call:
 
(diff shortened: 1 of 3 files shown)


.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,191 @@
+name: Coyote Docker Build
+
+on:
+  push:
+    branches: ['main']
```
svix-jbrown: I would prefer not to have individual workflows have action triggers; can you change this to `on: workflow_call:` and then add it to the dispatch list in the `on-push`? Also, that lets us easily sequence it so we only push a docker image after CI passes.

svix-jbrown (approved):

PR #343, opened 2026-03-03
Change fjall-utils::TableRow to not require key within the struct

svix-gabriel:
The `TableRow` trait required the fields that make up the "key" for a row be defined within the struct.

This limitation made some methods (like `insert`) simpler, but at the cost of fatter structs being inserted in `fjall`, and duplicate trait types that tried to get around this limitation, or skipping serialization of select fields, etc.

Ultimately, there's no reason why logical rows need all of their fields within the struct, and there's no reason to force redundancy between fields stored in the key and fields stored in the value.

This ends up letting us delete quite a lot of code, and serialize thinner structs into fjall.


crates/fjall-utils/src/table_row.rs
```diff
@@ -115,9 +108,10 @@ pub trait TableRow: Sized + Serialize + DeserializeOwned {
         keyspace.get(&key)?.map(Self::from_fjall_value).transpose()
     }
 
-    fn insert(keyspace: &fjall::Keyspace, row: &Self) -> Result<()> {
-        let (key, value) = row.to_fjall_entry()?;
-        keyspace.insert(key, value)?;
+    fn insert(keyspace: &fjall::Keyspace, key: &Self::Key, row: &Self) -> Result<()> {
```
svix-jbrown, line 111: I don't like having `&Self::Key` here because in the common case of String keys, that gives us a required arg of `&String`.

Can we instead do what std does for the map types and use another generic parameter and the `Borrow` trait to allow people to pass `&str` or `Arc<str>` or their other favorite reborrow type?

I think this will require refactoring the `TableKey` trait a bit, since right now it's defined for `String` instead of `str`.

crates/kv/src/lib.rs
```diff
@@ -115,16 +115,16 @@ impl KvStore {
         let mut batch = self.db.batch();
 
         let row = KvPairRow {
-            key: key.to_string(),
             value: model.value.clone(),
             expiry: model.expiry,
         };
 
-        batch.insert_row(&self.tables, &row)?;
+        batch.insert_row(&self.tables, &key.to_string(), &row)?;
```
svix-jbrown, line 122: this is an example of the cloning I think we can probably avoid

crates/msgs/src/tables.rs
```diff
@@ -1,77 +1,46 @@
...
     pub partitions: u16,
 }
 
 impl TopicRow {
-    pub(crate) fn construct_key(namespace_id: NamespaceId, topic: &TopicName) -> Vec<u8> {
-        let parts = ["top".as_bytes(), namespace_id.as_bytes(), topic.as_bytes()];
-
-        construct_key(&parts)
+    pub(crate) fn key_for(namespace_id: NamespaceId, topic: &TopicName) -> Vec<u8> {
+        let mut key = Vec::with_capacity(16 + topic.len());
```
tasn, line 22: Putting the numbers explicitly that is error-prone, it's also unnecessary. The code you removed is as efficient but much cleaner than this.

https://github.com/svix/coyote-private/pull/343/changes#diff-d80920f2294f35878943ad4c6a438c476881d5e6b9b3273c1ea58d849cf00033L13

crates/msgs/src/tables.rs
```diff
@@ -1,77 +1,46 @@
...
-    pub name: TopicName,
     pub partitions: u16,
 }
 
 impl TopicRow {
-    pub(crate) fn construct_key(namespace_id: NamespaceId, topic: &TopicName) -> Vec<u8> {
-        let parts = ["top".as_bytes(), namespace_id.as_bytes(), topic.as_bytes()];
-
-        construct_key(&parts)
+    pub(crate) fn key_for(namespace_id: NamespaceId, topic: &TopicName) -> Vec<u8> {
```
tasn, line 21: Should probably return `fjall::Key` (or whatever it's called), no?

crates/msgs/src/tables.rs
```diff
@@ -1,77 +1,46 @@
...
-        topic: &TopicName,
-    ) -> Result<Option<Self>> {
-        let key = Self::construct_key(namespace_id, topic);
-        Self::fjall_fetch(keyspace, &key)
-    }
 }
 
-impl TableRow for TopicRow {}
+impl TableRow for TopicRow {
+    const TABLE_PREFIX: &'static str = "top";
```
tasn, line 41: We talked about this privately where I mentioned this. Doing it this way is much worse, because this means that we now create a Vec<u8> to create the key, copy values to it, then now we create a larger key to copy top to it, and then copy all the values to it.

crates/fjall-utils/src/table_row.rs
tasn: Ha, I actually made the same comment elsewhere. I agree, no reason why not just have a fjall key, as we generate the keys anyway.

@svix-jbrown, fwiw, I don't think string keys are common, it's probably always Vec<u8> (due to `\0` delimiters), but point still stands. (though maybe different in the raft layer).

svix-gabriel: Yeah, I actually don't like using String keys, and am trying to move away from them for the msgs module. I didn't want to add a bunch of type noise to optimize this, but I'll see what I can make work with the Borrow trait

crates/msgs/src/tables.rs
svix-gabriel: Yes. There's a bunch of things I need to change with this table (a lot of the key construction in this module is a mess). I didn't want it to pollute this diff though

PR #345, opened 2026-03-03
Put hashed key in request trace

svix-james:


src/core/cluster/handle.rs
```diff
@@ -76,6 +77,22 @@ impl RequestWithContext {
...
+        let key = match &self.inner {
+            Request::Kv(op) => op.key_name().to_string(),
+            Request::RateLimiter(op) => op.key_name().to_string(),
+            Request::Idempotency(op) => op.key_name().to_string(),
+            Request::Cache(op) => op.key_name().to_string(),
+            Request::Msgs(op) => op.key_name(),
+            Request::CreateCache(op) => op.key_name().to_string(),
+            Request::CreateIdempotency(op) => op.key_name().to_string(),
+            Request::CreateKv(op) => op.key_name().to_string(),
+            Request::ClusterInternal(_) => return None,
```
svix-jbrown: why all the `.to_string` calls? As far as I know, `Sha256::digest` takes any `AsRef<[u8]>`, so these clones should be unnecessary?

svix-james: Yeah mostly unnecessary.

crates/msgs/src/operations/mod.rs
```diff
@@ -29,3 +29,59 @@ raft_module_operations!(
     },
     state = MsgsRaftState<'_>,
 );
+
+impl MsgsOperation {
+    pub fn key_name(&self) -> String {
+        match self {
+            MsgsOperation::CreateNamespace(op) => op.name.to_string(),
```
svix-james, line 36: I'm not actually clear whether we want the namspace embedded in these keys, but I assume for most modules it will be so ensuring the same thing is done here.

crates/msgs/src/operations/mod.rs
```diff
@@ -29,3 +29,59 @@ raft_module_operations!(
     },
     state = MsgsRaftState<'_>,
 );
+
+impl MsgsOperation {
+    pub fn key_name(&self) -> String {
```
svix-jbrown, line 34: any reason to have this one return `String` instead of `str`?

svix-james: Yeah, it's depending on the `Display` impl for `TopicName`, which prepends the raw topic name with the namespace, if applicable.

svix-jbrown (approved):

PR #347, opened 2026-03-04
justfile: change clippy lint to not override channel.

tasn:
I have an alternative PR where I let us override the channel, but tbh,
I don't see the point if just forcing it for clippy.
We can just all use beta/nightly (we already do?) so let's just set this to that.
In CI we force beta anyway, so no need for this either.

### DIFF
--- a/justfile
+++ b/justfile
@@ -3,8 +3,7 @@ set quiet := true
 HERE := justfile_directory()
 
 lint:
-    # keep this beta to keep it in sync with CI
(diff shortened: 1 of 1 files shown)


svix-jbrown (commented):
I don't know about anyone else, but I use stable for everything; I don't really see the point of using beta

tasn:
OK, if you use stable, that means that this PR is no good. I'll close it.

PR #351, opened 2026-03-04
Fix lease getting improperly released on partial commit

svix-gabriel:
Fixes a bug where partially committing a batch would release a Lease prematurely.

### DIFF
--- a/src/core/cluster/logs.rs
+++ b/src/core/cluster/logs.rs
@@ -309,7 +309,7 @@ async fn flush_worker(
                     done = true
                 },
                 _ = ticker.tick() => {
-                    sync_now = pending.len() > 0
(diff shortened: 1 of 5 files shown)


src/core/cluster/logs.rs
```diff
@@ -309,7 +309,7 @@ async fn flush_worker(
                     done = true
                 },
                 _ = ticker.tick() => {
-                    sync_now = pending.len() > 0
+                    sync_now = !pending.is_empty()
```
svix-gabriel, line 312: `just lint` decided to make these changes, not me.

I didn't want to remove them since `just lint` will fix them anyway 🤷

tasn (approved):

PR #358, opened 2026-03-04
Revert http/2 by default for Rust SDK / CLI

svix-jplatte:
Reverts the client-side changes from #288, because we're seeing it hurt latency.

### DIFF
--- a/clients/cli/Cargo.toml
+++ b/clients/cli/Cargo.toml
@@ -17,14 +17,6 @@ name = "coyote"
 path = "src/main.rs"
 doctest = false
 
-[features]
-default = ["http1", "http2", "rustls-tls"]
-
-http1 = ["coyote-client/http1"]
-http2 = ["coyote-client/http2"]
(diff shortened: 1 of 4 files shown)


clients/cli/Cargo.toml
```diff
@@ -17,14 +17,6 @@ name = "coyote"
...
 doctest = false
 
-[features]
-default = ["http1", "http2", "rustls-tls"]
-
-http1 = ["coyote-client/http1"]
-http2 = ["coyote-client/http2"]
-native-tls = ["coyote-client/native-tls"]
-rustls-tls = ["coyote-client/rustls-tls"]
-
```
svix-jbrown, line 27: It seems reasonable to leave this forwarding in place and just change the default to `["http1", "rustls-tls"]`; I imagine anyone using TLS will want HTTP/2.

svix-jbrown (approved):
whatever, let's just land this now and I'll put the feature back in another PR

PR #363, opened 2026-03-04
Fjall utils: refactor how we do TableRow/TableKey

tasn:
This is based on discussions I've had with Gabriel, but would love other people's feedback.

The main goals:
1. Enforce (in the type system) that key and values belong together.
2. Make it harder to structure keys wrong by having a helper.
3. Have a consistent way of structuring keys.
4. Mostly fixed size keys (where possible) so that they are easy to parse.

I spend a few hours trying to change logs.rs to this, but gave up and just kept table_row2. The main problem was that I wasn't familiar with the codebase so it was hard for me to confirm I'm setting the right values. Though changing it to explicitly use the bounds with fjall (without the range_helper) shouldn't be too hard, I just gave up because I didn't trust *ME* doing those changes.

### DIFF
--- a/crates/fjall-utils/src/table_row.rs
+++ b/crates/fjall-utils/src/table_row.rs
@@ -142,5 +142,5 @@ impl<'a, Tag: TableRow> TableKey<Tag> {
 pub trait TableKeyFromFjall {
     type Key;
 
-    fn key_from_key(key: fjall::UserKey) -> Result<Self::Key>;
+    fn key_from_fjall_key(key: fjall::UserKey) -> Result<Self::Key>;
 }
--- a/crates/fjall-utils/src/table_row.rs
+++ b/crates/fjall-utils/src/table_row.rs
@@ -91,7 +91,7 @@ pub type TableKeyType = u8;
 
 pub struct TableKey<Tag: TableRow> {
     key: fjall::UserKey,
-    _unit: PhantomData<Tag>,
+    _table: PhantomData<Tag>,
 }
 
 impl<'a, Tag: TableRow> TableKey<Tag> {
@@ -123,14 +123,14 @@ impl<'a, Tag: TableRow> TableKey<Tag> {
 
         Self {
             key: ret.into(),
-            _unit: PhantomData,
+            _table: PhantomData,
         }
     }
 
     pub fn init_from_bytes(key: &'a [u8]) -> Self {
         Self {
             key: key.into(),
-            _unit: PhantomData,
+            _table: PhantomData,
         }
     }
 
--- a/crates/fjall-utils/src/table_row.rs
+++ b/crates/fjall-utils/src/table_row.rs
@@ -1,296 +1,146 @@
-use byteorder::{BigEndian, ByteOrder};
-use fjall::Guard;
-use std::{
-    borrow::Cow,
-    ops::{Bound, RangeBounds},
-};
-use uuid::Uuid;
+use std::marker::PhantomData;
 
 use super::readonly_db::ReadableKeyspace;
-use coyote_error::{Error, Result, ResultExt};
+use coyote_error::{Error, Result};
 use serde::{Serialize, de::DeserializeOwned};
-
-/// A trait for primary keys in fjall.
-pub trait TableKey: Clone {
-    fn as_bytes(&self) -> Cow<'_, [u8]>;
-
-    fn try_from_bytes(bytes: &[u8]) -> Result<Self>;
-}
-
-impl TableKey for String {
-    fn as_bytes(&self) -> Cow<'_, [u8]> {
-        Cow::Borrowed(self.as_bytes())
-    }
-
-    fn try_from_bytes(bytes: &[u8]) -> Result<Self> {
-        let owned: Vec<u8> = bytes.to_owned();
-        Self::from_utf8(owned).map_err_generic()
-    }
-}
-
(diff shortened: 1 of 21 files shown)


crates/fjall-utils/src/table_row.rs
```diff
@@ -1,296 +1,146 @@
...
-pub trait MonotonicTableKey {}
-
 /// A trait for types that can be stored as rows in a fjall keyspace.
 ///
 /// This is useful for having logical "tables" within the same keyspace.
 pub trait TableRow: Sized + Serialize + DeserializeOwned {
-    const TABLE_PREFIX: &'static str;
-
-    type Key: TableKey;
+    // FIXME: can probably get rid of this, and encode it in the type system.
```
svix-gabriel, line 11:
```suggestion
```

I don't think we (safely) can get rid of this. The moment multiple tables are in the same fjall keyspace, we need someway to safely iterate over only the keys within a keyspace, and make sure we don't accidentally iterate over the wrong entries.

crates/fjall-utils/src/table_row.rs
```diff
@@ -1,296 +1,146 @@
...
 
-/// Adds convenient methods to fjall's Keyspace to work with TableRow
-pub trait KeyspaceExt {
-    fn ingest_rows<T: TableRow, I: Iterator<Item = T>>(&self, rows: I) -> Result<()>;
+/// Can't change the size, will break everything.
+pub type TableKeyType = u8;
+
+pub struct TableKey<Tag: TableRow> {
+    key: fjall::UserKey,
+    _unit: PhantomData<Tag>,
```
svix-gabriel:
```suggestion
    _table: PhantomData<Tag>,
```
nit

crates/fjall-utils/src/table_row.rs
```diff
@@ -1,296 +1,146 @@
...
 }
 
-impl KeyspaceExt for fjall::Keyspace {
-    fn ingest_rows<T: TableRow, I: Iterator<Item = T>>(&self, rows: I) -> Result<()> {
-        let mut i = self.start_ingestion()?;
-        for row in rows {
-            let (k, v) = row.to_fjall_entry()?;
-            i.write(k, v)?;
+impl<'a, Tag: TableRow> TableKey<Tag> {
```
svix-gabriel, line 97:
```suggestion
impl<Tag: TableRow> TableKey<Tag> {
```

crates/fjall-utils/src/table_row.rs
```diff
@@ -1,296 +1,146 @@
...
+    /// Construct the key to be used for fjall
+    ///
+    /// In the future: should probably just have a big enough key on the stack and use that.
+    pub fn init_key(
+        row_type: TableKeyType,
+        fixed_parts: &[&[u8]],
+        nul_delimited_parts: &[&str],
+    ) -> Self {
+        let len = 1 /* u8 */
```
svix-gabriel: Does
```suggestion
        let len = size_of::<Tag::ROW_TYPE>()
```
work instead?

crates/kv/src/tables.rs
```diff
@@ -1,15 +1,13 @@
...
 use jiff::Timestamp;
 use serde::{Deserialize, Serialize};
 
-// IMPORTANT. Since these are all shared in the same fjall::Keyspace, the table prefixes must be unique.
-static_assertions::const_assert!(fjall_utils::are_all_unique(&[
-    KvPairRow::TABLE_PREFIX,
-    ExpirationRow::TABLE_PREFIX,
-]));
+#[repr(u8)]
+enum RowType {
```
svix-gabriel, line 8: I wonder if we should add a `///` comment here making it clear that any additions need to be done in a backwards compatible way (e.g. no re-assigning u8 values)

At least long term (I know right now it's not a concern)

svix-gabriel (commented):
I'm pretty sure we can get rid of `TableRow2`, but yeah, doesn't necessarily have to happen in this PR

crates/fjall-utils/src/table_row2.rs
```diff
@@ -0,0 +1,296 @@
+use byteorder::{BigEndian, ByteOrder};
```
tasn, line 1: It's the file that was there before, only used by logs.rs. I have changes that move that fail to the new way, but as I said in the PR description, I didn't feel confident in my changes.

crates/fjall-utils/src/table_row.rs
tasn: Sorry, let me clarify: I meant the `const ROW_TYPE`, not the concept of row type. :)

crates/fjall-utils/src/table_row.rs
tasn: Nope, errors, I also tried to remove it. 😅

crates/kv/src/tables.rs
tasn: Let me do that, one sec.

crates/fjall-utils/src/table_row.rs
tasn: I did something close. Though can't do it based on the tag.

crates/fjall-utils/src/table_row.rs
svix-gabriel: If you change init_from_bytes to not have `'a`, does it still complain?

svix-gabriel (approved):

PR #365, opened 2026-03-04
Add some db/log metrics

svix-james:
This adds some metrics around DB/log sizes.

### DIFF
--- a/src/cfg/mod.rs
+++ b/src/cfg/mod.rs
@@ -403,6 +403,7 @@ pub struct ConfigurationInner {
     pub opentelemetry_address: Option<String>,
 
     /// The OpenTelemetry address to send metrics to if given.
+    ///
     /// If not specified, the server will attempt to fall back
     /// to `opentelemetry_address`
     pub opentelemetry_metrics_address: Option<String>,
--- a/src/cfg/mod.rs
+++ b/src/cfg/mod.rs
@@ -402,6 +402,11 @@ pub struct ConfigurationInner {
     /// The OpenTelemetry address to send events to if given.
     pub opentelemetry_address: Option<String>,
 
+    /// The OpenTelemetry address to send metrics to if given.
+    /// If not specified, the server will attempt to fall back
+    /// to `opentelemetry_address`
+    pub opentelemetry_metrics_address: Option<String>,
+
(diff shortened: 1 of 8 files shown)


src/cfg/mod.rs
```diff
@@ -402,6 +402,11 @@ pub struct ConfigurationInner {
     /// The OpenTelemetry address to send events to if given.
     pub opentelemetry_address: Option<String>,
 
+    /// The OpenTelemetry address to send metrics to if given.
+    /// If not specified, the server will attempt to fall back
```
svix-jplatte, line 407:
```suggestion
    /// The OpenTelemetry address to send metrics to if given.
    ///
    /// If not specified, the server will attempt to fall back
```

src/core/metrics.rs
```diff
@@ -0,0 +1,88 @@
...
+    node_id: NodeId,
+}
+
+impl DbMetrics {
+    pub fn new(meter: &Meter, node_id: NodeId) -> Self {
+        Self {
+            bytes_used: meter
+                .u64_gauge("coyote.db.bytes_used")
+                .with_description("DB size in bytes")
+                .with_unit("By")
```
svix-jplatte, line 30: I have never seen "By" as a unit, shouldn't it just be "B"?

src/core/metrics.rs
```diff
@@ -0,0 +1,88 @@
...
+    node_id: NodeId,
+}
+
+impl LogMetrics {
+    pub fn new(meter: &Meter, node_id: NodeId) -> Self {
+        Self {
+            bytes_used: meter
+                .u64_gauge("coyote.raft.log.bytes_used")
+                .with_description("Raft log DB size in bytes")
+                .with_unit("By")
```
svix-jplatte, line 59: Same

src/core/metrics.rs
svix-james: Actually, looking at this, I'm even more confused, although it seems to show both: https://ucum.org/ucum

If I grep the opentelemetry crates for examples, I see `By` used exclusively.

svix-jplatte (approved):

src/core/metrics.rs
svix-jplatte: If this is what's standard in otel, let's keep it :shrug:

PR #366, opened 2026-03-05
Kv modules: follow raft best practices and cleanup code sharing

### LINKED ISSUE #185: Make sure that the "empty" group is saved with a `:` prefix
Just making sure that when people use the default config group, we either use the name `default:` or the `:` for the empty group when storing in fjall.

The reason for this, is that if we want to be able to clear the default group or something using an operation, we should be able to refer to it.

Maybe it doesn't matter, just a thought.

**BETTER:** use the namespace ID as a fixed sized prefix, and that just solves it.

tasn:
Raft best practices: no longer accessing the databases from places we are not allowed to.
We also moved the kv controller actions to a kvcontroller module to make it more reusable.

I changed the kv module to align to `msgs` where multiple namespaces share the same keyspace. Though I can also see a world where we change it back to keyspace per namespace. Whatever we do, I think it makes sense to be consistent here, so I opted for consistency in this direction for now.
Waiting for a response here to know what we should do: https://github.com/fjall-rs/fjall/discussions/236#discussioncomment-16011405
Edit: looks like we should keep keyspaces to a minimum if possible.


crates/coyote-namespace/src/tables.rs
```diff
@@ -47,6 +47,10 @@ impl<C: ModuleConfig> Namespace<C> {
     }
 
     pub(crate) fn fetch_all(keyspace: &Keyspace) -> Result<impl Iterator<Item = Self>> {
-        Self::values(keyspace)
+        let prefix = Self::key_for("");
```
svix-gabriel, line 50: I'm probably missing something, but why the change here away from the method that the `TableRow` trait gave us for free?

src/v1/endpoints/cache.rs
```diff
@@ -152,23 +145,47 @@ async fn cache_set(
...
     State(state): State<AppState>,
+    Extension(repl): Extension<RaftState>,
     MsgPackOrJson(data): MsgPackOrJson<CacheGetIn>,
 ) -> Result<MsgPackOrJson<CacheGetOut>> {
-    let mut cache_store = state.get_cache_store_by_key(&data.key.0).await?;
-    let model = cache_store
-        .get(&data.key)?
-        .ok_or_else(|| HttpError::not_found(None, None))?;
-    Ok(MsgPackOrJson(CacheGetOut::from_model(data.key, model)))
+    // FIXME: gotta do this route!
```
svix-gabriel: I don't quite understand what this FIXME is saying

crates/kv/src/tables.rs
```diff
@@ -22,38 +23,49 @@ impl TableRow for KvPairRow {
...
     }
-}
 
-impl TableKeyFromFjall for ExpirationRow {
-    type Key = String;
-
-    fn key_from_fjall_key(key: fjall::UserKey) -> Result<Self::Key> {
-        String::from_utf8(key.to_vec()).map_err_generic()
+    pub(crate) fn extract_key_from_fjall_key(key: &fjall::UserKey) -> Result<(NamespaceId, &str)> {
+        let namespace_offset = 1 /* row_type */ + size_of::<i64>() /* timestamp */;
```
svix-gabriel, line 55: It does pain me that we have to define this manually for the table, instead of getting it "for free" with the `TableRow`/`TableKey` stuff. (or even an inverse function of `TableKey::init_key`

I'm not immediately sure how to change it (so doesn't have to block this PR)

crates/idempotency/src/lib.rs
```diff
@@ -89,30 +112,57 @@ impl IdempotencyStore {
...
+{
+    let mut timer = tokio::time::interval(Duration::from_secs(1));
+    let controller = KvController::new(db, IDEMPOTENCY_KEYSPACE);
+
+    loop {
+        if is_shutting_down() {
+            break;
+        }
+
+        timer.tick().await;
```
svix-gabriel, line 149: nit, but the first `tick().await` always returns immediately, so it's usually considered best practice to put the tick first in the loop.

svix-gabriel (commented):
High level: I support the change. I had a few questions / nits, but mostly seems like a step forward

src/v1/endpoints/cache.rs
tasn: I guess the route was done at this point, let me remove the FIXME.

crates/kv/src/tables.rs
tasn: We didn't get it for free at all there, we would have to manually implement exactly the same thing with the previous TableKey as well. The only reason why you didn't see this code before was because were writing data on both the keys and the values so we would just load them from the value. This PR fixes that to read it from the key and serialize nothing in the value.

crates/idempotency/src/lib.rs
tasn: It's the code that was there, I didn't change it. Though it's first in the loop other than the "shutdown abort".

crates/coyote-namespace/src/tables.rs
tasn: It's not. Values returns all the values for the specific row_type, but it doesn't filter to only get the ones relevant to this specific Config type (so type of namespace). That's why this had to change accordingly.

svix-gabriel (approved):

crates/kv/src/tables.rs
svix-gabriel: Yeah, I wasn't saying we _were_ getting it for free, just that that's the end state we should be targeting.

tasn: Ohh, I thought you meant we used to get it for free but no longer. Yeah, 100% agreed, I'd love to fix this to be less stupid.

Issue #338, opened 2026-03-03
Cleanup the rust error handling in the server - less and cleaner types


tasn:
Cleanup the error handling in both raft (where the modules apply stuff to disk) and HTTP layer.
- Significantly reduce the amount of error types in coyote_error::Error
- Remove the notion of HttpError and instead have specific types (but all should have code + description like http error has). We don't care about HTTP types anymore.
- Raft requires serializable errors, so let's see if we can have something instead of HttpError which is "operational error" or whatever we want to call it, that has a few simple types, and is serializable for raft. E.g: not_found (for entity not found), bad_request (when a user makes an error), internal_error (when we want to expose the error to customer, but it's a 500 essentially), and that's it for now? Document them well to explain what they should be used for, and let's just use them?
    - Can potentially add a conflict, though I feel like that is just a confusing subset of bad_request?
    - Or maybe can split bad_request to two, one is a sort of validation/code did something wrong, and another is operational (e.g. a conflict, or some runtime case) - so essentially differentiate between permanent bad request and situational/context based.

Related to #302, as it enables us to have a small set of error types that people can `catch`.

PR #377, opened 2026-03-05
Simplify construction of not-found errors

svix-jplatte:
Part of https://github.com/svix/coyote-private/issues/338.

### DIFF
--- a/crates/coyote-error/src/lib.rs
+++ b/crates/coyote-error/src/lib.rs
@@ -16,10 +16,12 @@ use serde::Serialize;
 use serde_json::json;
 use tokio::task::JoinError;
 
+mod option_ext;
 mod result_ext;
 mod validation;
 
 pub use self::{
+    option_ext::OptionExt,
     result_ext::ResultExt,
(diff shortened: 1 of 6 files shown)


crates/coyote-namespace/src/lib.rs
```diff
@@ -80,7 +80,7 @@ impl State {
     pub fn fetch_namespace<C: ModuleConfig>(
         &self,
         namespace_name: Option<&str>,
-    ) -> Result<Option<Namespace<C>>> {
+    ) -> Result<Namespace<C>> {
```
svix-jplatte: Was there any reason this plus the associated `_admin` method were returning `Result<Option<_>>`? Every single callsite called `.ok_or_else(|| HttpError::not_found(...))?`.

svix-gabriel: I think it was to let the caller decide, in the event they want to handle a missing namespace differently than just propagate an error.

I do think it's a bit cleaner to keep the function signature as is, and let the caller use `.ok_or_not_found()` when appropriate

svix-lucho (approved):

PR #378, opened 2026-03-05
Mark .NOTES in .gitignore

svix-gabriel:
I had this in a separate PR, but it's worth just splitting out.

I like to keep things in the `.NOTES/` folder, but I don't want the `git status` pollution.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 /target
 .claude
 .vscode
-.TODO.md
+.NOTES
 logs
 .data/
 __pycache__/


.gitignore
```diff
@@ -2,7 +2,7 @@
 /target
 .claude
 .vscode
-.TODO.md
```
svix-james, line 5: Do we not want both? I don't really care, just wondering.

svix-james (approved):

.gitignore
svix-gabriel: I kind of assumed I was the only one using `.TODO.md`

PR #396, opened 2026-03-06
Queue RFC implementation

svix-gabriel:
This adds the `queue.receive` and `queue.ack` functionality from the new RFCs.

There's more endpoints to add, but this is the most important stuff.

I also went ahead and added benchmarking (so stream and queue can be benchmarked separately), and deduped some stuff between Stream and Queue.

There are larger optimizations I need to make, especially around the leases ☠️, but this PR was already getting disgustingly large, so... it's easier if those optimizations come later. Especially because my ideas apply to both stream and queue's implementation.


benchmarks/benches/queue.rs
svix-james, line 1: This is great, but we should also (maybe in followup PR, of course) add some cli benchmarks (ie, under `clients/cli/src/cmds/benchmark.rs`).

svix-gabriel: Oh yeah I forgot about those. I'll do in a followup (I hate how fat this PR has gotten)

svix-james:
This looks fine to me. If you want me to approve I will, but probably someone who's been more in the loop w/ recent stream changes should look first.

svix-gabriel:
I wouldn't mind an approval just because I want to get ahead of merge conflicts. Bugs/implementation issues are easy to fix later

svix-james (approved):

PR #397, opened 2026-03-06
Add connection/request metrics

svix-james:
* coyote.connections.total
* coyote.request.success
* coyote.request.client_error
* coyote.request.server_error
* coyote.request.duration
* coyote.request.content_length

### DIFF
--- a/src/core/metrics.rs
+++ b/src/core/metrics.rs
@@ -1,4 +1,7 @@
-use opentelemetry::metrics::{Gauge, Meter};
+use std::fmt::{self, Display, Formatter};
+
+use http::StatusCode;
+use opentelemetry::metrics::{Counter, Gauge, Histogram, Meter};
 
 use super::cluster::NodeId;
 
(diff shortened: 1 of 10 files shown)


src/core/metrics.rs
```diff
@@ -86,3 +89,107 @@ impl LogMetrics {
...
+                .build(),
+            latency: meter
+                .u64_histogram("coyote.request.duration")
+                .with_description("Request latency")
+                .with_unit("ms")
+                .build(),
+            content_length: meter
+                .u64_histogram("coyote.request.content_length")
+                .with_description("Content length")
+                .with_unit("By")
```
svix-jbrown, line 163: I know this is the OTLP standard but it's so ugly

src/core/otel_spans.rs
```diff
@@ -66,6 +100,12 @@ impl<B> MakeSpan<B> for AxumOtelSpanCreator {
             .is_valid()
             .then(|| span_context.trace_id().to_string());
 
+        let content_length = request
+            .headers()
+            .get(header::CONTENT_LENGTH)
+            .and_then(|value| value.to_str().ok())
+            .and_then(|value| value.parse::<u64>().ok());
+
```
svix-jbrown, line 108: This is not a great way to get the content length (it doesn't work with content-length-less connection: close requests, or with `Transfer-encoding: chunked`). @svix-jplatte is there any way in axum to get out of the number of body bytes that we actually read so we don't have to try to guess?

The only thing I can think of is to change the `Body` type on the request from `axum::body::Body` to something that sets a task-local variable with the number of bytes it reads, which seems intensive...

svix-james: If it's not ideal, that's fine, but does this need to block merging?

And is there any scenario right now where a request would be chunked?

src/core/metrics.rs
svix-james: What is ugly exactly?

svix-jbrown: naming the byte unit "By" even though every other unit is either a single letter or a full word

src/core/otel_spans.rs
svix-jbrown: Some clients chunk everything over a certain size (libcurl likes to do this).

I guess it doesn't have to block merging, but we really shouldn't use the content-length header for anything normative...

svix-jbrown (approved):
would still prefer if the content-length were the actual content length, but ¯\_(ツ)_/¯

src/core/metrics.rs
svix-james: fwiw I think just "B" is fine as well, but all otel examples internally use "By" -- I don't think this is OTEL, per se, it's just a description added to the metric and is based on "UCUM" conventions, whoever the heck that is...

svix-james:
ok I'm going to keep the content-length metric in, we can remove/revisit later.

PR #406, opened 2026-03-06
server: hoist monotonic time up to the top of requests

### LINKED ISSUE #405: Hoist timestamp up to the top level of requests and make sure it's monotonic
Many of our operations have a time component. Let's make sure that a monotonic timestamp is included in every raft op and pass it through to the modules.

svix-jbrown:
This does a few things:

 1. It introduces a notion of a monotonic time that won't go down even if we go to a leader whose clock is substantially behind our own. The rationale for having a non-rewinding time is described in https://github.com/svix/rfc/pull/48
 2. It hoists the operation timestamp to be a top-level property of the raft request so that modules don't need to include in every op.

Because we're internally storing time as an i64, we drop down to millisecond resolution. I think that's probably okay. The alternative would be to drop the atomic (since there's no AtomicU128) and switch to a mutex, but yuck.

Fixes #405.

### DIFF
--- a/crates/coyote-operations/src/monotime.rs
+++ b/crates/coyote-operations/src/monotime.rs
@@ -30,16 +30,21 @@ impl Monotime {
     }
 
     /// Get a monotonic version of the current time as the number of millis since epoch, updating our internal knowledge about time.
-    pub fn now_raw(&self) -> i64 {
+    fn now_raw(&self) -> i64 {
         let now = Timestamp::now().as_millisecond();
-        self.bump(now)
+        self.bump_raw(now)
     }
 
-    pub fn bump(&self, other: i64) -> i64 {
+    /// Ingest another timestamp
+    pub fn bump(&self, other: Timestamp) {
+        self.bump_raw(other.as_millisecond());
+    }
+
+    fn bump_raw(&self, other: i64) -> i64 {
         self.0.fetch_max(other, Ordering::AcqRel).max(other)
     }
 
-    pub fn as_i64(&self) -> i64 {
+    fn as_i64(&self) -> i64 {
         self.0.load(Ordering::Acquire)
     }
 }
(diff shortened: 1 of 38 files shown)


crates/coyote-operations/src/monotime.rs
```diff
@@ -0,0 +1,50 @@
...
+///
+/// This structure is cheaply cloneable and clones all point at the same underlying data
+/// (internally, this is an Arc).
+#[derive(Debug, Clone)]
+pub struct Monotime(Arc<AtomicI64>);
+
+impl Monotime {
+    pub fn initial() -> Self {
+        Self(Arc::new(AtomicI64::new(0)))
+    }
```
svix-jplatte, line 18: What's the use case for initializing with 0 / unix epoch? It seems in the one place this is constructed, it is immediately bumped. Would it not make more sense to provide an `fn new`, either with a timestamp parameter or no parameters (initializing to the current system time)?

Alternatively, maybe rename this to `unix_epoch` for clarity?

svix-jplatte (dismissed):

crates/coyote-operations/src/monotime.rs
svix-jbrown: The problem if you initialize with the current system time is that you might be initializing on a replica, and its clock might be substantially off from the rest of the cluster. Replicas clocks only advance when they receive writes.

svix-jplatte: Can you point me to where replica clocks get initialized? Or is that code yet to be written?

svix-jbrown: `Store::new` in `state_machine.rs` gets called before the cluster is joined.

svix-jplatte: But it calls `let _ = time.now();` immediately, advancing the monotime to the current system time just like the constructor I suggested.

svix-jbrown: you're right, that call is wrong

svix-jplatte: Glad I caught a bug :)

svix-jplatte (approved):

PR #415, opened 2026-03-09
Implement stream.seek

svix-gabriel:
This implements seek functionality, so you can change positions of a consumer group in a topic (in stream).

There's two ways to seek

```
{
    "position": "earliest" | "latest"
    // XOR
    "offset": 1
}
```

`position` based seeking works on both a parent topic (eg `topic="my-topic"`), and on a per-partition level (eg `topic="my-topic~5"`). `offset` based seeking only works per partition, for obvious reasons.

### DIFF
--- a/tests/it/msgs/stream_seek.rs
+++ b/tests/it/msgs/stream_seek.rs
@@ -1,11 +1,8 @@
-use std::time::Duration;
-
 use serde_json::json;
 use test_utils::{
     StatusCode, TestResult,
     server::{TestContext, start_server},
 };
-use tokio::time::sleep;
 
 #[tokio::test]
 async fn seek_earliest_replays_all_messages() -> TestResult {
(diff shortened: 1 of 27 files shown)


tests/it/msgs/stream_seek.rs
```diff
@@ -0,0 +1,513 @@
...
+        .json(json!({ "name": "ns-seek-topic" }))
+        .await?
+        .expect(StatusCode::OK);
+
+    // Configure 2 partitions
+    client
+        .post("msgs/topic/configure")
+        .json(json!({
+            "topic": "ns-seek-topic:t1",
+            "partitions": 16,
```
svix-jplatte, line 344: The comment and the code seem to contradict each other (2 partitions vs. 16 partitions)

svix-jplatte (approved):

tests/it/msgs/stream_seek.rs
svix-gabriel: Ooops yeah I'll fix that

PR #418, opened 2026-03-09
Add `key_count` metric

svix-james:
This adds metrics to get the number of keys in a namespace, scoped by 
module (which right now is only idempotency/kv/cache).

### DIFF
--- a/crates/kv/src/kvcontroller.rs
+++ b/crates/kv/src/kvcontroller.rs
@@ -161,6 +161,12 @@ impl KvController {
         KvPairRow::values(&self.keyspace)
     }
 
+    pub fn count_for_namespace(&self, namespace_id: NamespaceId) -> Result<usize> {
+        let mut prefix = vec![KvPairRow::ROW_TYPE];
+        prefix.extend_from_slice(namespace_id.as_bytes());
+        Ok(self.keyspace.prefix(prefix).count())
+    }
+
     pub fn clear_expired(&self, now: Timestamp) -> Result<()> {
         let start = ExpirationRow::key_for(NamespaceId::nil(), Timestamp::MIN, "").into_fjall_key();
         let end = ExpirationRow::key_for(NamespaceId::max(), now, "").into_fjall_key();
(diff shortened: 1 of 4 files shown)


src/core/metrics.rs
```diff
@@ -45,6 +72,28 @@ impl DbMetrics {
...
+        &self,
+        module: Module,
+        namespace: &str,
+        storage_type: StorageType,
+        count: usize,
+    ) {
+        let storage_type_str: &'static str = match storage_type {
+            StorageType::Persistent => "persistent",
+            StorageType::Ephemeral => "ephemeral",
+        };
```
svix-jplatte, line 86: Not terribly important for this PR, but would be nice to move this to an `impl From<StorageType> for opentelemetry::Value`.

svix-jplatte (dismissed):

crates/kv/src/kvcontroller.rs
```diff
@@ -161,6 +161,12 @@ impl KvController {
         KvPairRow::values(&self.keyspace)
     }
 
+    pub fn count_for_namespace(&self, namespace_id: NamespaceId) -> Result<usize> {
+        let mut prefix = vec![KvPairRow::ROW_TYPE];
+        prefix.extend_from_slice(namespace_id.as_bytes());
+        Ok(self.keyspace.prefix(prefix).count())
```
svix-jbrown, line 167: this is just using `Iterator::count()`, right? So it's going to read the entire disk? Seems expensive for metrics. I wonder if it would instead make sense to keep a counter and increment/decrement it on operations?

svix-james: Ok yeah this is actually a terrible approach. When you say counter, do you mean like an otel `Counter`, or something that's actually stored in the database that tracks totals?

I can't imagine a `Counter` would be particularly useful. But I have no idea offhand if it's practical or wise to try and update something in the database either.

svix-jbrown: I was thinking something in-memory, actually (which would require augmenting the `SerializedStateMachine` to dump and load it). Just a usize that gets incremented on set, decremented on delete / expiration, etc.

PR #420, opened 2026-03-09
repl: improve linearizable reads

svix-jbrown:
This makes it possible to do either linearizable **or** stale reads from replicas for both KV and Cache, and also to read namespaces from replicas.

### DIFF
--- a/crates/fjall-utils/src/table_row.rs
+++ b/crates/fjall-utils/src/table_row.rs
@@ -94,6 +94,12 @@ pub struct TableKey<Tag: TableRow> {
     _table: PhantomData<Tag>,
 }
 
+impl<Tag: TableRow> ::std::fmt::Debug for TableKey<Tag> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.key.fmt(f)
+    }
+}
+
 impl<'a, Tag: TableRow> TableKey<Tag> {
     /// Construct the key to be used for fjall
     ///
--- a/crates/fjall-utils/src/table_row.rs
+++ b/crates/fjall-utils/src/table_row.rs
@@ -94,6 +94,7 @@ pub struct TableKey<Tag: TableRow> {
     _table: PhantomData<Tag>,
 }
 
+#[cfg(debug_assertions)]
 impl<Tag: TableRow> ::std::fmt::Debug for TableKey<Tag> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         self.key.fmt(f)
--- a/src/v1/endpoints/cache.rs
+++ b/src/v1/endpoints/cache.rs
@@ -57,6 +57,13 @@ pub struct CacheSetOut {}
 pub struct CacheGetIn {
     #[validate(nested)]
     pub key: EntityKey,
+    /// Whether or not the read should be linearizable
+    ///
+    /// If this is `true`, the read is guaranteed to see all previous operations, but will
+    /// have to make at least one additional round-trip to the leader. If this is false, stale
+    /// reads will be performed against the replica which receives this request.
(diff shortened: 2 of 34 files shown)


src/v1/endpoints/cache.rs
```diff
@@ -57,6 +57,13 @@ pub struct CacheSetOut {}
 pub struct CacheGetIn {
     #[validate(nested)]
     pub key: EntityKey,
+    /// Whether or not the read should be linearizable
+    ///
+    /// If this is `true`, the read is guaranteed to see all previous operations, but will
+    /// have to make at least one additional round-trip to the leader. If this is false, stale
+    /// reads will be performed against the replica which receives this request.
+    #[serde(default)]
+    pub linearizable: bool,
```
tasn:
```suggestion
    pub consistent: bool,
```
I think that's more common of a name for this? E.g. dynamo uses it.

src/v1/endpoints/kv.rs
```diff
@@ -49,6 +49,13 @@ pub struct KvSetOut {}
 pub struct KvGetIn {
     #[validate(nested)]
     pub key: EntityKey,
+    /// Whether or not the read should be linearizable
+    ///
+    /// If this is `true`, the read is guaranteed to see all previous operations, but will
+    /// have to make at least one additional round-trip to the leader. If this is false, stale
+    /// reads will be performed against the replica which receives this request.
+    #[serde(default)]
+    pub linearizable: bool,
```
tasn: Same comment as above about the name, but in addition:

I think for KV the default should be true for this?

crates/fjall-utils/src/table_row.rs
```diff
@@ -94,6 +94,12 @@ pub struct TableKey<Tag: TableRow> {
     _table: PhantomData<Tag>,
 }
 
+impl<Tag: TableRow> ::std::fmt::Debug for TableKey<Tag> {
```
tasn, line 98: I'm not sure where this is used below, but we don't want to log keys, as they may contain user data.

src/v1/endpoints/cache.rs
svix-jbrown: I don't know that there's a single name (etcd calls their query paramater `quorum`). I don't love `consistent` as a name because there are many kinds of consistency and this is a specific one.

crates/fjall-utils/src/table_row.rs
svix-jbrown: I was using it to debug the totally broken namespace stuff.

src/v1/endpoints/cache.rs
tasn: Other suggestions:
- strong_consistency
- synchronous

I think probably synchronous? I just feel like linearizable is db-geek jargon that won't be immediately obvious to our target audience.

crates/fjall-utils/src/table_row.rs
tasn: As long as we remember not to accidentally log keys it's probably fine.

src/v1/endpoints/cache.rs
svix-jbrown: synchronous is way more confusing. I renamed it into an enum with hopefully enough words that anyone who knows what they're doing can find it.

tasn: FWIW `linear` is probably also better than `linearizable` as it's at least easier to spell. 😅

But looking forward to seeing what else you suggest!

crates/fjall-utils/src/table_row.rs
svix-jbrown: I put a `#[cfg(debug_assertions)]` on it so we can still use it in debug builds to make it easier to... debug... stuff, but we can't accidentally ship code that uses it.

tasn (approved):

PR #422, opened 2026-03-09
Fix namespace/topic id creation

svix-james:
Create relevant IDs before operations are applied, else
each node will create different IDs when it processes the 
operations.

### DIFF
--- a/crates/coyote-namespace/src/entities.rs
+++ b/crates/coyote-namespace/src/entities.rs
@@ -12,6 +12,15 @@ pub use fjall_utils::StorageType;
 pub type NamespaceId = Uuid;
 pub type NamespaceName = String;
 
+pub fn gen_namespace_id() -> NamespaceId {
+    let ts = jiff::Timestamp::now();
+    Uuid::new_v7(uuid::Timestamp::from_unix(
+        uuid::NoContext,
+        ts.as_second() as u64,
+        ts.subsec_nanosecond() as u32,
+    ))
+}
+
 #[derive(Serialize, Deserialize)]
 #[repr(u8)]
 pub enum Module {
--- a/crates/coyote-namespace/src/operations/create_namespace.rs
+++ b/crates/coyote-namespace/src/operations/create_namespace.rs
@@ -20,6 +20,7 @@ pub struct CreateNamespace<C: ModuleConfig> {
     storage_type: StorageType,
     max_storage_bytes: Option<NonZeroU64>,
     config: C,
+    id: NamespaceId,
 }
 
 #[derive(Deserialize, Serialize)]
@@ -36,6 +37,7 @@ pub struct CreateNamespaceOutput<C: ModuleConfig> {
 
 impl<C: ModuleConfig> CreateNamespace<C> {
     pub fn new(
+        id: NamespaceId,
         name: String,
         config: C,
         storage_type: StorageType,
@@ -43,6 +45,7 @@ impl<C: ModuleConfig> CreateNamespace<C> {
     ) -> Self {
         Self {
             timestamp: Timestamp::now(),
+            id,
             name,
             config,
             storage_type,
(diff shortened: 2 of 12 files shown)


crates/coyote-namespace/src/operations/create_namespace.rs
```diff
@@ -36,13 +37,15 @@ pub struct CreateNamespaceOutput<C: ModuleConfig> {
...
     pub fn new(
+        id: NamespaceId,
         name: String,
         config: C,
         storage_type: StorageType,
         max_storage_bytes: Option<NonZeroU64>,
     ) -> Self {
         Self {
             timestamp: Timestamp::now(),
+            id,
```
tasn, line 48: You probably want to generate it with the timestamp in timestamp.

tasn, line 48: You can generate it inside this `new()` function, which will save you from manually passing a namespace ID in kv/idempotency/etc.

Will also mean that those won't have to change.

crates/coyote-namespace/src/entities.rs
```diff
@@ -12,6 +12,15 @@ pub use fjall_utils::StorageType;
 pub type NamespaceId = Uuid;
 pub type NamespaceName = String;
 
+pub fn gen_namespace_id() -> NamespaceId {
```
tasn, line 15: Move it somewhere generic (core-utils vibe) and call it `generate_uuid` or something? No reason why it should live here.

crates/msgs/src/tables.rs
```diff
@@ -60,12 +56,12 @@ impl TopicRow {
         batch: &mut fjall::OwnedWriteBatch,
         namespace_id: NamespaceId,
         topic: &TopicName,
-        now: Timestamp,
+        maybe_topic_id: TopicId,
```
tasn, line 59: The `maybe_topic_id` is very confusing, and also feels unnecessary potentially? I feel like this is maybe something more inherent in the `msgs` module that we should change.
We know at the API layer (read: much earlier) whether there's a topic or not, so we don't need to be implicit about it, and we can pass the needed info for creating it earlier potentially.

crates/msgs/src/operations/topic_configure.rs
```diff
@@ -37,6 +38,7 @@ impl TopicConfigureOperation {
             topic,
             partitions,
             now: Timestamp::now(),
+            topic_id: gen_topic_id(),
```
tasn, line 41: Like above, we should probably use the same value as `now` to create it.

Though it's so confusing that we generate a topic id though it most likely won't be used.

crates/msgs/src/tables.rs
svix-james: I agree that this is all weird / not a great solution. In the short term I want to just fix the bug, though.

crates/coyote-namespace/src/operations/create_namespace.rs
svix-gabriel: +2

crates/coyote-namespace/src/entities.rs
svix-gabriel: +2, with the added comment that I think this should accept the `Timestamp` as a parameter to build the `Uuid`. That'll go well with the other suggested changes.

crates/msgs/src/operations/topic_configure.rs
svix-gabriel: Agreed. We don't need `maybe_topic_id` if we can easily generate the TopicId from the `now` timestamp

crates/msgs/src/operations/stream/receive.rs
```diff
@@ -46,6 +50,7 @@ impl StreamReceiveOperation {
             batch_size,
             lease_duration_millis,
             now: Timestamp::now(),
+            maybe_topic_id: gen_topic_id(),
```
svix-gabriel, line 53: Ok, my preference would be

* derive the topic ID from the same timestamp we use from `now`
* Definitely keep the topic id generation inside of StreamReceiveOperation::new(...)

crates/coyote-namespace/src/operations/create_namespace.rs
svix-james: This is not called until `apply_real` is called, though, right? So I don't see how that's an option.

PR #424, opened 2026-03-10
Kill HttpError

### LINKED ISSUE #338: Cleanup the rust error handling in the server - less and cleaner types

svix-jplatte:
Closes https://github.com/svix/coyote-private/issues/338.

Best reviewed commit by commit.


crates/coyote-error/src/lib.rs
```diff
@@ -147,9 +139,14 @@ impl<T> Traceable<T> for Result<T> {
 #[derive(Debug)]
 pub enum ErrorType {
     /// A generic error
-    Generic(String),
-    /// Any kind of HttpError
-    Http(HttpError),
+    Generic {
```
svix-gabriel, line 142: Selfishly, I wonder if we should rename this from `Generic` to `Internal`.

crates/coyote-error/src/lib.rs
```diff
@@ -147,9 +139,14 @@ impl<T> Traceable<T> for Result<T> {
...
+        message: String,
+        trace: Vec<&'static Location<'static>>,
+    },
+    /// Bad user input
+    BadRequest(StandardErrorBody),
+    /// Entity not found
```
svix-gabriel, line 148:
```suggestion
    BadRequest(StandardErrorBody),
```

Nits, but I think the enums here are self documenting enough that we don't really need these comments (I realize a lot of these comments were here before, but since we're refactoring things anyway....)

svix-gabriel (approved):

svix-gabriel:
My feedback is non blocking. Feel free to resolve the comments if you disagree or want to address them in a separate PR

PR #427, opened 2026-03-10
Rename ErrorType::{Generic => Internal} and improve docs

svix-jplatte:
Follow-up to #424, based on @svix-gabriel's feedback.

### DIFF
--- a/crates/coyote-error/src/lib.rs
+++ b/crates/coyote-error/src/lib.rs
@@ -33,8 +33,8 @@ impl Error {
     }
 
     #[track_caller]
-    pub fn generic(s: impl fmt::Display) -> Self {
-        Self::new(ErrorType::Generic {
+    pub fn internal(s: impl fmt::Display) -> Self {
+        Self::new(ErrorType::Internal {
             message: s.to_string(),
             trace: vec![Location::caller()],
         })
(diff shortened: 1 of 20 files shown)


svix-jbrown (dismissed):

crates/coyote-error/src/lib.rs
```diff
@@ -138,17 +138,21 @@ impl<T> Traceable<T> for Result<T> {
 
 #[derive(Debug)]
 pub enum ErrorType {
-    /// A generic error
-    Generic {
+    /// An unexpected error (HTTP 500)
```
tasn:
```suggestion
    /// An unexpected internal error
```
Let's mentally disconnect the notion of HTTP status codes. We shouldn't be thinking about them.

crates/coyote-error/src/lib.rs
```diff
@@ -138,17 +138,21 @@ impl<T> Traceable<T> for Result<T> {
...
+    Internal {
         message: String,
         trace: Vec<&'static Location<'static>>,
     },
+
     /// Bad user input
```
tasn: This is insufficient as an explanation. I have no idea when I should use this vs. validation, for example.

svix-jplatte: That's probably because we haven't really made any decisions and I was focused on the main goal of getting away from `HttpError` towards a much simpler error enum at first. I've added "to be further refined" for now, and we can iterate on this.

svix-gabriel (approved):

PR #437, opened 2026-03-10
server: clean up coyote-core; move shutdown and instance id there; fix Cargo.toml

svix-jbrown:
I think global properties like the shutdown token and the instance ID belong in coyote-core. This moves them there, makes it a workspace dependency, and does some related cleanup.


svix-james:
I was tinkering w/ similar changes on the PR I abandoned earlier..... it seems wrong for coyote-core to have dependencies on other crates, though, as it currently does, ie likely to create dep cycles.

svix-jbrown:
> I was tinkering w/ similar changes on the PR I abandoned earlier..... it seems wrong for coyote-core to have dependencies on other crates, though, as it currently does, ie likely to create dep cycles.

I think it's good for it to depend on coyote-error, but I am also skeptical of the dependency on coyote-namespace.

svix-james:
> I think it's good for it to depend on coyote-error, but I am also skeptical of the dependency on coyote-namespace.

Ok, leave it then. I'll remove that if I get back to the namespace-creation fix.

svix-james (dismissed):

svix-jbrown:
sorry, rebased now

I think we should move the `parse_namespace` function into `coyote-core::types` but we can do that in a different PR.

svix-james (approved):

Issue #341, opened 2026-03-03
Stream optimizations


svix-gabriel:
Specifically the performance of `receive`, `commit`, and `publish`.

PR #444, opened 2026-03-11
Add ability to configure retry schedules, DLQ topics on Queue consumer groups

### LINKED ISSUE #128: Stream/Queue API: implement changes from review

svix-gabriel:
This closes https://github.com/svix/coyote-private/issues/128, implementing the last needed endpoints for configuring consumer groups for Queues.

With queue consumer groups, we can configure
* A custom retry schedule
* A DLQ topic (where events are written when DLQ'd)

The retry schedule applies to NACKd messages. e.g. if I have a retry schedule of

`[100, 200, 400]`

then the first time I nack a message, it will go back to the main topic after 100ms. 2nd nack, it will go back after 200ms, etc. Once the retry schedule is exhausted, the msg is moved to the DLQ.

I don't think this implementation is suuuuper optimal. Cleanups coming as part of https://github.com/svix/coyote-private/issues/341


crates/msgs/src/operations/queue/nack.rs
```diff
@@ -36,24 +38,72 @@ impl QueueNackOperation {
...
         let topic_row = TopicRow::fetch(
             &state.metadata_tables,
             TopicRow::key_for(self.namespace_id, &self.topic),
         )?
         .ok_or_else(|| Error::invalid_user_input("topic must exist"))?;
 
+        let config = QueueConfigRow::fetch(
+            &state.metadata_tables,
+            QueueConfigRow::key_for(topic_row.id, &self.consumer_group),
+        )?;
```
svix-jbrown, line 55: I wonder if it would be worth keep an LRU cache of these (and invalidating them in the relevant apply, obv) so we can avoid hitting the database for them all the time?

crates/msgs/src/operations/queue/nack.rs
```diff
@@ -62,11 +112,63 @@ impl QueueNackOperation {
...
+    batch.insert_row(
+        &state.metadata_tables,
+        QueueLeaseRow::key_for(source_topic_id, msg_id, consumer_group),
+        &QueueLeaseRow {
+            expiry: Timestamp::MAX,
+            dlq: false,
+            retry_count: 0,
+        },
+    )?;
+
```
svix-jbrown: I think this should probably be an `ack` method since you do it in two places now.

crates/msgs/src/tables.rs
```diff
@@ -119,6 +120,8 @@ pub(crate) struct QueueLeaseRow {
     pub expiry: Timestamp,
     #[serde(default)]
     pub dlq: bool,
+    #[serde(default)]
+    pub retry_count: u32,
```
svix-jbrown: I personally find `retry_count` somewhat confusing and might prefer to store it as `attempt_count` (and increment everything else by one).

crates/msgs/src/operations/queue/nack.rs
svix-gabriel: I think that's reasonable. There's a lot I want to clean up across the module w/reducing reads + writes of leases, so I think it'll be easier to save it for a future PR

svix-jbrown (approved):

PR #451, opened 2026-03-11
kv: implement concurrent expiration pruning

svix-jbrown:
This implements a new algorithm for concurrently and safely (I believe) pruning expired data from Kv stores. It's only enabled for `Kv` itself for now. This is inspired by the algorithms underlying concurrent mark-sweep garbage collectors (which is effectively what this is).

This is an implementation of Proposal 2 from svix/rfc#48

The main pieces are as follows:

 1. The background worker is now invoked on the same `KvState` as the applier (well, an arc clone of it, but with the same pieces).
 2. The background worker now uses the monotonic time introduced in #406 so that it knows time can never rewind
 3. The background worker uses a new set of mutex-based fences to do mark-sweep, designed such that the "mark" phase can be concurrent with other writes (== raft applies), although the sweep phase cannot. This would obviously be a lot simpler if we just put the whole thing inside a mutex and inhibited all writes during both mark and sweep, but I'm trying to make it nicer than that.

Obviously all of the complexity is in how to do value coloring to implement mark/sweep correctly. I put lots of comments in the files, but when thinking about it, the inteleavings I found most helpful are:

 - start_marking, intent_to_gc, want_to_write, drain _(in this case, we first add the value to the protected set, then add it to the GC set, then don't operate on it because of the intersection)_
 - start_marking, want_to_write, intent_to_gc, drain _(in this case, we first add the value to the GC set, then to the protected set, then we don't operate on it because of the intersection)_
 - want_to_write, start_marking, intent_to_gc, drain _(the ReadLock guard in want_to_write ensures that start_marking can't run until the outstanding writes have finished)_
 - start_marking, intent_to_gc, drain, want_to_write _(the guard in drain ensures that want_to_write can't start until we finish draining the current chunk)_


svix-jbrown:
closing this because I'm going to make a new one with a somewhat different approach

PR #455, opened 2026-03-12
convert kv, cache, and idempotency modules to use raft for replicating cleanup actions

### LINKED ISSUE #83: allow background workers to only run on the leader and write through consensus
Right now, workers run on every node and directly mutate state everywhere. We should make it so that leaders can ergonomically ask to only run on the leader and that they can emit log entries.

svix-jbrown:
This is an alternative to #451 that does expirations in batches through Raft.

This is an implementation of Anti-Proposal 2 from svix/rfc#48, with a per-tick limit in place to prevent causing too many write slowdowns. I still kind of think #451 is a better approach, because it can run concurrently with other writes instead of serially with them, but this is definitely much less scary.

The way of threading the needle I chose for exposing Raft to the modules is to use a new trait. I really don't like the extra `F::Request` bounds that I had to put on the workers, but I couldn't figure out a way around it.

Fixes #83


svix-jplatte:
> The way of threading the needle I chose for exposing Raft to the modules is to use a new trait. I really don't like the extra `F::Request` bounds that I had to put on the workers, but I couldn't figure out a way around it.

You called? I made #456 to fix this.

crates/cache/src/lib.rs
```diff
@@ -29,8 +30,16 @@ pub struct State {
 impl State {
     pub fn init(dbs: Databases) -> Result<Self> {
         Ok(Self {
-            persistent_controller: KvController::new(dbs.persistent, CACHE_KEYSPACE),
-            ephemeral_controller: KvController::new(dbs.ephemeral, CACHE_KEYSPACE),
+            persistent_controller: KvController::new(
+                StorageType::Persistent,
```
tasn: Small nitpick: but we are only using it for printing a log, maybe we should just pass it explicitly to "clean_expired" so that we use it there rather than having it as state on the controller?

It doesn't feel like something the controller should be aware of (though maybe, dunno).

crates/cache/src/lib.rs
```diff
@@ -42,37 +51,50 @@ impl State {
...
+}
 
-    pub value: Vec<u8>,
+#[tracing::instrument(skip_all)]
+pub async fn worker_loop<F>(state: &State, writer: &F) -> coyote_operations::BackgroundResult<()>
+where
+    F: OperationWriter<operations::CacheOperation>,
+{
+    writer
+        .write_request(operations::ClearExpiredOperation::new(
```
tasn: So it sends a raft request every 1 second whether we have some to clean or not? Feels wasteful, especially since we do it one per storage type times all the modules.
It also feels like if we have more than 10,000 to clean, we may start getting a backlog?

Alternative idea: maybe instead of having it go over raft, have it locally in a module, but have it use the raft timestamp (now that we have it). So, for example, you can have the worker just run the cleanup task once a second (or whatever) and you can be certain it does exactly the same thing, as it uses the timestamp from the cluster to do it.

svix-jbrown: That's what the other PR does, but it's complicated because of concurrency.

I guess I could modify it to only send if the exist any expired items.

tasn: Let me look at the other PR, but other than locking on store (which this one does anyway implicitly, because it blocks raft for all the modules, the other idea at least locks just the relevant stores), what's complicated there? We also don't really need to lock until we've even decided there's something to wipe, so we actually only lock if we know we need to wipe stuff (and then we can look again at what to wipe exactly).

tasn: OK, looking at the other one, the complexity comes from:
> The background worker uses a new set of mutex-based fences to do mark-sweep, designed such that the "mark" phase can be concurrent with other writes (== raft applies), although the sweep phase cannot. This would obviously be a lot simpler if we just put the whole thing inside a mutex and inhibited all writes during both mark and sweep, but I'm trying to make it nicer than that.

Though if instead of mark and sweep, you literally just do:
1. WITHOUT LOCKING: scan the db if there are any items that should be expired. If no, sleep, if yes, continue.
2. Lock just this specific store/controller/whatever the module uses.
3. Find all the items to wipe, and wipe them (up to 10,000 items), the yield. If we had less than 10k items, sleep, otherwise sleep less and continue after the yield point.

This feel simple, creates less contention on raft (essentially only creates contention if the same module is being used, and only if there's anything to wipe), gives us the ability to change when we call it dynamically (e.g. stop doing it under load), and doesn't pollute raft.

crates/cache/src/lib.rs
svix-jbrown: I don't have strong feelings. It's mostly in there for debugging. I can make it an arg.

crates/cache/src/lib.rs
svix-jbrown: I think we might go down that road in the future, but the details of how to do that locking without either resulting in **many second** pauses during garbage collection, or doing more-sophisticated fine-grained locking are very complicated. That's why I want to land this PR for now.

tasn: So exactly the same approach as the other PR, just simplify the mark and sweep part with a simpler version.

svix-jbrown: I would still like to land this one prior to doing more work on #451.

tasn: > I would still like to land this one prior to doing more work on https://github.com/svix/coyote-private/pull/451.

Aren't these two changes opposites though? As in, none of this code is useful if you do 451, no?

> I think we might go down that road in the future, but the details of how to do that locking without either resulting in many second pauses during garbage collection, or doing more-sophisticated fine-grained locking are very complicated. That's why I want to land this PR for now.

We already do this locking just now, just implicitly. By being a raft operation we lock the whole cluster while we process it, so the current implementation is already the worst case scenario, no?

svix-jbrown: Most of this code continues to be useful if we do #451, both for other modules that want to write and for the improvements to expiry. It's really just the worker code (which is about 8 useful lines, copy-pasted 3 times) that changes.

tasn (approved):

crates/cache/src/lib.rs
tasn: I don't see it, it looks to me like most of the code is the `writer` stuff, having worker only run on leader, and the operations. Though I'm sure you know this code much better than I do, so approving. Anyway, no need to hold out on this, as worst case we just delete a bit more in #451.

PR #456, opened 2026-03-13
Move some bounds around to DRY module code

svix-jplatte:
I accepted @svix-jbrown's challenge from #455.


svix-gabriel (approved):

svix-jbrown (approved):
ah, two traits. I didn't think of it that way! Let's merge this into my branch and then merge my branch, shall we?

svix-jplatte:
Actually if you look at the first commit, you can see it worked before I split the trait in two.

PR #457, opened 2026-03-13
Ratelimit: various changes to API and internals

tasn:


crates/rate-limit/src/lib.rs
```diff
@@ -258,230 +61,3 @@ where
         tokio::time::sleep(Duration::from_secs(1)).await;
     }
 }
```
svix-jbrown, line 63: this is going to get rewritten after #455 lands anyway

svix-jbrown (approved):

crates/rate-limit/src/lib.rs
tasn: Yeah, it's half a placeholder.

PR #459, opened 2026-03-13
Add k8s operator and supporting helm chart

svix-james:
Adds a basic operator and Helm chart for deploying it.

Right now this is being manually deployed to Dev, but it
is deploying a stable-ish cluster that I can run benchmarks
against, so I think it's worth getting this code merged to have
a starting point.


infra/cluster.yaml
svix-james, line 1: Right now this is the cluster manifest, which I'm manually pushing to Dev.

infra/crd.yaml
svix-james, line 1: Same w/ CRD

.pre-commit-config.yaml
```diff
@@ -13,12 +13,14 @@ repos:
       - id: check-toml
       - id: check-yaml
         args: ["--allow-multiple-documents"]
+        exclude: "^infra/helm"
```
svix-james, line 16: Linting doesn't like Helm's templated yaml files.

tasn (approved):

svix-jplatte (approved):
Rubberstamping as per discussion in https://svixhq.slack.com/archives/C02QBTJSY3Y/p1773760361529549

PR #466, opened 2026-03-14
Kv: return success status on insert/update conflicts.

tasn:
We should return an indication when we fail to insert/update because keys already exist/don't exist. We definitely shouldn't fail silently.

Also changed the `deleted` in kv/cache responses to say `success` to match these.

Discussion on Slack: https://svixhq.slack.com/archives/C0A72988962/p1773457868083409

### DIFF
--- a/crates/kv/src/kvcontroller.rs
+++ b/crates/kv/src/kvcontroller.rs
@@ -113,7 +113,7 @@ impl KvController {
         expiry: Option<Timestamp>,
         behavior: OperationBehavior,
         now: Timestamp,
-    ) -> Result<()> {
+    ) -> Result<bool> {
         match behavior {
             OperationBehavior::Upsert => {
                 self.insert_with_expiration(namespace_id, key, value, expiry)?;
@@ -124,7 +124,7 @@ impl KvController {
                 if !exists {
                     self.insert_with_expiration(namespace_id, key, value, expiry)?;
                 } else {
-                    // FIXME(@svix-lucho): Do nothing?
+                    return Ok(false);
                 }
             }
             OperationBehavior::Update => {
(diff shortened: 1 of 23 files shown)


crates/kv/src/kvcontroller.rs
```diff
@@ -113,7 +113,7 @@ impl KvController {
         expiry: Option<Timestamp>,
         behavior: OperationBehavior,
```
tasn, line 114: 1. I called the value `success` though maybe `done` is a better name? or `ok`?
2. Is it clear in the context of deleted and the other operation what this actually means? The operation was successful after all, but it just didn't do the full operation based on the conditional/behavior.

svix-jbrown (commented):
you should probably add docstring comments to these `success` fields so that the openapi spec says what they actually indicate

tasn:
Good call, pushing now!

Though having now written the docs it's even more obvious we really need to figure out https://svixhq.slack.com/archives/C0A72988962/p1773457868083409 as well for this. Probably not a blocker for this PR (as this is better than what's there), but still important to figure out.

crates/kv/src/kvcontroller.rs
svix-lucho: What about `updated`? I'm worried `success`, `done`, `ok` can be confused with the operation failing

tasn: We defo need a better word, though I think that's a different discussion (that we should have!). See my comment: https://github.com/svix/coyote-private/pull/466#issuecomment-4067484471

tasn: Opened https://github.com/svix/rfc/pull/50

svix-jbrown (approved):

PR #468, opened 2026-03-14
KV: implement opportunistic concurrency control.

tasn:
We allow passing a `version` field which is evaluated against the known version we have on disk. If they match it goes through, otherwise it aborts the request.

At the moment it throws an error on version mismatch, which I think makes sense, but let's see how this discussion evolves: https://svixhq.slack.com/archives/C0A72988962/p1773457868083409

It's currently only implemented for kv. I need to think whether it makes sense to implement it for `cache` as well. We do it in svix webhooks private to ensure we cache the right value in highly contended environments (CMA). Though I wonder whether there are more elegant things we can do.

### DIFF
--- a/clients/go/internal/models/kv_get_out.go
+++ b/clients/go/internal/models/kv_get_out.go
@@ -9,4 +9,7 @@ import (
 type KvGetOut struct {
 	Expiry *time.Time `json:"expiry,omitempty"` // Time of expiry
 	Value  []uint8    `json:"value,omitempty"`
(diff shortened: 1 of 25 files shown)


svix-gabriel (dismissed):

svix-jbrown (dismissed):

tasn:
Alas, this PR is part of a chain and will need to be rebased. :(

svix-jbrown (approved):

Issue #155, opened 2026-02-19
Mark durations as such in OpenAPI spec


svix-jplatte:
Right now, they are just integers as far as the spec (and thus codegen) is concerned. I'm pretty sure JSON Schema offers some way of declaring custom named types (not sure whether we can easily make use of it with schemars though).

tasn:
Related: #160

svix-jplatte:
I couldn't actually find anything about custom named types. `"format": "duration"` exists but applies to duration *strings* as described in [RFC3339](https://www.rfc-editor.org/rfc/rfc3339.txt). I'll go ahead with #156 by just looking at the `_ms` / `_millis` suffix (whichever one we decide for) and type.

PR #473, opened 2026-03-16
Add DurationMs, DurationS types and use in API

svix-jplatte:
Part of https://github.com/svix/coyote-private/issues/155.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -1428,8 +1428,7 @@
           "ttl": {
             "description": "Time to live in milliseconds",
             "type": "integer",
-            "format": "uint64",
-            "minimum": 0
+            "format": "uint64"
           }
         },
         "required": [
@@ -2255,7 +2254,6 @@
           "lease_duration_millis": {
             "type": "integer",
             "format": "uint64",
-            "minimum": 0,
             "default": 30000
           }
         },
@@ -2351,7 +2349,6 @@
           "lease_duration_millis": {
             "type": "integer",
             "format": "uint64",
-            "minimum": 0,
             "default": 300000
           }
         },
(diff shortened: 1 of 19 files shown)


svix-gabriel (dismissed):

openapi.json
```diff
@@ -1428,8 +1428,7 @@
           "ttl": {
             "description": "Time to live in milliseconds",
             "type": "integer",
-            "format": "uint64",
-            "minimum": 0
```
tasn, line 1432: This seems wrong, no? We probably want the min 0?

openapi.json
```diff
@@ -2212,7 +2211,6 @@
           "lease_duration_millis": {
             "type": "integer",
             "format": "uint64",
-            "minimum": 0,
             "default": 30000
```
tasn, line 2257: I would have expected more stuff in the spec file following this change, no?

openapi.json
svix-jplatte: The format already says it's unsigned, so the minimum was redundant. I can add it back though if you prefer. Let me know.

openapi.json
svix-jplatte: I like minimal commits / PRs. Will do type tagging as followup.

openapi.json
tasn: I don't think it's redundant? Because the format is more of a "oh hey, if you know what this is, this is how you can extra special use that" the minimum is for validation.
Though anyway, we probably want format to be `duration` or something, so I guess for that it really is >0 anyway.

tasn (approved):

PR #478, opened 2026-03-16
add administrative cluster APIs and custom CLI

### LINKED ISSUE #237: add cluster management CLI tools
Right now, there's no way to manage the cluster. We should provide a CLI tool for doing so (this could just be a subcommand of `coyote-server`).

Some things we need to support:

 - List the nodes in the cluster
 - Remove a node from the cluster (required if a node dies for some reason)

svix-jbrown:
This does a couple of things:

 1. It adds new (public, non-interserver-port) APIs for removing a node from the cluster and for displaying the cluster status
 2. It adds new CLI commands for interacting with these APIs which have non-standard behavior so are hand-implemented instead of generated.
 3. It modifies the removal logic to be more reliable; in particular, there isn't any safe way to add a removed node from the cluster without wiping it entirely, so we now record a new "poisoned" state and refuse to start up with that database if it's been poisoned.

```plain_text
$ ./target/debug/coyote cluster-admin status
Cluster Information

 Cluster ID    a953df88a2924d93ab7f6dc1ddce4316 
 Cluster Name  coyote                           

Responding Server

 Server ID                 3e7cc67b26e348f790787f2222f79e54     
 Server State              Leader                               
 Last Committed Timestamp  2026-03-16T21:59:44.167Z (253ms ago) 

Nodes

┌──────────────────────────────────┬──────────────────────┬──────────┬──────────────────┐
│ Node ID                          ┆ Address              ┆ State    ┆ Last Transaction │
╞══════════════════════════════════╪══════════════════════╪══════════╪══════════════════╡
│ 0003e893f62c41bc83ba452996e404e6 ┆ http://0.0.0.0:18051 ┆ Follower ┆ 1-19             │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ 3e7cc67b26e348f790787f2222f79e54 ┆ http://0.0.0.0:18050 ┆ Leader   ┆ 1-19             │
└──────────────────────────────────┴──────────────────────┴──────────┴──────────────────┘
```

```
$ ./target/debug/coyote cluster-admin remove-node 0003e893f62c41bc83ba452996e404e6
Are you sure you want to remove node 0003e893f62c41bc83ba452996e404e6 (http://0.0.0.0:18051) (y/N)? y
```

Fixes #237.


src/core/cluster/handle.rs
```diff
@@ -427,6 +481,152 @@ impl RaftState {
...
+        let Some(leader_id) = self.raft.current_leader().await else {
+            return Err(CoyoteErrorOrForwardToLeader::internal(
+                "unable to determine leader",
+            ));
+        };
+        if leader_id != self.node_id {
+            let Some(leader_node) = self
+                .raft
+                .with_raft_state(move |s| {
+                    s.membership_state.effective().get_node(&leader_id).cloned()
```
svix-james, line 537: `effective` vs `committed`.... I have no idea which one we should be using / when / why

svix-james:
Any reason not to add some basic tests for some of this functionality?

src/core/cluster/handle.rs
svix-jbrown: They're nearly always the same. The effective state contains nodes which are in the process of joining the cluster but haven't voted on a committed log, and the committed state does not.

svix-jbrown:
> Any reason not to add some basic tests for some of this functionality?

nope, just added some

crates/test-utils/src/server.rs
```diff
@@ -82,24 +104,25 @@ impl TestServerBuilder {
         let repl_listener = if let Some(listener) = self.repl_listener {
             listener
         } else {
-            TcpListener::bind("0.0.0.0:0").await.unwrap()
+            TcpListener::bind("127.0.0.1:0").await.unwrap()
```
svix-james, line 107: Previously this was causing something to fail.... not sure if it's been fixed, but I had to do this here: https://github.com/svix/coyote-private/pull/123#discussion_r2789763693

svix-jbrown: I think that's because we used to only discover peer addresses by enumerating the system's IPs and using the first one; as long as `seed_nodes` is populated, though, you can use whatever you want (as evidenced by this PR. :-) )

svix-james (approved):

clients/cli/src/cmds/cluster.rs
```diff
@@ -0,0 +1,162 @@
...
+#[derive(Args)]
+#[command(args_conflicts_with_subcommands = true, flatten_help = true)]
+pub struct ClusterAdminArgs {
+    #[command(subcommand)]
+    pub command: ClusterCommands,
+}
+
+#[derive(Args)]
+pub struct RemoveNodeArgs {
+    #[arg()]
```
svix-jplatte, line 16: Pretty sure this is a no-op.

clients/cli/src/cmds/cluster.rs
```diff
@@ -0,0 +1,162 @@
...
+    let mut table = Table::new();
+    let rows = raw_status
+        .nodes
+        .iter()
+        .sorted_unstable_by_key(|r| &r.node_id)
+        .map(|row| {
+            let term = if let Some(term) = row.last_committed_term
+                && let Some(index) = row.last_committed_log_index
+            {
+                format!("{}-{}", term, index)
```
svix-jplatte, line 117:
```suggestion
                format!("{term}-{index}")
```

crates/test-utils/src/server.rs
```diff
@@ -43,9 +45,29 @@ impl TestServerBuilder {
...
+            repl_listener: None,
+        }
+    }
+
+    pub fn tap_cfg(mut self, f: impl FnOnce(&mut ConfigurationInner)) -> Self {
+        if let Some(cfg) = &mut self.cfg {
+            f(cfg)
+        }
+        self
+    }
```
svix-jplatte, line 69: Is it ever not a bug for this to be called when `self.cfg` is `None`? I think we should guard against that, maybe just by changing this to `fn cfg_mut(&mut self) -> &mut ConfigurationInner` which panics if it's `None`.

crates/test-utils/src/server.rs
```diff
@@ -151,37 +181,72 @@ async fn check_initialized(client: &reqwest::Client, url: &str) -> anyhow::Resul
...
             Err(_) => anyhow::bail!("timed out waiting for server to boot"),
         }
+        // Then, make sure that the regular comms port is up
+        match tokio::time::timeout_at(
+            deadline,
+            client
+                .get(&main_url)
+                .timeout(Duration::from_millis(10))
+                .send()
+                .and_then(|r| futures_util::future::ready(r.error_for_status())),
```
svix-jplatte, line 234: This is available in `std` (`std::future::ready`), no need to pull futures-util for it.

svix-jplatte: Or alternatively, just use an async block? Something like

```rust
async {
    let res = client.get(&main_url).timeout(Duration::from_millis(10)).send().await?;
    Ok(r.error_for_status())
}
```

src/core/cluster/state_machine.rs
```diff
@@ -165,6 +180,12 @@ impl Store {
             time.bump(timestamp);
         }
 
+        if logs.is_poisoned().await? {
+            anyhow::bail!(
+                "this node was previously removed from a cluster and must be erased before it can be re-added"
+            );
+        }
```
svix-jplatte, line 187: Can use `anyhow::ensure!`.

src/v1/endpoints/admin.rs
```diff
@@ -0,0 +1,219 @@
...
+            (this_node_state, committed, nodes)
+        })
+        .await
+        .or_internal_error()?;
+    let cluster_id = repl.state_machine.cluster_id().await;
+    let cluster_name = if cluster_id.is_some() {
+        Some(app_state.cfg.cluster.name.to_owned())
+    } else {
+        None
+    };
```
svix-jplatte, line 139: nitpick: could use `bool::then`, or `cluster_id.map(|_| ...)`.

src/v1/endpoints/admin.rs
```diff
@@ -0,0 +1,219 @@
...
+            let members = s.membership_state.effective().membership();
+            let voters = members.voter_ids().collect::<BTreeSet<NodeId>>();
+            let learners = members.voter_ids().collect::<BTreeSet<NodeId>>();
+            let nodes = members
+                .nodes()
+                .map(|(node_id, node)| {
+                    let node_id = *node_id;
+                    let node = node.clone();
+                    let state = if let Some(id) = &leader_id
+                        && id == &node_id
```
svix-jplatte, line 113: This can probably be `if leader_if == Some(node_id)`?

clients/cli/src/cmds/cluster.rs
svix-jplatte: Feel free to address all of my comments post-merge, by the way.

crates/test-utils/src/server.rs
svix-jbrown: I'm going to simplify it in the next PR

PR #480, opened 2026-03-16
server: add explicit panic handling to leader-background tasks

### LINKED ISSUE #134: Add "panic catchers" in the background tasks so that we can recover from panics
Kind of what axum does for HTTP handlers.

We don't want background tasks to stop working all of a sudden, or bring the whole sever down when they go down.

svix-jbrown:
fixes #134


src/core/cluster/background.rs
```diff
@@ -73,73 +89,129 @@ impl BackgroundJob for Tick {
...
+                }
+            }
+            shutting_down.run_until_cancelled(backoff.backoff()).await;
+        }
+    });
 }
 
 impl BackgroundJobRunner {
     fn spawn_all(cfg: Configuration, handle: RaftState) -> Self {
         let mut jobs = JoinSet::new();
```
svix-jplatte, line 190: Could also construct `Self` first here with an empty `JoinSet`, and make `spawn_job` a method.

svix-jplatte (approved):

Issue #392, opened 2026-03-06
Add positional parameters to Go, Java, JS


svix-jplatte:
Python and Rust have it so it's only a matter of repeating the same sort of changes for the different codegen templates.

Issue #183, opened 2026-02-24
Go SDK: switch to msgpack


svix-jplatte:
Because of course it does. Maybe just solve this by switching from json to msgpack?!

tasn:
I changed the title, as we don't care about the other thing.

PR #486, opened 2026-03-17
ci: Add java build job

svix-jplatte:
Make sure this compiles when we change codegen (e.g. for #183, #392).

### DIFF
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -1,0 +1,21 @@
+name: Java
+
+on:
+  workflow_dispatch:
+  workflow_call:
+  push:
+    branches: [main]
+    paths:
+      - '.github/workflows/java.yml'
+      - 'clients/java/**'
+
+jobs:
+  build:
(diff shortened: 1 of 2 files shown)


.github/workflows/java.yml
```diff
@@ -0,0 +1,21 @@
+name: Java
+
+on:
+  workflow_dispatch:
+  workflow_call:
+  push:
+    branches: [main]
+    paths:
+      - '.github/workflows/java.yml'
+      - 'clients/java/**'
```
svix-jbrown, line 10: I don't love having all these little workflows invoked on push, because it means we can't gate releases on completion of the workflows. can we move this to a central dispatch workflow like we have for PR

svix-jplatte: Sure, I have something partially ready for that locally, can it be a follow-up PR?

svix-jbrown: ok

svix-jbrown (approved):

PR #497, opened 2026-03-17
Make `/repl/raft/last-id` a POST request

svix-james:
This method currently fails in benchmarking with a 405. Everything else is a POST, so I'm guessing this was just an oversight?


svix-jbrown (approved):
yep! fun fact, this fix is also in #478

PR #499, opened 2026-03-17
Build cli docker image

svix-james:
This tries to reuse the server Dockerfile as much as possible, adding a separate target for cli that will be invoked separately in github.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,6 +41,7 @@ COPY --from=planner /app/recipe.json recipe.json
 
 # Build dependencies - this is the caching Docker layer
 RUN cargo chef cook --release --package coyote-server --features coyote/openapi --recipe-path recipe.json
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -90,3 +92,33 @@ EXPOSE 8050
 COPY --chown=root:root --chmod=755 --from=build /app/target/release/coyote-server /usr/local/bin/coyote-server
```
svix-jbrown, line 92: we should probably also put the CLI into the server image so people can do stuff to the running server from interactive shells

```suggestion
COPY --chown=root:root --chmod=755 --from=build /app/target/release/coyote-server /usr/local/bin/coyote-server
COPY --chown=root:root --chmod=755 --from=build /app/target/release/coyote /usr/local/bin/coyote
```

svix-jbrown (approved):

PR #502, opened 2026-03-17
server: add panic handling to the HTTP servers

svix-jbrown:
Apparently, this isn't "free" in axum so you need to use Tower.

This does a few things:

 1. Adds a panic handler that returns msgpack or JSON (based on the tokio task-local storage property) with a basic body
 2. Changes the `coyote_error::Error::internal` body to have an `error` field instead of just being empty JSON
 3. Adds new `/api/v1/health/error` and `/api/v1/health/panic` endpoints to force a request to error or panic. The panic endpoint is undocumented and only built in debug mode.

related to #134

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,19 +7,23 @@
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::{sync::Arc, time::Duration};
 
+use ::serde::Serialize;
 use aide::axum::ApiRouter;
-use axum::{Extension, extract::DefaultBodyLimit, middleware, serve::ListenerExt as _};
+use axum::{
+    Extension, extract::DefaultBodyLimit, middleware, response::IntoResponse as _,
+    serve::ListenerExt as _,
+};
 use coyote_core::Monotime;
 use coyote_error::Error;
 use fjall_utils::{Databases, ReadonlyDatabases};
 use opentelemetry::metrics::Meter;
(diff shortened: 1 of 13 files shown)


src/lib.rs
```diff
@@ -89,6 +93,21 @@ pub struct AppState {
     pub(crate) time: Monotime,
 }
 
+fn handle_panic(_err: Box<dyn std::any::Any + Send + 'static>) -> axum::response::Response {
```
svix-jplatte: The panic payload is almost always a string. We should downcast and include it in the tracing event if it's a string.

src/v1/endpoints/health.rs
```diff
@@ -20,8 +23,31 @@ async fn ping() -> Result<MsgPackOrJson<PingOut>> {
...
-    ApiRouter::new().api_route_with("/health/ping", get_with(ping, ping_operation), &tag)
+    let router = ApiRouter::new()
+        .api_route_with("/health/ping", get_with(ping, ping_operation), &tag)
+        .api_route_with("/health/error", post_with(error, error_operation), &tag);
+
+    if cfg!(debug_assertions) {
+        router.route("/health/panic", post(panic))
+    } else {
+        router
+    }
```
svix-jplatte:
```suggestion
    #[cfg(debug_assertions)]
    let router = router.route("/health/panic", post(panic));

    router
```

svix-jplatte (approved):

PR #507, opened 2026-03-18
Operator Docker: Fix the `cargo chef` step

svix-james:
Docker build broke after moving this to be part of the workspace. Should be fixed now.

### DIFF
--- a/infra/operator/Dockerfile
+++ b/infra/operator/Dockerfile
@@ -40,7 +40,7 @@ EOF
 COPY --from=planner /app/recipe.json recipe.json
 
 # Build dependencies - this is the caching Docker layer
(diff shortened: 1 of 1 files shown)


svix-jplatte (approved):

svix-jplatte:
I have approved, but some things are definitely still off here:

- `cargo chef cook` vs `cargo build` should use the same flags, right now one of them uses `--bin coyote-operator` (in addition to `-p coyote-operator) and the other doesn't. Right now it's not going to make a difference because that Cargo package only contains the binary, but it might in the future.

- The `Dockerfile` says to use its parent directory as the context directory, but that's not what CI does and if one tries it, it fails

PR #513, opened 2026-03-18
Tokio: bring back multithreaded runtime.

### LINKED ISSUE #495: Try multithreaded tokio runtime
… to see if it improves performance.

tasn:
We changed it to the single threaded one at some point because we thought we may go down a certain architecture path. Though we haven't gone down that path, so it's no longer needed to do the single threaded.

If we ever want to do that architecture change or we think this can potentially improve performance somehow, we can change it. Though at the moment it feels unnecessary and limiting (e.g. no concurrent kv.get).

Fixes #495

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -200,7 +200,7 @@ sysinfo = "0.38.2"
 tap = "1"
 tempfile = "3.24.0"
 test-utils.path = "crates/test-utils"
-tokio = { version = "1.24.2", features = ["rt", "signal", "fs"] }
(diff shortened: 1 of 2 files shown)


svix-jplatte (approved):

tasn:
For those who are curious: more than 2x throughput increase and 6x latency decrease in the benchmarks (which makes sense, and is very obvious).
Writes are not affected as raft is single-threaded at the moment, but that too will be fixed soon.

PR #518, opened 2026-03-18
Operator: Add health checks to containers

svix-james:
Pretty straightforward. Also update the makeshift deploy documentation

### DIFF
--- a/infra/crd.yaml
+++ b/infra/crd.yaml
@@ -47,14 +47,14 @@ spec:
                   type: string
                 cluster:
                   default:
-                    secretRef: null
-                    heartbeatIntervalMs: null
-                    electionTimeoutMinMs: null
                     electionTimeoutMaxMs: null
-                    replicationRequestTimeoutMs: null
-                    snapshotAfterWrites: null
-                    snapshotAfterMs: null
+                    electionTimeoutMinMs: null
+                    heartbeatIntervalMs: null
                     logLevel: null
+                    replicationRequestTimeoutMs: null
+                    secretRef: null
+                    snapshotAfterMs: null
+                    snapshotAfterWrites: null
                   description: Cluster/replication configuration.
                   properties:
                     electionTimeoutMaxMs:
@@ -149,8 +149,8 @@ spec:
                   type: object
                 resources:
                   default:
-                    requests: null
                     limits: null
+                    requests: null
                   description: CPU and memory resource requests and limits for the coyote container.
                   properties:
                     limits:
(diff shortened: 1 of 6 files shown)


infra/crd.yaml
svix-james, line 1: The yaml formatting here sucks. Working on making it consistent.

infra/operator/src/resources/statefulset.rs
```diff
@@ -263,7 +265,30 @@ fn build_container(
...
+                path: Some(HEALTH_CHECK_ENDPOINT.into()),
+                port: IntOrString::Int(cluster_port as i32),
+                ..Default::default()
+            }),
+            initial_delay_seconds: Some(5),
+            period_seconds: Some(10),
+            failure_threshold: Some(6),
+            success_threshold: Some(1),
+            ..Default::default()
+        }),
```
svix-jbrown, line 280: this should probably be the `/api/v1/health/ping` endpoint on the main port; we don't want the pod to be killed while it's joining the cluster at bringup

infra/operator/src/resources/statefulset.rs
```diff
@@ -280,7 +280,7 @@ fn build_container(
         readiness_probe: Some(Probe {
             http_get: Some(HTTPGetAction {
                 path: Some(HEALTH_CHECK_ENDPOINT.into()),
-                port: IntOrString::Int(cluster_port as i32),
+                port: IntOrString::Int(spec.api_port as _),
                 ..Default::default()
```
svix-jbrown, line 285: sorry, that previous comment was unclear

the **liveness** probe should use `/api/v1/health/ping` (because it shouldn't bring down the pod just because the cluster isn't booted)

the **readiness** probe should continue to use `/repl/health` (because we shouldn't route traffic to it until it's booted).

svix-james: Done.

infra/operator/src/resources/statefulset.rs
```diff
@@ -263,7 +266,30 @@ fn build_container(
...
+            initial_delay_seconds: Some(5),
+            period_seconds: Some(10),
+            failure_threshold: Some(6),
+            success_threshold: Some(1),
+            ..Default::default()
+        }),
+        readiness_probe: Some(Probe {
+            http_get: Some(HTTPGetAction {
+                path: Some(CLUSTER_HEALTH_ENDPOINT.into()),
+                port: IntOrString::Int(spec.api_port as _),
```
svix-jbrown: sorry

```suggestion
                port: IntOrString::Int(cluster_port as _),
```

svix-jbrown (approved):

Issue #492, opened 2026-03-17
experiment with running KV pruning on each node instead of through raft


svix-jbrown:
This was originally in #451 and got superceded by the logically-simpler #455.

PR #520, opened 2026-03-18
cluster: run kv cleanups in the background

svix-jbrown:
This is an alternative to #451 which skips the fancy locking in favor of just a big mutex around the whole controller.

I've only done this for `kv` but if it looks okay I can make the same change to the other crates that use `kvcontroller` (cache and idempotency) pretty easily.

Part of #492

### DIFF
--- a/crates/kv/src/kvcontroller.rs
+++ b/crates/kv/src/kvcontroller.rs
@@ -1,15 +1,22 @@
+use std::{
+    sync::Arc,
+    time::{Duration, Instant},
+};
+
 use coyote_error::{Error, Result};
 use coyote_namespace::entities::NamespaceId;
 use fjall::{KeyspaceCreateOptions, KvSeparationOptions};
 use fjall_utils::{StorageType, TableRow, WriteBatchExt};
 use itertools::Itertools;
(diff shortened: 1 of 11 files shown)


crates/kv/src/kvcontroller.rs
```diff
@@ -1,15 +1,22 @@
...
+    time::{Duration, Instant},
+};
+
 use coyote_error::{Error, Result};
 use coyote_namespace::entities::NamespaceId;
 use fjall::{KeyspaceCreateOptions, KvSeparationOptions};
 use fjall_utils::{StorageType, TableRow, WriteBatchExt};
 use itertools::Itertools;
 use jiff::Timestamp;
+use parking_lot::{Mutex, MutexGuard};
```
svix-jplatte, line 12: Why parking_lot instead of std's Mutex?

svix-jbrown: I'm not actually using it in this PR, but I do find the `lock_arc` functionality to atomically lock and clone the arc helpful.

svix-jplatte (approved):

PR #525, opened 2026-03-19
Add strongly-typed IDs

svix-jplatte:
I wanted to put them in `coyote-core`, but that created a circular dependency with `coyote-namespace`.

Ref. https://svixhq.slack.com/archives/C0A72988962/p1773865378309009

Internal-only IDs are serde-compatible directly (compact format) and declared like:

```rust
id_marker!(Foo);
type FooId = Id<Foo>;
```

Public IDs are declared like:

```rust
id_marker!(Foo, "foo_");
type FooId = Id<Foo>;
```

and serialized / deserialized as such:

```rust
// part of public API
struct Thing {
    id: Public<FooId>, // obtained from foo_id.public()
    // ...
}

// serialized to fjall or whatever
struct OtherThing {
    id: Internal<FooId>, // obtained from foo_id.internal()
    // ...
}
```


tasn:
TBH we should probably move namespace to core or something anyway. It being in a separate crate has been a pain in the ass.

crates/coyote-id/src/lib.rs
```diff
@@ -0,0 +1,158 @@
...
+pub use self::{internal::Internal, public::Public};
+
+pub type NamespaceId = Id<m::Namespace>;
+pub type TopicId = Id<m::Topic>;
+
+// Don't want these types to be nameable,
+// so they're defined in a private submodule
+mod m {
+    // Markers for internal-only IDs
+    id_marker!(Namespace);
```
tasn, line 22: What's this? We don't have it in svix-webhooks.

crates/coyote-id/src/lib.rs
```diff
@@ -0,0 +1,158 @@
+use std::{fmt, marker::PhantomData};
```
tasn, line 1: Can you add somewhere (e.g. even the PR description/comments on the PR) high level example/instructions on how to use this correctly while ensuring it serializes correctly in the public API, but also not a problem externally?

tasn, line 1: Just to make sure I understand:
1. Internal: serializes to bytes, efficiently, meant for IDs we don't share with customers.
2. public: serializes to string, meant for IDs we use public facing?

crates/coyote-id/src/lib.rs
tasn: Ah, that's how you mark internal vs. external? Why not make it explicit above?
```
pub type NamespaceId = PublicId<m::Namespace>;
pub type TopicId = InternalId<m::Topic>;
```

crates/coyote-id/src/lib.rs
```diff
@@ -0,0 +1,158 @@
...
+
+use self::marker::{IdMarker, InternalUse, PublicIdMarker};
+pub use self::{internal::Internal, public::Public};
+
+pub type NamespaceId = Id<m::Namespace>;
+pub type TopicId = Id<m::Topic>;
+
+// Don't want these types to be nameable,
+// so they're defined in a private submodule
+mod m {
```
tasn, line 20: What happens if we want to turn an ID external at some point? E.g. we used it internally and now want to expose it in an API?

crates/coyote-id/src/lib.rs
svix-jplatte: There are tests at the end of this file. I can update the PR description though.

svix-jplatte: I went a bit more sophisticated:

- Internal IDs implement `Serialize`, `Deserialize` directly and always uses the compact format
- Public IDs don't do so, you have to call `.public()` / `.private()` (to obtain `Public<FooId>` / `Internal<FooId>`) to decide the format to use within a certain context

crates/coyote-id/src/lib.rs
svix-jplatte: Have to update existing uses in serde structs from `FooId` to `Public<FooId>` and add conversion method calls.

crates/coyote-id/src/lib.rs
svix-jplatte: That would mean duplicating most / all the `Id` methods, but would be a little less complex implementation-wise. I don't have a strong opinion.

crates/coyote-id/src/lib.rs
tasn: Tests don't explain how to use something, they just show, but don't explain the why. And anyway, not reasonable to expect someone looking here 5 months from now to find this PR and read the tests.

tasn: I'll wait for the usage instructions, but cool, if the ID itself is not serialized by default it sounds like it would be hard to get it wrong. Let's see!

Though just a thought:
```
type NamespaceId = Id<Namespace>;
// And then to use it in a public context:
Public<NamespaceId>
```
This way the internals can just do it easily, and the public variant would be very explicit?

crates/coyote-id/src/lib.rs
tasn: I made an alternative suggestion above, though curious to first see how you envision using it.

crates/coyote-id/src/lib.rs
svix-jplatte: Is the new PR description clear enough?

tasn: Yes, thanks a lot! One thing that's missing is how to convert Public <-> Internal.

tasn: To build on this (now that I saw the examples):
```
// part of public API
struct Thing {
    id: Public<FooId>, // obtained from foo_id.public()
    // ...
}

// serialized to fjall or whatever
struct OtherThing {
    id: FooId,
    // ...
}
```

I know it makes type safety slightly weaker, but tbh, adding `.into()` in the code anyway is not huge friction, so I don't think it detracts that much type safety. As in, even without it, converting from public <> internal is easy enough to make a mistake.

What would be great if we could make it so people can't use these IDs in public-facing without Public<> but that can be a future thing.

crates/coyote-id/src/public.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+impl<M: PublicIdMarker> Serialize for Public<Id<M>> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        // FIXME: We may want to switch to a more compact form
+        //        (make sure to also update Deserialize, of course)
+        format!("{}{}", M::PREFIX, self.0.inner.as_simple()).serialize(serializer)
```
tasn: Let's change to base62.

tasn: Discussed on slack: Crawford base32, lowercase.

crates/coyote-id/src/lib.rs
svix-jplatte: I don't really understand what you're saying. There is no `.into()`, only `.public()` and `.internal()` (and `.into_inner()` to unwrap these wrapper types).

tasn (approved):

crates/coyote-id/src/lib.rs
tasn: In the API I'm going to use Public<FooId>, right? Now, how do I turn it into an internal? `foo.internal()`?

Though anyway, what I was saying that it feels a bit tedious to call it `Internal<FooId>` everywhere, and it's easier if the internal version would just be `FooId` and the external version would be `Public<FooId>`.

svix-jplatte: When passing it around, I'd just use plain `FooId`, and wrap it only for serde structs.

PR #528, opened 2026-03-19
server: expose Monotime to tests so we can test expiration without sleeps

### LINKED ISSUE #523: expose monotonic time to integration test context struct and allow fast-forwarding/rewinding it

svix-jbrown:
This does a few things:

 1. It extends the `Monotime` class to have a `fast_forward` method which we can use to mock time. This is behind a default-enabled feature, and we disable that feature when building the docker container. This kind of sucks, but I couldn't find a way to make cargo integration tests set a custom feature.
 2. It modifies `kv` to use Monotime when constructing timestamps (as #522)
 3. It modifies kv tests to time-travel instead of sleeping
 4. It initializes tracing for integration tests, which wasn't happening at all before.

Fixes #523.

### DIFF
--- a/crates/coyote-core/Cargo.toml
+++ b/crates/coyote-core/Cargo.toml
@@ -6,9 +6,6 @@ license.workspace = true
 rust-version.workspace = true
 edition = "2024"
 
-[features]
-mockable-time = []
-
 [dependencies]
 coyote-namespace.workspace = true
 coyote-proto.workspace = true
--- a/crates/coyote-core/Cargo.toml
+++ b/crates/coyote-core/Cargo.toml
@@ -6,6 +6,9 @@ license.workspace = true
 rust-version.workspace = true
 edition = "2024"
 
+[features]
+mockable-time = []
+
 [dependencies]
 coyote-namespace.workspace = true
 coyote-proto.workspace = true
--- a/crates/coyote-core/src/monotime.rs
+++ b/crates/coyote-core/src/monotime.rs
@@ -10,11 +10,15 @@ use std::sync::{
 /// This structure is cheaply cloneable and clones all point at the same underlying data
 /// (internally, this is an Arc).
 #[derive(Debug, Clone)]
-pub struct Monotime(Arc<AtomicI64>);
+pub struct Monotime {
+    time: Arc<AtomicI64>,
+}
 
 impl Monotime {
     pub fn initial() -> Self {
-        Self(Arc::new(AtomicI64::new(0)))
+        Self {
+            time: Arc::new(AtomicI64::new(0)),
+        }
     }
 
     /// Get the last time that the leader set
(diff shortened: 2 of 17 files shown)


crates/coyote-core/Cargo.toml
```diff
@@ -6,6 +6,9 @@ license.workspace = true
 rust-version.workspace = true
 edition = "2024"
 
+[features]
+mockable-time = []
```
svix-james: I would feel better if this feature were explicitly enabled only for tests, if possible.

svix-jbrown: Unfortunately, we use it in the integration tests, and as far as I can tell, there's intentionally no way to only enable certain features during integration tests.

I guess I could change everything to run the integration tests with `cargo test --features mockable-time` and they'd just not work otherwise?

svix-jplatte: What you could do is have the production build set a compile cfg that's not a feature cfg by way of env var shenanigans (build.rs lets you set cfgs for a specific crate so you can do that based on any old env var, just make sure to have it print the cargo hint that it needs rerunning when that env var changes).

svix-jplatte: Which would be better because features get unified and you can't ever really reliably know you've disabled one. Unless it's directly on a leaf crate you're compiling.

svix-jbrown: That sounds like a possible future improvement, but here's what I'm going to do for now: put it behind `debug_assertions` so that it can't end up in a release build.

crates/coyote-core/src/monotime.rs
```diff
@@ -41,10 +45,68 @@ impl Monotime {
...
+        assert_eq!(mt.last(), Timestamp::UNIX_EPOCH);
+        // calling .now returns the current time and bumps `last`
+        let now = Timestamp::now();
+        assert_approximately_equal(now, mt.now());
+        assert_approximately_equal(now, mt.last());
+        let future = now + Duration::from_mins(10);
+        // bumping into the future prevents now from rewinding
+        mt.bump(future);
+        assert_approximately_equal(mt.last(), future);
+        assert_approximately_equal(mt.now(), future);
```
svix-jplatte, line 95: Could use exact equals here, right?

src/lib.rs
```diff
@@ -308,9 +318,15 @@ pub async fn run_with_listeners(
     tracing::debug!("we're outta here!");
 }
 
+static TEST_TRACING_INITIALIZED: AtomicBool = AtomicBool::new(false);
```
svix-jplatte, line 321: Could use `std::sync::Once` for this.

svix-jplatte (approved):

crates/coyote-core/src/monotime.rs
svix-jbrown: monotime is truncated to millisecond (so it fits in a u64), but `jiff::Timestamp` has microsecond, so I'd need to do something like

```
assert_eq!(mt.last(), future.round(Unit::Millisecond)
```

meh

PR #553, opened 2026-03-20
Auth module: create an auth module for Coyote users.

### LINKED ISSUE #428: Add auth tokens / authentication middleware
See https://github.com/svix/rfc/blob/main/prds/coyote/auth.md#auth-tokens.

This is separate from ACLs (#68).

tasn:
This is for coyote users to use, so similar to kv, cache, etc.

This module is *NOT* ready for usage by Coyote customers, but I want it in as I want to use it internally in a subsequent PR.

### DIFF
--- a/crates/auth-token/src/entities.rs
+++ b/crates/auth-token/src/entities.rs
@@ -1,0 +1,76 @@
+use std::fmt;
+
+use base64::{Engine, engine::general_purpose::URL_SAFE_NO_PAD};
+use coyote_error::Result;
+use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
+
+const KEY_SIZE: usize = 24;
+
+#[derive(Clone)]
+pub struct TokenPlaintext(String);
+
+impl TokenPlaintext {
+    pub fn generate(prefix: &str, suffix: Option<&str>) -> Result<Self> {
+        let mut buf = [0u8; KEY_SIZE];
+        rand::RngCore::fill_bytes(&mut rand::rng(), &mut buf);
+
+        let generated = URL_SAFE_NO_PAD.encode(buf);
+
+        let suffix = suffix.unwrap_or("api");
+        let plaintext = format!("{prefix}_{generated}.{suffix}");
+        Ok(Self::from_plaintext_dangerous(plaintext))
+    }
+
+    pub fn from_plaintext_dangerous(pt: String) -> Self {
+        Self(pt)
+    }
+
+    pub fn hash(&self) -> TokenHashed {
+        self.0.as_str().into()
+    }
+
+    /// Exposes the plaintext. Ensure this value is only rendered to users when a token is created. This value should never be logged.
+    pub fn expose_plaintext_dangerously(self) -> String {
+        self.0
+    }
+}
+
+/// A hashed auth token. The inner bytes are never printed to prevent accidental exposure.
+#[derive(Clone, Serialize, Deserialize)]
+pub struct TokenHashed([u8; 32]);
+
+impl TokenHashed {
+    pub fn inner(&self) -> &[u8; 32] {
+        &self.0
+    }
+}
+
+impl From<[u8; 32]> for TokenHashed {
+    fn from(value: [u8; 32]) -> Self {
+        Self(value)
+    }
+}
+
+impl From<&str> for TokenHashed {
+    fn from(value: &str) -> Self {
+        Self(Sha256::digest(value.as_bytes()).into())
+    }
+}
+
+impl fmt::Debug for TokenHashed {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "TokenHashed({})", URL_SAFE_NO_PAD.encode(self.0))
+    }
+}
+
+impl TryFrom<&[u8]> for TokenHashed {
+    type Error = coyote_error::Error;
+
+    fn try_from(bytes: &[u8]) -> Result<Self> {
+        let arr: [u8; 32] = bytes.try_into().map_err(|_| {
+            coyote_error::Error::internal(format!("expected 32 bytes, got {}", bytes.len()))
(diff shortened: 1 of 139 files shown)


crates/auth-token/src/operations/create.rs
```diff
@@ -0,0 +1,87 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct CreateResponseData {
+    pub model: AuthTokenModel,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct CreateAuthTokenOperation {
+    namespace_id: NamespaceId,
+    storage_type: StorageType,
```
svix-james, line 23: Why is `storage_type` here? It doesn't appear to be used for anything.

crates/auth-token/src/operations/delete.rs
```diff
@@ -0,0 +1,42 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct DeleteResponseData {
+    pub success: bool,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct DeleteAuthTokenOperation {
+    namespace_id: NamespaceId,
+    storage_type: StorageType,
```
svix-james, line 19: Same here.

crates/auth-token/src/operations/expire.rs
```diff
@@ -0,0 +1,51 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ExpireResponseData {
+    /// `None` if the token was not found.
+    pub model: Option<AuthTokenModel>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ExpireAuthTokenOperation {
+    namespace_id: NamespaceId,
+    storage_type: StorageType,
```
svix-james, line 22: And here.... I assume there's a reason for this, but it's lost on me.

tasn: It's because I haven't implemented storage type support yes, as we only use it internally and we use persistent there.

crates/auth-token/src/operations/create_namespace.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+impl AuthTokenRequest for CreateAuthTokenNamespaceOperation {
+    async fn apply(
+        self,
+        state: AuthTokenRaftState<'_>,
+        ctx: &coyote_operations::OpContext,
+    ) -> CreateNamespaceResponse {
+        CreateNamespaceResponse(self.apply_real(state.namespace, ctx.timestamp))
```
svix-jbrown:
```suggestion
    async fn apply_real(
        self,
        namespace_state: &coyote_namespace::State,
        now: Timestamp,
    ) -> coyote_operations::Result<CreateAuthTokenNamespaceResponseData> {
        let op: CreateNamespace<AuthTokenConfig> = self.into();
        let out = op.async_apply_operation(namespace_state, now).await?;
        Ok(out.into())
    }
}

#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct CreateAuthTokenNamespaceResponseData {
    pub name: String,
    pub max_storage_bytes: Option<NonZeroU64>,
    pub storage_type: StorageType,
    pub created: Timestamp,
    pub updated: Timestamp,
}

impl From<CreateNamespaceOutput<AuthTokenConfig>> for CreateAuthTokenNamespaceResponseData {
    fn from(value: CreateNamespaceOutput<AuthTokenConfig>) -> Self {
        Self {
            name: value.name,
            max_storage_bytes: value.max_storage_bytes,
            storage_type: value.storage_type,
            created: value.created_at,
            updated: value.updated_at,
        }
    }
}

impl AuthTokenRequest for CreateAuthTokenNamespaceOperation {
    async fn apply(
        self,
        state: AuthTokenRaftState<'_>,
        ctx: &coyote_operations::OpContext,
    ) -> CreateNamespaceResponse {
        CreateNamespaceResponse(self.apply_real(state.namespace, ctx.timestamp).await)
```

this noise will go away Soon™

crates/auth-token/src/tables.rs
```diff
@@ -0,0 +1,173 @@
...
+/// A complete auth token — primary row plus all index keys bundled together.
+/// Use `upsert` or `remove` on a batch; never write individual rows directly.
+pub(crate) struct AuthTokenEntity {
+    pub namespace_id: NamespaceId,
+    pub token_hashed: TokenHashed,
+    pub row: AuthTokenRow,
+}
+
+impl AuthTokenEntity {
+    pub(crate) fn upsert(
```
svix-jbrown, line 132: do we want to document somewhere that it's illegal to change any of the fields that are in the indices (since we just insert and don't remove)?

is there a way we can encode that invariant into the type system?

crates/auth-token/src/entities.rs
```diff
@@ -0,0 +1,76 @@
...
+
+const KEY_SIZE: usize = 24;
+
+#[derive(Clone)]
+pub struct TokenPlaintext(String);
+
+impl TokenPlaintext {
+    pub fn generate(prefix: &str, suffix: Option<&str>) -> Result<Self> {
+        let mut buf = [0u8; KEY_SIZE];
+        getrandom::fill(&mut buf).or_internal_error()?;
```
svix-jbrown: we should probably just use `rand` here instead of dragging in a new dep

crates/auth-token/src/tables.rs
tasn: I don't know. it was just an idea to make it less likely to mess up forgetting updating the indices (everywhere else we just willy nilly update them separately).

crates/auth-token/src/entities.rs
tasn: Done.

svix-jbrown (approved):
Image

PR #561, opened 2026-03-20
Rate limit: add expiry to rate-limit keys.

tasn:
Keys expire when they would have been completely full if we kept accounting of that. So essentially we calculate when they would be full, and set the expiry to that.


tasn:
TBH, maintaining an index that's updated in the hot-path is not optimal, we should probably just sample the keys (or walk through all the keys) and do it that way, so we don't affect the write hot path.

PR #562, opened 2026-03-20
Use cli-like syntax for bootstrapping

svix-james:
No more yaml. This parses a new-line-delimited string
for commands to create the namespaces. The API structs are
used directly here, so there's not much intermediate typing
to bloat things.

Example:
```sh
kv namespace create {"name":"default","storage_type":"Ephemeral","max_storage_bytes":1000}
```


src/bootstrap.rs
```diff
@@ -1,151 +1,215 @@
...
-        // Configure default namespace for stream (msgs), if not part of the config file
-        if let Entry::Vacant(v) = config
-            .stream
-            .get_or_insert_default()
-            .entry("default".to_owned())
-        {
-            v.insert(NamespaceIn::default());
-        }
+    ensure_default!(
+        Stream,
```
tasn: The module is called msgs, can you please correct it?

tests/it/static/bootstrap.test
```diff
@@ -0,0 +1,5 @@
+kv namespace create {"name":"default","storage_type":"Ephemeral","max_storage_bytes":1000}
```
tasn, line 1: Can you copy this line to the PR description to show how people should use it?

tasn (approved):

PR #564, opened 2026-03-20
Auth token admin: implement admin functions to manage auth tokens

### LINKED ISSUE #428: Add auth tokens / authentication middleware

tasn:
These are auth tokens to be used by Coyote, so users of Coyote create them and use them in order to access Coyote.

It uses our own auth token module internally as a user of Coyote.

There are a few action items missing:
1. Mark the internal namespace as internal so that it's not accessible to users of coyote. Maybe we should just enforce it with ACL instead?
2. The API is still work-in-progress, need to review and make sure we like it.


tasn:
Need to update based on https://github.com/svix/coyote-private/pull/591

src/v1/endpoints/admin/auth_token.rs
```diff
@@ -0,0 +1,326 @@
...
+    let mut metadata = HashMap::new();
+    metadata.insert("role".to_string(), data.role);
+
+    let req = AuthTokenCreateIn::new(data.name, TOKEN_OWNER.to_string())
+        .with_namespace(INTERNAL_NAMESPACE.to_string())
+        .with_expiry_millis(data.expiry_millis)
+        .with_metadata(Some(metadata))
+        .with_enabled(data.enabled);
+    let out = app_state
+        .coyote_client
```
tasn: Use jonas's thing instead of coyote_client.

src/v1/endpoints/admin/auth_token.rs
```diff
@@ -0,0 +1,326 @@
...
+        .auth_token()
+        .create(req)
+        .await
+        .map_err(|e| crate::error::Error::internal(e.to_string()))?;
```
tasn: Better error handling?

crates/coyote-core/src/fifo_cache.rs
```diff
@@ -0,0 +1,29 @@
...
+        self.map
+            .get(key)
+            .filter(|(inserted_at, _)| inserted_at.elapsed() < ttl)
+            .map(|(_, v)| v)
+    }
+
+    pub fn put(&mut self, key: String, value: V) {
+        self.map.insert(key, (Instant::now(), value));
+        if self.map.len() > self.capacity {
+            self.map.shift_remove_index(0);
```
svix-jplatte: This has to move every other element in the map in memory, so every call is going to do a big memmove once capacity has been reached.

src/core/auth.rs
```diff
@@ -84,15 +94,49 @@ async fn authorization_inner(
 
     let token = bearer.token();
 
-    let perms = if let Some(admin_token) = state.cfg.admin_token.as_ref()
+    if let Some(admin_token) = state.cfg.admin_token.as_ref()
```
svix-jplatte:
```suggestion
    if let Some(admin_token) = &state.cfg.admin_token
```

(non-blocking)

src/v1/endpoints/admin/auth_token.rs
```diff
@@ -0,0 +1,363 @@
+// SPDX-FileCopyrightText: © 2022 Svix Authors
```
svix-jplatte, line 1: It's not 2022 by the way 😛

src/v1/endpoints/admin/auth_token.rs
```diff
@@ -0,0 +1,363 @@
...
+    let data = out
+        .data
+        .into_iter()
+        .map(|t| AdminAuthTokenOut {
+            id: t.id,
+            name: t.name,
+            created: t.created,
+            updated: t.updated,
+            expiry: t.expiry,
+            role: t.metadata.get("role").unwrap().to_string(),
```
svix-jplatte: I saw an unwrap like this fail when running tests on your branch previously. I know tests succeed now, but this should probably be `Error::internal` rather than a panic, right?

src/v1/endpoints/admin/auth_token.rs
tasn: Copy pasta

src/v1/endpoints/admin/auth_token.rs
tasn: We need to figure out how we want to do metadata in the API in a way that I can just use serde for this.

This should never happen (famous last words), though yeah, I'll change it.

crates/coyote-core/src/fifo_cache.rs
tasn: Is that how index map works? I'll look again, I assumed it was just a pointer move.

src/v1/endpoints/admin/auth_token.rs
tasn: Done.

crates/coyote-core/src/fifo_cache.rs
```diff
@@ -0,0 +1,103 @@
...
+            while let Some((oldest_key, oldest_time)) = self.order.pop_front() {
+                match self.map.get(&oldest_key) {
+                    Some((map_time, _)) if *map_time == oldest_time => {
+                        self.map.remove(&oldest_key);
+                        break;
+                    }
+                    _ => {}
+                }
+            }
+        }
```
svix-jplatte, line 51: What's this supposed to do? I would have expected a single `pop_front`, then remove the associated map entry, nothing more.

svix-jplatte: Unclear why the `Vec` contains the time at all.

svix-james (approved):

crates/coyote-core/src/fifo_cache.rs
tasn: Consider this:
1. Write to cache.
2. Cache expires (so get returns None)
3. We write the same key to cache.

At this point the vector will have two entries for the same key, so this just makes sure we don't wipe the cache if it was updated recently.

PR #570, opened 2026-03-20
set up cargo-deny

svix-jbrown:
sets up `cargo-deny` and fixes the things it complains about (namely, removing my now-unneeded patch for dotenvy, and replacing paste with pastey).

branched off of #569

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,3 +126,10 @@ jobs:
           else
             echo 'all good.'
           fi
+  security-audit:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v6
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -126,3 +126,10 @@ jobs:
           else
             echo 'all good.'
           fi
+  security-audit:
```
tasn, line 129: Make it so PR reviews are blocked on this.

tasn (approved):

.github/workflows/ci.yml
svix-jbrown: appears that everything in ci.yml is included by default

svix-jplatte (approved):

PR #580, opened 2026-03-21
Add ability to configure stream consumer group to start at beginning of the stream

svix-gabriel:
Stream consumer groups, by default, start at the end of the stream when they're first created. Similar to Kafka's default.

Kafka has the ability to tell a consumer group to start at the beginning (if the consumer group hasn't already committed offsets) - which is a feature we USE in svix-webhooks-private. (Specifically, when a server may write messages to a topic before workers with relevant consumer groups have been spun up).

To make this work, I added a `msgs.stream.configure` endpoint, analogous to the already extant `msgs.queue.configure` for configuring queue consumer groups. This lets the client configure the default starting position for the consumer group.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -3353,6 +3353,10 @@
             "type": "integer",
             "format": "uint64",
             "default": 300000
+          },
+          "default_starting_position": {
+            "type": "string",
+            "default": "latest"
           }
         },
         "required": [
--- a/src/v1/endpoints/msgs.rs
+++ b/src/v1/endpoints/msgs.rs
@@ -188,6 +188,8 @@ struct MsgStreamReceiveIn {
     pub batch_size: NonZeroU16,
     #[serde(default = "default_lease_duration_millis")]
     pub lease_duration_millis: DurationMs,
+    #[serde(default)]
+    pub default_starting_position: SeekPosition,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Validate, JsonSchema)]
@@ -216,6 +218,7 @@ async fn stream_receive(
         data.consumer_group,
         data.batch_size,
         data.lease_duration_millis,
+        data.default_starting_position,
     )?;
     let response = repl.client_write(operation).await.or_internal_error()?.0?;
 
--- a/clients/go/internal/apis/msgs_stream.go
+++ b/clients/go/internal/apis/msgs_stream.go
@@ -28,11 +28,12 @@ func (msgsStream MsgsStream) Receive(
 	msgStreamReceiveIn coyote_models.MsgStreamReceiveIn,
 ) (*coyote_models.MsgStreamReceiveOut, error) {
 	body := coyote_models.MsgStreamReceiveIn_{
-		Namespace:           msgStreamReceiveIn.Namespace,
-		Topic:               topic,
-		ConsumerGroup:       consumerGroup,
-		BatchSize:           msgStreamReceiveIn.BatchSize,
-		LeaseDurationMillis: msgStreamReceiveIn.LeaseDurationMillis,
+		Namespace:               msgStreamReceiveIn.Namespace,
+		Topic:                   topic,
+		ConsumerGroup:           consumerGroup,
+		BatchSize:               msgStreamReceiveIn.BatchSize,
+		LeaseDurationMillis:     msgStreamReceiveIn.LeaseDurationMillis,
+		DefaultStartingPosition: msgStreamReceiveIn.DefaultStartingPosition,
 	}
 
 	return coyote_proto.ExecuteRequest[coyote_models.MsgStreamReceiveIn_, coyote_models.MsgStreamReceiveOut](
--- a/clients/go/internal/models/msg_stream_receive_in.go
+++ b/clients/go/internal/models/msg_stream_receive_in.go
@@ -3,15 +3,17 @@ package coyote_models
 // This file is @generated DO NOT EDIT
 
 type MsgStreamReceiveIn struct {
-	Namespace           *string `msgpack:"namespace,omitempty"`
-	BatchSize           *uint16 `msgpack:"batch_size,omitempty"`
-	LeaseDurationMillis *uint64 `msgpack:"lease_duration_millis,omitempty"`
+	Namespace               *string `msgpack:"namespace,omitempty"`
+	BatchSize               *uint16 `msgpack:"batch_size,omitempty"`
+	LeaseDurationMillis     *uint64 `msgpack:"lease_duration_millis,omitempty"`
+	DefaultStartingPosition *string `msgpack:"default_starting_position,omitempty"`
 }
 
 type MsgStreamReceiveIn_ struct {
-	Namespace           *string `msgpack:"namespace,omitempty"`
-	Topic               string  `msgpack:"topic"`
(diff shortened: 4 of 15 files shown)


tests/it/msgs/stream_configure.rs
```diff
@@ -0,0 +1,233 @@
...
+        .await?
+        .expect(StatusCode::OK);
+
+    // Configure consumer group with auto_offset_reset=earliest
+    let response = client
+        .post("msgs/stream/configure")
+        .json(json!({
+            "topic": "ns-scfg-earliest:t1",
+            "consumer_group": "cg1",
+            "auto_offset_reset": "earliest",
```
svix-gabriel: I'm not sold on `auto_offset_reset` as a param name here, so I'm fine w/changing this to something else

src/v1/endpoints/msgs.rs
```diff
@@ -273,6 +273,54 @@ async fn stream_commit(
...
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Validate, JsonSchema)]
+struct MsgStreamConfigureOut {
+    pub auto_offset_reset: SeekPosition,
+}
+
+/// Configures a stream consumer group on a topic.
+///
+/// `auto_offset_reset` controls where a new consumer group starts reading: `"earliest"` (offset 0)
+/// or `"latest"` (after the last existing message). This only affects the first `stream/receive`
+/// call for a consumer group that has no existing lease.
```
tasn: What does this mean that it only affects the first call for a consumer that doesn't have an existing lease?

src/v1/endpoints/msgs.rs
```diff
@@ -273,6 +273,54 @@ async fn stream_commit(
...
+// ---------------------------------------------------------------------------
+
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Validate, JsonSchema)]
+#[schemars(extend("x-positional" = ["topic", "consumer_group"]))]
+struct MsgStreamConfigureIn {
+    #[serde(default)]
+    pub namespace: Option<NamespaceName>,
+    pub topic: TopicName,
+    pub consumer_group: ConsumerGroup,
+    pub auto_offset_reset: SeekPosition,
```
tasn: We probably want a default for this one, so that it doesn't show as positional. What you did above only controls order I believe?

src/v1/endpoints/msgs.rs
tasn: OK, I just realized I thought it was seek.

src/v1/endpoints/msgs.rs
```diff
@@ -273,6 +273,54 @@ async fn stream_commit(
...
+struct MsgStreamConfigureOut {
+    pub auto_offset_reset: SeekPosition,
+}
+
+/// Configures a stream consumer group on a topic.
+///
+/// `auto_offset_reset` controls where a new consumer group starts reading: `"earliest"` (offset 0)
+/// or `"latest"` (after the last existing message). This only affects the first `stream/receive`
+/// call for a consumer group that has no existing lease.
+#[aide_annotate(op_id = "v1.msgs.stream.configure")]
```
tasn: As noted on slack: https://svixhq.slack.com/archives/C02QBTJSY3Y/p1774054836528489?thread_ts=1774053068.555019&cid=C02QBTJSY3Y

I don't think this is an API that we discussed or need. What do you envision it being used for? It looks like everything it may be able to do "seek" does, and I don't see what you'd want this for. Seek example: https://coyote.svix.io/modules/msgs/stream#seeking

Do we need the `auto_` aspect of it? Can you share more about the use-case, as I'm not sure why it would be useful (but I know kafka has it). We can defo just change "seek" to have a parameter indicating it only does something if there's nothing set yet than a separate configure call.

svix-gabriel: Seek is different.

Seek forces the position of the cg to a specific offset, even if it's read from the topic before.

`auto_offset_reset` forces the consumer group to start from a position, _but only if it hasn't read from the topic yet_.

This is useful, it's exactly the same feature kafka has https://www.confluent.io/learn/kafka-auto-reset/.

And we do use it on the svix-webhooks-private side. If you imagine a bootstrapped environment,

API servers -> write to topics

Consumers - read from topics

It's possible for the API servers to write to the topic _before_ the consumer groups have spun up, but we still need the consumer groups to start from the beginning of the stream when reading messages in this circumstance.

src/v1/endpoints/msgs.rs
svix-gabriel: fixed

src/v1/endpoints/msgs.rs
tasn: > Seek is different.
> 
> Seek forces the position of the cg to a specific offset, even if it's read from the topic before.
> 
> auto_offset_reset forces the consumer group to start from a position, but only if it hasn't read from the topic yet.

I know, I addressed that in my comment: "We can defo just change "seek" to have a parameter indicating it only does something if there's nothing set yet than a separate configure call."

> It's possible for the API servers to write to the topic before the consumer groups have spun up, but we still need the consumer groups to start from the beginning of the stream when reading messages in this circumstance.

Yes, that's how stream works in general, streams default to reading from the start.

> And we do use it on the svix-webhooks-private side.

Yes, you said that, but how and why?

## More complete thoughts

The way kafka does it is by setting it as a configuration on the consumer SDK which they can do because they are stateful. For us it would be akin to making it on the receive call (which I think we have discussed at a previous RFC at some point), as you said, you want to bootstrap, so it makes sense to bootstrap at receipt time without having to make a separate API call.
If we don't think it's important to do it without a separate API call, then I think `seek` with a parameter `only_if_new` (new name needed) is probably better than an explicit configure call on consumer groups. As this is inherently just a seek that you want to noop if not new. We can probably also have the API set a special flag on receive that returns an error if consumer group is new, which would then make the SDK (only on the first call) make another call to Coyote to seek to the wanted position (with `only_if_new` set), and then start receiving.

svix-gabriel: > Yes, you said that, but how and why?

I DID explain why.

We have API servers that write to a kafka topic. We have workers that read from a kafka topic. Now consider the time period when these instances first spin up.

The API servers may start writing data to the kafka topic _before_ the consumers in the workers can start reading from the topic. If Consumers always start from the end of the topic, they will miss data written to the topic before the consumers were initialized. This is why we need the ability to tell consumers to start at the beginning of a stream on initialization.

> We can defo just change "seek" to have a parameter indicating it only does something if there's nothing set yet than a separate configure call.

We can do this, but I don't like this design. `seek` has the implication (both from kafka, and currently) of "we are changing the position of the consumer group". That's not what `auto.offset.reset` is. `auto.offset.reset` is determining the start position of a consumer_group on initialization, it's literally part of the cg's configuration.

tasn (dismissed):

openapi.json
```diff
@@ -3353,6 +3353,10 @@
             "type": "integer",
             "format": "uint64",
             "default": 300000
+          },
+          "default_starting_position": {
+            "type": "string",
+            "default": "latest"
           }
```
svix-jplatte, line 3360: This is wrong, but I see there's a fixme comment on the manual `JsonSchema` impl. I'll look into it.

svix-jplatte (approved):

PR #581, opened 2026-03-21
Auth token: implement token rotation

tasn:


crates/auth-token/src/controller.rs
```diff
@@ -240,27 +278,69 @@ impl AuthTokenController {
...
+                &mut batch,
                 &keyspace,
-                AuthTokenRow::key_for(namespace_id, &id_index.token_hashed),
-            )?
+                namespace_id,
+                old_row.id,
+                input.old_expiry,
+                input.now,
+            )?;
+            new_entity.upsert(&mut batch, &keyspace)?;
```
svix-jbrown, line 322: I think this is going to result in the indices getting out of sync for some of the fields, so we probably need to be more intentional than just calling upsert. maybe I'm crazy

tasn: I have this bug somewhere else, but this is not a problem because all of these values are immutable. :)

(and expiry is not indexed, because we don't cleanup in the background)

svix-jbrown (approved):

PR #585, opened 2026-03-21
go: Switch from JSON to msgpack

### LINKED ISSUE #183: Go SDK: switch to msgpack

tasn:
This was done with the help of Claude.

Fixes #183

### DIFF
--- a/clients/go/internal/proto/nullable.go
+++ b/clients/go/internal/proto/nullable.go
@@ -1,6 +1,6 @@
 package coyote_proto
 
-import "encoding/json"
+import msgpack "github.com/vmihailenco/msgpack/v5"
 
 type Nullable[T any] struct {
 	val   *T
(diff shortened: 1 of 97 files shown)


clients/go/internal/proto/nullable.go
```diff
@@ -42,25 +42,28 @@ func (v *Nullable[T]) Unset() {
 	v.isSet = false
 }
 
-func (n *Nullable[T]) UnmarshalJSON(data []byte) error {
+func (n Nullable[T]) IsZero() bool {
```
svix-jbrown, line 45: why is this passed by value instead of ptr?

clients/go/internal/proto/nullable.go
```diff
@@ -42,25 +42,28 @@ func (v *Nullable[T]) Unset() {
...
+	return !n.isSet
+}
+
+func (n *Nullable[T]) UnmarshalMsgpack(data []byte) error {
 	n.isSet = true
-	if string(data) == "null" {
+	if len(data) == 0 {
```
svix-jbrown, line 51: is there a difference in this encoding between null and the empty string?

svix-jbrown (commented):
this feels weird enough that it'd be worth seeing some tests that it actually works

svix-jplatte (approved):
Tested this and it works :+1:

The reason I failed before was I failed to update the `UnmarshalJSON` specializations, I think.

svix-jplatte:
@svix-jbrown I will take care of your concerns about `nullable.go` - by deleting it :smile: 

I thought I had done that already, but just confirmed locally that it's unused.

PR #591, opened 2026-03-23
Add an internal API server for loopback calls

svix-jplatte:
… so we don't have to go through the system's networking stack for making self-calls (and can special-case self-call authorization more easily).

Part of https://github.com/svix/coyote-private/issues/68 (ish).


crates/coyote-proto/src/internal_client.rs
```diff
@@ -0,0 +1,104 @@
...
+    pub fn error_for_status(self) -> anyhow::Result<Self> {
+        let status = self.status();
+        if !status.is_success() {
+            // FIXME: Also get the error code / message from body
+            anyhow::bail!("received error response with status code {status}");
+        }
+        Ok(self)
+    }
+
+    pub async fn body<T: DeserializeOwned>(self) -> anyhow::Result<T> {
```
tasn: Can you make it more ergonomic? E.g. similar to using the libs? At the moment using it is like this, and it really should be like the libs...
Using it:
```
let ret = internal.post("/api/v1/kv/set", KvSetIn(...)).await?;
let ret = ret.error_for_status()?;
let body: KvSetOut = ret.body()?
```

Can we change it to just be:
```
let body = internal.post("/api/v1/kv/set", KvSetIn(...)).await?;
```

Not the end of the world, but I assume it's not hard to make it more like the libs?

Just to be clear: I'm ok if everything just turns to an error log + "internal server error" to customers. Though I should be able to have error types I can use to see what's up.

tasn (approved):

PR #592, opened 2026-03-23
Allow more characters in entity keys

svix-lucho:
Allow '=' and '+' so the base64 standard alphabet can be used in keys

### DIFF
--- a/crates/coyote-core/src/types/mod.rs
+++ b/crates/coyote-core/src/types/mod.rs
@@ -184,7 +184,7 @@ string_wrapper!(
     StringSchema {
         string_validation: Some(json_schema!({
             "maxLength": 256,
-            "pattern": r"^(?:[a-zA-Z0-9\-_.]+:)?[a-zA-Z0-9\-/_.]+$",
+            "pattern": r"^(?:[a-zA-Z0-9\-_.]+:)?[a-zA-Z0-9\-/_.=+]+$",
         })),
         example: Some("some_key".to_string()),
     }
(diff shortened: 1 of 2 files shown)


crates/coyote-core/src/types/mod.rs
```diff
@@ -191,7 +184,7 @@ string_wrapper!(
     StringSchema {
         string_validation: Some(json_schema!({
             "maxLength": 256,
-            "pattern": r"^(?:[a-zA-Z0-9\-_.]+:)?[a-zA-Z0-9\-/_.]+$",
+            "pattern": r"^(?:[a-zA-Z0-9\-_.]+:)?[a-zA-Z0-9\-/_.=+]+$",
```
svix-james: Might not hurt to have at least one test somewhere for this as well.

svix-james (dismissed):

crates/coyote-core/src/types/mod.rs
tasn: We no longer accept namespaces, so you want to remove the first part, and add `:` to the alaphabet.

```suggestion
            "pattern": r"^[a-zA-Z0-9\-/_.=+:]+$",
```

svix-lucho: right. updated

svix-lucho: updated the test as well

crates/coyote-core/src/types/mod.rs
```diff
@@ -287,15 +280,15 @@ mod tests {
         allowed("foo/bar/baz");
         allowed("foo:bar");
         allowed("foo:bar/baz");
+        allowed("1");
+        allowed("foo/+==");
     }
```
svix-james, line 290: So should these be acceptable now? If so, should we add them to allowed assertions?

svix-lucho: yeah. That's what I did here, no? not sure what you mean

svix-james, line 290: Sorry, I mean these assertions. Should we move them to `accepted` instead of just removing them since Tom has requested you put `:` in allowed characters now?

svix-lucho: oh got it got it. Updated

svix-james (approved):

PR #596, opened 2026-03-23
Allow deploying coyote onto specific nodes

svix-james:
Allow specifying node selection and taint toleration so we can control which nodes the operator/cluster gets scheduled on.

We are not quite ready to actually use these yet, but let's get support for it ready.

### DIFF
--- a/infra/helm-coyote/values.yaml
+++ b/infra/helm-coyote/values.yaml
@@ -69,3 +69,10 @@ cluster:
         value: grpc://datadog-agent.datadog.svc.cluster.local:4317
       - name: COYOTE_ADMIN_TOKEN
         value: "admin_abcdefghijlmnopqrstuvwxyz12345"
+    nodeSelector:
+      db-name: coyote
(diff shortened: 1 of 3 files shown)


infra/helm-coyote/values.yaml
```diff
@@ -69,3 +69,13 @@ cluster:
         value: grpc://datadog-agent.datadog.svc.cluster.local:4317
       - name: COYOTE_ADMIN_TOKEN
         value: "admin_abcdefghijlmnopqrstuvwxyz12345"
+    nodeSelector:
+      - key: db-name
+        operator: Equal
+        value: coyote
+        effect: NoSchedule
```
svix-jbrown: huh? this isn't the syntax for nodeSelector; nodeSelector looks like `{labelName: labelValue}`

svix-jbrown (changes requested):

svix-jbrown (approved):

PR #611, opened 2026-03-24
Rename clients directory to z-clients

svix-jplatte:
Yay for Go making our repo structure a public API concern (I'm very glad we're doing this before the first release).
I wonder if they support symlinks so we can make the import path `github.com/svix/coyote/go` instead of `github.com/svix/coyote/z-clients/go` :thinking:


svix-jbrown (approved):
no love for my Ω plan, huh?

PR #613, opened 2026-03-24
Make the URL be based on the op_id

### LINKED ISSUE #524: Change the RPC URLs to use the op_id?
We already have op_id, e.g. `v1.kv.set`, but the URLs for the operation is `/api/v1/kv/set`.
Should we just change it so that the URL is `/api/v1.kv.set`? Will save some code/redundancy/mistakes/decisions, but also feels cleaner?

tasn:
This now uses the op_id as the path. So `api/v1/kv/set` -> `api/v1.kv.set`.
I kept the openapi.json file under `api/v/1/openapi.json` though.

I also fixed a few wrong op-ids that had underscores in them, and blocked underscores from op-ids going forward.

I also fixed an issue with the python generation with nested resources.

Fixes #524

### DIFF
--- a/z-clients/javascript/src/apis/adminCluster.ts
+++ b/z-clients/javascript/src/apis/adminCluster.ts
@@ -1,6 +1,14 @@
 // this file is @generated
 
 import {
+    type ClusterRemoveNodeIn,
+    ClusterRemoveNodeInSerializer,
+} from '../models/clusterRemoveNodeIn';
+import {
+    type ClusterRemoveNodeOut,
+    ClusterRemoveNodeOutSerializer,
+} from '../models/clusterRemoveNodeOut';
+import {
     type ClusterStatusOut,
     ClusterStatusOutSerializer,
 } from '../models/clusterStatusOut';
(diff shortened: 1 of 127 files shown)


z-clients/javascript/src/apis/adminCluster.ts
```diff
@@ -19,6 +27,25 @@ export class AdminCluster {
...
             ClusterStatusOutSerializer._fromJsonObject,
         );
+    }/**
+* Remove a node from the cluster.
+* 
+* This operation executes immediately and the node must be wiped and reset
+* before it can safely be added to the cluster.
+*/
+    public removeNode(
+        clusterRemoveNodeIn: ClusterRemoveNodeIn,
```
svix-jbrown, line 37: at some point recently all the javascript whitespace got screwed up

svix-jbrown (commented):
I really don't like the idea of not being able to quickly search for what implementation corresponds to a given path; there's no way that if I jump into this code base a few months down the line I'm going to want to to dig through a proc macro to figure out that a request to `/foo/bar` corresponds to a call named `foo_bar_get_path` or whatever. Discoverability trumps ease of typing for me.

z-clients/javascript/src/apis/adminCluster.ts
svix-jplatte: No. It's been screwed up the whole time. I tried adding a biome formatting step (containerized like the rest) but somehow it didn't do anything. I will revisit that post MVP.

svix-jplatte (approved):

PR #617, opened 2026-03-25
More coyote-authorization base type stuff

svix-jplatte:
More sophisticated types (prep for evaluating requested operations against rules) and serde tests.

Part of https://github.com/svix/coyote-private/issues/68. Best reviewed commit by commit.

### DIFF
--- a/crates/coyote-authorization/src/pattern.rs
+++ b/crates/coyote-authorization/src/pattern.rs
@@ -1,0 +1,180 @@
+#![expect(unused_qualifications)] // triggered by schema_with
+
+use std::{fmt, str::FromStr};
+
+use itertools::Itertools;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize, de};
+
+#[derive(JsonSchema)]
+#[schemars(schema_with = "String::json_schema")]
+pub struct ResourcePattern {
+    module: Module,
+    namespace: NamespacePattern,
+    key: KeyPattern,
+}
+
+pub enum Module {
+    AuthToken,
+    Cache,
+    Idempotency,
+    Kv,
+    Msgs,
+    RateLimit,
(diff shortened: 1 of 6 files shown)


crates/coyote-authorization/src/pattern.rs
```diff
@@ -0,0 +1,183 @@
...
+#[derive(Debug, PartialEq, Eq, JsonSchema)]
+#[schemars(schema_with = "String::json_schema")]
+pub struct ResourcePattern {
+    pub module: Module,
+    pub namespace: NamespacePattern,
+    pub key: KeyPattern,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum Module {
```
svix-james, line 18: Do we not have something common that defines this? I feel like we may have already duplicated this in multiple places.

svix-james (approved):

crates/coyote-authorization/src/pattern.rs
svix-jplatte: Yeah I found another one, will dedup post merge.

crates/coyote-authorization/src/pattern.rs
```diff
@@ -0,0 +1,183 @@
...
+
+#[derive(Debug, PartialEq, Eq, JsonSchema)]
+#[schemars(schema_with = "String::json_schema")]
+pub struct ResourcePattern {
+    pub module: Module,
+    pub namespace: NamespacePattern,
+    pub key: KeyPattern,
+}
+
+#[derive(Debug, PartialEq, Eq)]
```
tasn, line 17: We can't have yet another copy of the module list. Enough is enough. :)

You can move one of the dozen we already have and move it somewhere central and reuse it.

PR #618, opened 2026-03-25
Add required access metadata to input structs

svix-jplatte:
Please ignore the action strings in `request_input!()` invocations for review, and let's have the module owners assign sensible ones post merge :)

Part of https://github.com/svix/coyote-private/issues/68.

### DIFF
--- a/src/v1/endpoints/cache.rs
+++ b/src/v1/endpoints/cache.rs
@@ -5,6 +5,7 @@ use std::num::NonZeroU64;
 
 use aide::axum::{ApiRouter, routing::post_with};
 use axum::{Extension, extract::State};
+use coyote_authorization::RequestedOperation;
 use coyote_cache::{
     CacheModel,
     operations::{CreateCacheOperation, DeleteOperation, SetOperation},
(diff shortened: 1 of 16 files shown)


svix-jplatte:
Doing some adjustments based on other merged PRs.

src/v1/endpoints/cache.rs
```diff
@@ -5,26 +5,51 @@ use std::num::NonZeroU64;
...
 
 use crate::{AppState, core::cluster::RaftState, error::Result, v1::utils::openapi_tag};
 
+fn cache_access_metadata<'a>(
+    ns: Option<&'a str>,
+    key: &'a EntityKey,
+    action: &'static str,
+) -> AccessMetadata<'a> {
+    AccessMetadata::RuleProtected(RequestedOperation {
+        module: Module::Cache,
```
svix-lucho, line 36: why not make the module name part of the macro? so you only have to define it once. It seems it would save a bunch of repeated code

svix-lucho (approved):

src/v1/endpoints/cache.rs
svix-jplatte: It's not just the module name, the key field is also named different things across modules (and for `auth_token`, doesn't exist).

PR #619, opened 2026-03-25
rust client: Remove non_exhaustive from enums

svix-lucho:
it's a bit annoying for users - see https://github.com/svix/svix-webhooks-private/pull/6128#discussion_r2983965857


svix-gabriel (approved):
waow

PR #621, opened 2026-03-25
Move namespaces to the request body in msgs topics

svix-lucho:
Follow up of https://github.com/svix/coyote-private/pull/530

I'd missed removing the namespace from the topic name, which made me realize I'd actually missed quite a lot of tests. The namespace in the topic wasn't actually being used (the API was updated), but they probably still worked because they worked with the default ns.

Anyway, this fixes the remaining bits.

### DIFF
--- a/crates/msgs/src/entities.rs
+++ b/crates/msgs/src/entities.rs
@@ -199,7 +199,7 @@ impl TopicIn {
     pub fn topic_name(&self) -> &TopicName {
         match self {
             TopicIn::TopicName(topic_name) => topic_name,
-            TopicIn::TopicPartition(topic_partition) => &topic_partition.raw,
+            TopicIn::TopicPartition(topic_partition) => &topic_partition.topic,
         }
     }
 }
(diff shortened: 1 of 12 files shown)


crates/msgs/src/entities.rs
```diff
@@ -143,15 +132,14 @@ impl TryFrom<String> for TopicPartition {
...
+        let (topic, idx_str) = value
             .rsplit_once(TOPIC_PARTITION_DELIMITER)
             .ok_or_else(|| Error::internal("missing '~' separator in topic"))?;
         let idx: u16 = idx_str
             .parse()
             .map_err(|_| Error::internal("invalid partition index in topic"))?;
         let partition = Partition::new(idx)?;
-        let raw = TopicName::new(ns.map(|x| x.to_owned()), topic.to_owned())?;
+        let raw = TopicName::new(topic.to_owned())?;
         Ok(Self { raw, partition })
```
svix-jplatte: Why is this field called `raw` by the way..?

svix-jplatte (dismissed):

crates/msgs/src/entities.rs
svix-lucho: Hmm probably made sense when `raw` was ns+topic but not nwo that it's just topic

svix-lucho: will fixx

svix-jplatte (approved):

PR #623, opened 2026-03-25
🔪 Claude

svix-gabriel:
And the shareholders wept


svix-jplatte (changes requested):
If we plan on continuing to use LLMs, I think we should not do this.

svix-gabriel:
Consensus seems to be that the vibecoding will continue until morale improves

PR #634, opened 2026-03-25
msgs: add scheduled message delivery via delay_ms

tasn:
Add support for publishing messages with a future delivery time. We support `delay_ms` (so a delay, rather than a specific time).

Queue consumers skip messages whose `scheduled_at` is in the future, so the message sits invisible in the log until its time arrives. Stream consumers see scheduled messages immediately (with `scheduled_at` set in the response), keeping the stream be based on the order of publishing.

No background processes or secondary storage are needed, and the implementation just relies on the existing lease visibility mechanism.


svix-gabriel (approved):
Using the synthetic lease is clever (in a good way)

PR #637, opened 2026-03-25
Helm: Add support for upgrading the CRD

svix-james:
This mimics the `kube-prometheus-stack` Helm chart in allowing the CRD to be updated by setting the Helm value `crd.upgrade.enabled` to true. If enabled, the CRD subchart will run a hook that updates (forcibly, if enabled) the CRD via a separate kubectl job.


svix-jbrown (approved):
this is gross but does appear to match how kube-prometheus-stack does it.

PR #638, opened 2026-03-25
Add messagepack support for JS client

svix-lucho:
Use https://www.npmjs.com/package/msgpackr to change the content type to message pack in the JS client. Nothing too fancy, just pack/unpack the requests and responses, with a configuration that works with the messagepack implementation in the Coyote server. 

Tested this locally with a bunch of endpoints and it seems to work.

### DIFF
--- a/z-clients/javascript/src/request.ts
+++ b/z-clients/javascript/src/request.ts
@@ -96,7 +96,7 @@ export class CoyoteRequest {
     private path: string
   ) { }
 
-  private body?: Uint8Array;
+  private body?: BodyInit;
   private queryParams: Record<string, string> = {};
   private headerParams: Record<string, string> = {};
 
@@ -146,7 +146,7 @@ export class CoyoteRequest {
 
   // biome-ignore lint/suspicious/noExplicitAny: intentional any
   public setBody(value: any) {
-    this.body = MSGPACK_CODEC.pack(value);
+    this.body = MSGPACK_CODEC.pack(value) as BodyInit;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


z-clients/javascript/tsconfig.json
```diff
@@ -3,6 +3,9 @@
     // Basic Options
     "strict": true,
     "target": "es2017",
+    "module": "ESNext",
+    "moduleResolution": "bundler",
```
svix-lucho, line 7: This is still wrong if we want to ship this lib for browsers **and** Node, but I'd rather have this than the hack this replaces.

z-clients/javascript/src/request.ts
```diff
@@ -78,9 +94,9 @@ export class CoyoteRequest {
   constructor(
     private readonly method: HttpMethod,
     private path: string
-  ) {}
+  ) { }
 
-  private body?: string;
+  private body?: BodyInit;
```
svix-jplatte, line 99: What is this type? Where is it coming from?

svix-jplatte (approved):

z-clients/javascript/tsconfig.json
svix-jplatte: Is there a ticket for this?

z-clients/javascript/src/request.ts
svix-lucho: This is a Node global that's used by `undici`

svix-lucho:
```
export type BodyInit =
  | ArrayBuffer
  | AsyncIterable<Uint8Array>
  | Blob
  | FormData
  | Iterable<Uint8Array>
  | NodeJS.ArrayBufferView
  | URLSearchParams
  | null
  | string
```

z-clients/javascript/tsconfig.json
svix-lucho: Created this https://github.com/svix/coyote-private/issues/652

Issue #609, opened 2026-03-24
tidy up authentication for coyote cluster layer


svix-jbrown:
I think every request on the interserver port except health should require the interserver shared-secret, and every human-issued request should use regular coyote auth.

svix-jbrown:
This is done

PR #639, opened 2026-03-25
cluster: move force-snapshot command to admin

svix-jbrown:
part of #609 

also adds an integration test

### DIFF
--- a/src/core/cluster/background.rs
+++ b/src/core/cluster/background.rs
@@ -357,6 +357,7 @@ pub(super) async fn run_background_jobs_on_all_nodes(
         if should_snapshot {
             last_snapshot_time = std::time::Instant::now();
             last_snapshot_index = committed;
+            #[allow(clippy::disallowed_methods)]
             let last_snapshot_timestamp = jiff::Timestamp::now();
             let payload = if trigger_snapshot(&handle, state, purge_by, committed).await?
                 && let Some(index) = last_snapshot_index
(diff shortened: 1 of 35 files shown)


src/core/cluster/background.rs
```diff
@@ -247,6 +248,55 @@ enum PurgeBy {
...
+    if state.is_learner() {
+        tracing::warn!("refusing to snapshot a learner");
+        return Ok(false);
+    } else {
+        tracing::debug!("triggering background snapshot");
+        if let Err(err) = handle.raft.trigger().snapshot().await {
+            tracing::error!(?err, "error triggering background snapshot; ignoring");
+            return Ok(false);
+        }
+    }
```
svix-jplatte:
```suggestion
    }

    tracing::debug!("triggering background snapshot");
    if let Err(err) = handle.raft.trigger().snapshot().await {
        tracing::error!(?err, "error triggering background snapshot; ignoring");
        return Ok(false);
    }
```

src/core/cluster/background.rs
```diff
@@ -247,6 +248,55 @@ enum PurgeBy {
...
+
+    let offset_to_purge = match purge_by {
+        PurgeBy::Time(duration) => {
+            #[allow(clippy::disallowed_methods)]
```
svix-jplatte, line 274: A comment about why it's fine to use real time here would be nice. Also wondering, will this be a problem for testing?

svix-jplatte (dismissed):

src/core/cluster/background.rs
svix-jbrown: commented

svix-jplatte (approved):

PR #649, opened 2026-03-26
Admin auth token: change owner to operator, not admin.

tasn:
This was just something we did because we didn't have operator. We changed it elsewhere, but not here.

### DIFF
--- a/src/v1/endpoints/admin/auth_token.rs
+++ b/src/v1/endpoints/admin/auth_token.rs
@@ -33,8 +33,6 @@ use crate::{
 
 use crate::core::INTERNAL_NAMESPACE;
 
-const TOKEN_OWNER: &str = "admin";
-
 #[derive(Clone, Debug, Serialize, Deserialize, JsonSchema)]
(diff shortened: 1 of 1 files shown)


svix-jbrown (commented):
what's the difference between `RoleId::operator` and `RoleId::admin`? Why would this change be made?

src/v1/endpoints/admin/auth_token.rs
```diff
@@ -33,8 +33,6 @@ use crate::{
```
tasn, line 33: @svix-jbrown, copying it here so that we have a thread:
> what's the difference between RoleId::operator and RoleId::admin? Why would this change be made?

Admin is a user admin, so a user of coyote that has access to the cluster and is an admin.
Operator is internal admin, so coyote itself interacting with coyote modules.

For more info: https://github.com/svix/rfc/pull/49/changes#diff-a8c57adad1d4a3a177092b147046ecad14d7d8ddbd709b75b617c6b6656db7b6R225

svix-jbrown (approved):
ok

PR #657, opened 2026-03-26
Support `max_wait_millis` for stream.receive and queue.receive

svix-gabriel:
This adds support for blocking receive calls for a certain duration until the batch size is reached.

The implementation is really rudimentary. The algorithm is essentially

```
if max_wait_millis is specified
    check if batch size messages are available using read_only db handles
    if batch size is not available, sleep
endif

do normal stream/queue receive raft operation
```

Since this is purely a read-only operation, none of the logic is in the raft layer, which I think makes sense.

There's definitely room for improving/optimizing this, potentially using a fancy tokio notification scheme so that when publish on a topic is called, we can potentially avoid sleeping for the entire max_wait_millis duration.

### DIFF
--- a/crates/coyote-core/src/monotime.rs
+++ b/crates/coyote-core/src/monotime.rs
@@ -60,6 +60,41 @@ impl Monotime {
         self.time.load(Ordering::Acquire)
     }
 
+    /// Sleeps for the given duration.
+    ///
+    /// In debug builds, polls with short real-time sleeps so that
+    /// [`Monotime::fast_forward`] causes the sleep to exit promptly.
+    /// In release builds, delegates to [`tokio::time::sleep`].
+    ///
+    /// IMPORTANT: You cannot call this within the raft layer, or you risk deadlocking.
+    #[cfg(debug_assertions)]
+    pub async fn sleep(&self, duration: std::time::Duration) {
+        use std::time::Instant;
+
+        /// Completely arbitrary, but used to make sure we don't inadvertently run forever in case you forget to fast forward time.
+        const MAX_SLEEP_REAL_TIME: std::time::Duration = std::time::Duration::from_secs(20);
+
+        let deadline = self.now() + duration;
+        let poll_interval = std::time::Duration::from_millis(10);
+        let t0 = Instant::now();
+        while t0.elapsed() < MAX_SLEEP_REAL_TIME {
+            tokio::time::sleep(poll_interval).await;
+            if self.now() >= deadline {
+                return;
+            }
+        }
+
+        panic!("Failed to fast forward time after {:?}", t0.elapsed())
+    }
+
+    /// Sleeps for the given duration.
+    ///
+    /// Delegates directly to [`tokio::time::sleep`].
+    #[cfg(not(debug_assertions))]
+    pub async fn sleep(&self, duration: std::time::Duration) {
+        tokio::time::sleep(duration).await;
+    }
+
     /// Advance time by the given amount. This should only be used in integration tests.
     #[cfg(debug_assertions)]
     pub fn fast_forward(&self, by: std::time::Duration) -> Timestamp {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -91,9 +91,6 @@ pub struct AppState {
 
     namespace_state: coyote_namespace::State,
 
-    // FIXME: do we need this?
-    // OTHERFIXME: yes, I think so.
-    #[allow(unused)]
     pub(crate) ro_dbs: ReadonlyDatabases,
 
     // FIXME: temporarily here until we make ro_dbs usable.
@@ -107,7 +104,6 @@ pub struct AppState {
 
     pub(crate) auth_token_cache: Arc<parking_lot::RwLock<core::auth::FifoCache<Permissions>>>,
 
-    #[allow(unused)]
     pub(crate) time: Monotime,
 }
 
--- a/src/v1/endpoints/msgs.rs
+++ b/src/v1/endpoints/msgs.rs
@@ -23,6 +23,7 @@ use coyote_msgs::{
 };
 use coyote_namespace::entities::NamespaceName;
 use coyote_proto::{AccessMetadata, MsgPackOrJson, RequestInput};
+use fjall_utils::{ReadableDatabase, ReadonlyConnection, StorageType};
 use jiff::Timestamp;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
@@ -215,6 +216,9 @@ struct MsgStreamReceiveIn {
     pub lease_duration_ms: DurationMs,
     #[serde(default)]
     pub default_starting_position: SeekPosition,
+    /// Maximum time (in milliseconds) to wait for messages before returning.
+    #[serde(default)]
+    pub batch_wait_ms: Option<DurationMs>,
 }
 
 request_input!(MsgStreamReceiveIn, "Receive");
@@ -238,6 +242,41 @@ async fn stream_receive(
         .namespace_state
         .fetch_namespace(data.namespace.as_deref())?
         .ok_or_not_found()?;
+
+    // FIXME(@svix-gabriel) - there's potentially optimizations we can do using fancy tokio
+    // Notifiers to avoid sleeping the max duration. But this is simplest for now.
+    if let Some(max_wait) = data.batch_wait_ms {
+        let ro_db = state.ro_dbs.db_for(StorageType::Persistent);
+        let metadata_ks = ro_db
+            .keyspace(coyote_msgs::METADATA_KEYSPACE)
+            .or_internal_error()?;
+        let msg_ks = ro_db
+            .keyspace(coyote_msgs::MSG_KEYSPACE)
+            .or_internal_error()?;
+
+        let ns_id = namespace.id;
+        let topic_name = data.topic.name().clone();
+        let cg = data.consumer_group.clone();
+        let now = state.time.now();
+        let batch_size = data.batch_size;
+
+        let estimated = coyote_core::task::spawn_blocking_in_current_span(move || {
+            coyote_msgs::estimate_available_stream_messages(
+                &metadata_ks,
+                &msg_ks,
+                ns_id,
+                &topic_name,
+                &cg,
+                now,
+            )
+        })
+        .await
+        .or_internal_error()??;
+
+        if (estimated as usize) < batch_size.get() as usize {
+            state.time.sleep(max_wait.into()).await;
+        }
+    }
 
     let operation = StreamReceiveOperation::new(
         namespace.id,
@@ -378,6 +417,9 @@ struct MsgQueueReceiveIn {
     pub batch_size: NonZeroU16,
     #[serde(default = "default_queue_lease_duration_ms")]
     pub lease_duration_ms: DurationMs,
+    /// Maximum time (in milliseconds) to wait for messages before returning.
+    #[serde(default)]
+    pub batch_wait_ms: Option<DurationMs>,
 }
 
 request_input!(MsgQueueReceiveIn, "Receive");
@@ -402,6 +444,41 @@ async fn queue_receive(
         .namespace_state
         .fetch_namespace(data.namespace.as_deref())?
         .ok_or_not_found()?;
+
+    // FIXME(@svix-gabriel) - there's potentially optimizations we can do using fancy tokio
+    // Notifiers to avoid sleeping the max duration. But this is simplest for now.
+    if let Some(max_wait) = data.batch_wait_ms {
+        let ro_db = state.ro_dbs.db_for(StorageType::Persistent);
+        let metadata_ks = ro_db
+            .keyspace(coyote_msgs::METADATA_KEYSPACE)
+            .or_internal_error()?;
+        let msg_ks = ro_db
+            .keyspace(coyote_msgs::MSG_KEYSPACE)
+            .or_internal_error()?;
+
+        let ns_id = namespace.id;
+        let topic_name = data.topic.name().clone();
+        let cg = data.consumer_group.clone();
+        let now = state.time.now();
+        let batch_size = data.batch_size;
+
+        let estimated = coyote_core::task::spawn_blocking_in_current_span(move || {
+            coyote_msgs::estimate_available_queue_messages(
+                &metadata_ks,
+                &msg_ks,
+                ns_id,
+                &topic_name,
+                &cg,
+                now,
+            )
+        })
+        .await
+        .or_internal_error()??;
+
+        if (estimated as usize) < batch_size.get() as usize {
+            state.time.sleep(max_wait.into()).await;
+        }
+    }
 
     let operation = QueueReceiveOperation::new(
         namespace.id,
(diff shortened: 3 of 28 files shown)


crates/coyote-core/src/monotime.rs
```diff
@@ -60,6 +60,31 @@ impl Monotime {
         self.time.load(Ordering::Acquire)
     }
 
+    #[cfg(debug_assertions)]
+    pub async fn sleep(&self, deadline: Timestamp) {
```
tasn: Shouldn't sleep accept a Duration (like it does in other sleep functions)? This way it's an easy search and replace to use it, but also not confusing using it forward.

crates/coyote-core/src/monotime.rs
```diff
@@ -60,6 +60,31 @@ impl Monotime {
...
+        // In debug builds, polls with short real-time sleeps so that
+        // [`Monotime::fast_forward`] causes the loop to exit promptly.
+        // Release builds don't need this, so we just call tokio::time::sleep once
+        let poll_interval = std::time::Duration::from_millis(10);
```
tasn, line 78: I'd ask JB to review this to make sure he didn't have something else in mind.

src/v1/endpoints/msgs.rs
```diff
@@ -239,6 +243,32 @@ async fn stream_receive(
...
+            namespace.id,
+            data.topic.name(),
+            &data.consumer_group,
+            state.time.now(),
+        )?;
+
+        if (estimated as usize) < data.batch_size.get() as usize {
+            let deadline = state.time.now() + max_wait;
+            state.time.sleep(deadline).await;
+        }
```
tasn: Seems incorrect tome? So essentially what happens is that if we can't find enough messages we always sleep until max time?

E.g. think:
1. max wait time is 10 minutes, batch size is 10.
2. I do a receive and it found 5 so it's waiting.
3. Now 1second later there are 10 more available.
4. It'll still sleep for 10 minutes instead of returning immediately.

src/v1/endpoints/msgs.rs
```diff
@@ -239,6 +243,32 @@ async fn stream_receive(
...
+    if let Some(max_wait) = data.max_wait_millis {
+        let ro_db = state.ro_dbs.db_for(StorageType::Persistent);
+        let metadata_ks = ro_db
+            .keyspace(coyote_msgs::METADATA_KEYSPACE)
+            .or_internal_error()?;
+        let msg_ks = ro_db
+            .keyspace(coyote_msgs::MSG_KEYSPACE)
+            .or_internal_error()?;
+
+        let estimated = coyote_msgs::estimate_available_stream_messages(
```
tasn: We should be optimistic: so make the case where performance matters most (we have enough messages, so there's a backlog) fast, and potentially make the "we don't have enough" (so we'll sleep anyway) slow. This makes the common case slow, which is not good.

crates/coyote-core/src/monotime.rs
svix-gabriel: Changed from `sleep` to `sleep_until` (which is also a tokio::time method).

src/v1/endpoints/msgs.rs
```diff
@@ -378,6 +408,9 @@ struct MsgQueueReceiveIn {
     pub batch_size: NonZeroU16,
     #[serde(default = "default_queue_lease_duration_ms")]
     pub lease_duration_ms: DurationMs,
+    /// Maximum time (in milliseconds) to wait for messages before returning.
+    #[serde(default)]
+    pub max_wait_millis: Option<DurationMs>,
```
tasn:
```suggestion
    pub batch_wait_ms: Option<DurationMs>,
```
(that's what we called it in the RFC)

src/v1/endpoints/msgs.rs
svix-gabriel: It's correct, just suboptimal. I called this out in the PR description.

I don't want this trying to read from fjall in a loop. I think there is room to use tokio::notify to speed things up, but it's a non-trivial optimization and I want to do that in a separate PR where it's easier to reason about the change.

src/v1/endpoints/msgs.rs
svix-gabriel: We can't be naively optimistic here because applying the `StreamReceiveOperation` will claim leases on partitions, preventing subsequent calls from working correctly if they happen before the leases have expired

We'd have to update `StreamReceiveOperation` to avoid claiming leases if the batch size is insufficient... which I think would get complicated quick. Not against it long term, but for getting something workable ASAP, I think the approach in this PR is simplest.

src/v1/endpoints/msgs.rs
svix-gabriel: Fixed

crates/coyote-core/src/monotime.rs
svix-gabriel: I guess cc @svix-jbrown - idk if you had something specific in mind for implementing a "monotonic aware sleep" implementation

svix-jbrown: I'm not sure what you'd want it to do. The options I see are:

 1. Just move the clock forward instantly by that amount of time (potentially confusing)
 2. Do what you're doing, and sleep until some other thread moves the clock forward by that amount of time.
 3. Mix an match them by, e.g., sleeping briefly to allow other tasks to run and then immediately fast-forwarding the lock

If we proceed with #2, we should definitely put a bound on how much real time we can spend sleeping so that people don't create infinite loops if there's no other task moving the clock.

crates/coyote-core/src/monotime.rs
```diff
@@ -60,6 +60,31 @@ impl Monotime {
...
+
+    #[cfg(not(debug_assertions))]
+    pub async fn sleep_until(&self, deadline: Timestamp) {
+        let now = self.now();
+        if now >= deadline {
+            return;
+        }
+        let remaining = deadline - now;
+        let millis = remaining.as_millis().max(0) as u64;
+        tokio::time::sleep(std::time::Duration::from_millis(millis)).await;
```
svix-jbrown: do we want to sleep until the monotonic clock is advanced past the deadline, or just until we think real time has advanced past the deadline? For this use-case, I kind of think we want to look at real wall-time in non-testing mode.

src/v1/endpoints/msgs.rs
```diff
@@ -239,6 +243,32 @@ async fn stream_receive(
...
+    if let Some(max_wait) = data.batch_wait_ms {
+        let ro_db = state.ro_dbs.db_for(StorageType::Persistent);
+        let metadata_ks = ro_db
+            .keyspace(coyote_msgs::METADATA_KEYSPACE)
+            .or_internal_error()?;
+        let msg_ks = ro_db
+            .keyspace(coyote_msgs::MSG_KEYSPACE)
+            .or_internal_error()?;
+
+        let estimated = coyote_msgs::estimate_available_stream_messages(
```
svix-jbrown: this function does a bunch of disk I/O without being inside a `tokio::task::spawn_blocking`, right?

src/v1/endpoints/msgs.rs
```diff
@@ -239,6 +243,32 @@ async fn stream_receive(
...
+            namespace.id,
+            data.topic.name(),
+            &data.consumer_group,
+            state.time.now(),
+        )?;
+
+        if (estimated as usize) < data.batch_size.get() as usize {
+            let deadline = state.time.now() + max_wait;
+            state.time.sleep_until(deadline).await;
+        }
```
svix-jbrown: I think this is also a weakness of having a single `batch_size` instead of Kafka's dueling `fetch.min.bytes` and `fetch.max.bytes` (which allows it to know that it should early-return once it hits `fetch.min.bytes`. We should think about a similar option.

src/lib.rs
```diff
@@ -91,8 +91,6 @@ pub struct AppState {
 
     namespace_state: coyote_namespace::State,
 
-    // FIXME: do we need this?
-    // OTHERFIXME: yes, I think so.
     #[allow(unused)]
```
svix-jbrown: we can remove this attr now, right?

crates/coyote-core/src/monotime.rs
tasn: Why have a sleep_unitl though, and not just sleep? You have a delay there anyway, so why do specific time anyway?

crates/coyote-core/src/monotime.rs
tasn: @svix-jbrown, reason why I mentioned you: the idea is having a sleep that plays well with the cluster time. I can see it being dangerous for two reasons:
1. We only update cluster time in raft, so it can make it so we sleep but never wake up, as the raft time is not updated while we process a raft task. Though I guess we never want to sleep in raft anyway.
2. The raft time can be jumpy/stuck if there are no raft operations while we sleep.

So what's the right way of doing it?

src/v1/endpoints/msgs.rs
tasn: This is worse than just suboptimal, it's very broken. As in, we can't release coyote like this, I bet tests will fail in staging, etc. What are your plans for the optimization? Why not just have it as a separate commit on top of this in the same PR?

src/v1/endpoints/msgs.rs
tasn: We can't sleep while in raft, so that's definitely not an option. Though even a loop + sleep 50ms to check how many available messages there are and then the moment we have enough go to raft will probably be better already.
The sleeping for the whole period is not a good assurance anything good will happen anyway, because if you have contention it could be that another consumer would go and read what you thought you'd have, and you end up with an empty receive anyway.

src/v1/endpoints/msgs.rs
tasn: The same issue would happen with `fetch.min.bytes` anyway though, right? He's just waiting until min is hit here (it just so happens that min==max).

src/lib.rs
svix-jplatte: pro tip: can use `#[expect(lint)]` for lints that are meant to be temporarily silenced until code is refactored. Then rustc will complain if the code under the attribute no longer triggers that lint.

src/v1/endpoints/msgs.rs
svix-gabriel: Yeah, what's the pattern we settled on. Just spamming `spawn_blocking` everywhere?

crates/coyote-core/src/monotime.rs
svix-gabriel: IDK this feels like an annoying nitpick.

`sleep`/`sleep_until` are both ubiquitous in Tokio. I think `sleep_until` is marginally easier here because of how we interact with the monotonic timestamps.

src/v1/endpoints/msgs.rs
svix-gabriel: It's emphatically not broken. The semantics of  `batch_wait_ms` are - block at most `batch_wait_ms` or until the batch size is accumulated. We are honoring those semantics in this PR, just with the possibility that we might end up waiting `batch_wait_ms` when could've theoretically waited `batch_wait_ms - n` (for some small value of n). That's not broken, it's just suboptimal.

And again, I hinted at the possible optimizations in the PR description.

> There's definitely room for improving/optimizing this, potentially using a fancy tokio notification scheme so that when publish on a topic is called, we can potentially avoid sleeping for the entire max_wait_millis duration.

I'm trying to get the API _working_ so I can integrate with webhooks-private ASAP.

src/v1/endpoints/msgs.rs
svix-gabriel: I think `fetch.min.bytes`/`fetch.max.bytes` would be even harder to implement, since we operate on msgs as unit, rather than bytes. I also personally found the concept of batch_sizes (in messages) much easier to understand, so I'd rather not change to what Kafka does.

crates/coyote-core/src/monotime.rs
svix-gabriel: I think the former, since we use the monotonic time to determine which messages are available

src/v1/endpoints/msgs.rs
svix-gabriel: > Though even a loop + sleep 50ms to check how many available messages there are and then the moment we have enough go to raft will probably be better already.

Then we're just spamming reads against Raft. It's doable, but not particularly great either.

Again, I called out how I'd want to optimize this without the sleep in the PR description

> There's definitely room for improving/optimizing this, potentially using a fancy tokio notification scheme so that when publish on a topic is called, we can potentially avoid sleeping for the entire max_wait_millis duration.

We can avoid excessive reads entirely using tokio notifiyers. I'm trying to get the API working so I can integrate with webhooks-private ASAP though, and that optimization is worth it's own PR

crates/coyote-core/src/monotime.rs
tasn: We can stop talking about it if you want. I was just confused because the code that calls it is doing `deadline = now() + delay; sleep_until(deadline)`, which is maybe unnecessary as you can just call `sleep(delay)`. Though my concern is that I want to encourage something that people would actually use, and literally everywhere in the code we use `sleep` and never `sleep_until`. I just grepped coyote and we have 35 sleeps and 0 sleep_until.

src/v1/endpoints/msgs.rs
tasn: > Then we're just spamming reads against Raft. It's doable, but not particularly great either.

Not in raft, just in the local fjall.

> We can avoid excessive reads entirely using tokio notifiyers. I'm trying to get the API working so I can integrate with webhooks-private ASAP though, and that optimization is worth it's own PR

Fine, but what I'm saying is that "integrate with webhooks-private" won't get you anywhere because the sleep 30 seconds or however long we wait, will make all the tests fail. So this is false urgency, and let's just fix it correctly instead of putting a non-solution.

src/v1/endpoints/msgs.rs
tasn: > It's emphatically not broken. The semantics of batch_wait_ms are - block at most batch_wait_ms or until the batch size is accumulated.

Only in the most narrow sense, and even them I'm not sure. The API should return when there's data, and wait at most batch_wait_ms, not the API should just always wait the full time and then return whatever it has.

Though as I said elsewhere: (1) this likely won't make tests pass in webhooks-private, so I don't understand the urgency to get a broken solution in, (2) because of how this is implemented, races are likely, as in it's likely you'll have two consumers going in parallel thinking that they have the full batch but then get to raft and only one has the batch, and the other is an empty receive, etc.

src/v1/endpoints/msgs.rs
tasn: Other modules have controllers so that it's encapsulated. I started working on doing that for stream, but then stopped because I saw you're touching the module.

src/v1/endpoints/msgs.rs
tasn: You can push this through, if you want, I just don't understand the point of reviewing something that doesn't work and rushing it in. It just makes it very hard for us to review this way.

src/v1/endpoints/msgs.rs
tasn: @svix-gabriel, he was talking about the min/max, not necessarily bytes, but just min msgs / max msgs.

svix-jbrown (approved):

PR #663, opened 2026-03-26
cluster: write snapshots to tempfiles so they do not leak on crash

svix-jbrown:
if the cluster shuts down half-way through a snapshot, it'll leave a partially-written file in place. This changes the snapshot mechanism to write to a temporary file (in the same directory, to allow for atomic `rename(2)` calls) and only put the snapshot under the final filename at the end.

### DIFF
--- a/src/core/cluster/state_machine.rs
+++ b/src/core/cluster/state_machine.rs
@@ -1,5 +1,6 @@
 use std::{
     io::{Seek, SeekFrom},
+    os::unix::fs::PermissionsExt,
     path::{Path, PathBuf},
     pin::Pin,
     sync::Arc,
@@ -190,7 +191,7 @@ impl Store {
         let mut this = Self {
             stores: Arc::new(RwLock::new(stores)),
             state: app_state,
-            snapshot_directory: snapshot_directory.into(),
+            snapshot_directory: snapshot_directory.as_path().canonicalize()?,
             readonly_meta_keyspace: ReadonlyKeyspace::from(meta_keyspace.clone()),
             meta_keyspace,
             last_snapshot: Arc::new(RwLock::new(None)),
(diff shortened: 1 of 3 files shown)


src/core/cluster/state_machine.rs
```diff
@@ -614,11 +615,16 @@ impl StoredSnapshot {
...
-            tracing::info!(path=%path_c.display(), "writing snapshot");
-            let mut f = std::fs::File::create(path_c)?;
-            serialized_state_machine::serialize_to_file(targets, &mut f)?;
+            let mut tf = tempfile::Builder::new()
+                .permissions(std::fs::Permissions::from_mode(0o600))
+                .tempfile_in(directory)?;
+            tracing::debug!(final_path=%path_c.display(), temp_path = %tf.path().display(), "writing snapshot");
+            serialized_state_machine::serialize_to_file(targets, tf.as_file_mut())?;
+            tf.as_file_mut().sync_all()?;
+            let mut f = tf.persist_noclobber(path_c)?;
```
svix-jplatte, line 627: How about using [the `atomic-write-file` crate](https://docs.rs/atomic-write-file/0.3.0/atomic_write_file/) instead?

svix-jbrown: why do I need another crate that does the same thing?

the atomic-write-file API seems kind of annoying because it doesn't return a handle to the file when you call `commit()`, so you need to re-open it (or call `.try_clone()` first on the inner File)

svix-jplatte: Ah, didn't realize we needed that (didn't look closely). I don't think we _need_ the crate, it's just easy to get some detail wrong (for example you thought to `sync_all`, I didn't the one time I handcoded this), so to me it feels like a _why not_ situation (generally). Apparently the crate also goes to extra lengths (`openat` etc.) to handle weird cases like the directory moving in between syscalls, but probably not something we need to worry about.

svix-jplatte (approved):

PR #665, opened 2026-03-26
cluster: do not run background tasks (e.g., snapshots) until bootstrap finishes

### LINKED ISSUE #661: don't allow snapshots until after bootstrap finishes

svix-jbrown:
fixes #661

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -314,6 +314,36 @@ async fn fail_until_bootstrapped(
     next.run(request).await
 }
 
+#[derive(Debug, Clone)]
+pub struct Initialized {
+    inner: Arc<tokio::sync::SetOnce<()>>,
+}
+
+impl Initialized {
+    fn new() -> Self {
+        Self {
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -314,6 +314,36 @@ async fn fail_until_bootstrapped(
...
+    pub async fn wait(self) -> coyote_error::Result<()> {
+        let shutting_down_token = shutting_down_token();
+        if shutting_down_token
+            .run_until_cancelled(self.inner.wait())
+            .await
+            .is_none()
+        {
+            return Err(Error::shutting_down());
+        }
+        Ok(())
```
svix-jplatte, line 343: This should be equivalent, right?

```suggestion
        shutting_down_token
            .run_until_cancelled(self.inner.wait())
            .await
            .ok_or_else(Error::shutting_down)
```

svix-jplatte (approved):

src/lib.rs
svix-jbrown: will do in follow-up PR

PR #668, opened 2026-03-26
Prepare permission checking in MsgpackOrJson extractor

svix-jplatte:
Part of https://github.com/svix/coyote-private/issues/68.

### DIFF
--- a/crates/coyote-authorization/src/lib.rs
+++ b/crates/coyote-authorization/src/lib.rs
@@ -1,6 +1,6 @@
-use std::{collections::HashMap, fmt};
+use std::{collections::HashMap, fmt, sync::Arc};
 
-use coyote_id::Module;
+use coyote_id::{AuthTokenId, Module};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
@@ -9,7 +9,7 @@ mod verification;
 
 pub use self::{
     pattern::{KeyPattern, NamespacePattern, ResourcePattern},
-    verification::verify_operation,
+    verification::{Forbidden, verify_operation},
 };
 
 #[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize, JsonSchema)]
@@ -104,4 +104,24 @@ pub struct RequestedOperation<'a> {
     pub namespace: Option<&'a str>,
     pub key: Option<&'a str>,
     pub action: &'static str,
+}
+
+impl RequestedOperation<'_> {
+    pub fn resource_str(&self) -> String {
+        let module = self.module;
+        let namespace = self.namespace.unwrap_or("");
+        let key = self.key.unwrap_or("**");
+        format!("{module}:{namespace}:{key}")
+    }
+}
+
+/// The `Permissions` for a request
+#[derive(Clone)]
+pub struct Permissions {
+    /// The role of the requester
+    pub role: RoleId,
+    /// The auth token id, if we used auth token
+    pub auth_token_id: Option<AuthTokenId>,
+    /// The access rules of the requester's role
+    pub access_rules: Arc<[AccessRule]>,
 }
--- a/crates/coyote-proto/src/msgpack_or_json.rs
+++ b/crates/coyote-proto/src/msgpack_or_json.rs
@@ -123,6 +123,8 @@ where
             }
         }
 
+        // let permissions = req.extensions().get::<Permissions>().cloned();
+
         let content_type = classify_content_type(req.headers())?;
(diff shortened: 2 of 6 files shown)


src/core/auth.rs
```diff
@@ -128,6 +118,8 @@ async fn authorization_inner(
             let perms = Permissions {
                 role: RoleId::from_string(out_token.metadata.remove("role").unwrap()),
                 auth_token_id: Some(out_token.id.into_inner()),
+                // FIXME: Resolve access rules from auth token
+                access_rules: [].into(),
```
svix-jplatte, line 122: This is the one integration point that needs to be implemented before auth tokens become useful. (can't be implemented here because roles and policies don't yet exist on disk, Tom has a branch)

tasn: I wonder if there's a better format to have there than a list of stuff that needs to be iterated every run. At the very least we probably want a sorted list. Though I wonder if there's a good data structure that would make sense for us. E.g. we already have numeric values for models, maybe an array based on the module number would make sense, and then fear each module you'll have the rules in a list.

crates/coyote-proto/src/msgpack_or_json.rs
```diff
@@ -266,6 +289,11 @@ impl IntoResponse for MsgPackOrJsonRejection {
                 StatusCode::UNPROCESSABLE_ENTITY,
                 ValidationErrorBody::new(errors),
             ),
+            Self::Forbidden => standard_error_response(
+                StatusCode::FORBIDDEN,
+                "forbidden",
+                "You do not have permission to use this operation",
```
tasn: Let's say which scope is needed, so that they know what to add.

crates/coyote-proto/src/msgpack_or_json.rs
```diff
@@ -138,6 +142,18 @@ where
         };
         value.validate().map_err(make_validation_error)?;
 
+        if let Some(perms) = permissions {
+            match value.access_metadata() {
+                AccessMetadata::AdminOnly => {}
```
tasn, line 146: There should never be anything that's admin only or rule protected. I don't even know what this means. Everything should be ONLY rule protected, it's just that admin is assumed to have all the rules allowed. Though that's still just part of the rule evaluation.

You should be able to add ACL rules to use whatever route you want! Btw, a thought about that: auth_token/policy/role, maybe the module name for them should be `admin/policy` and `admin/auth_token`? Otherwise things can get confusing/clashing with normal modules.

crates/coyote-authorization/src/lib.rs
```diff
@@ -91,7 +91,7 @@ impl AccessRule {
     }
 }
 
-#[derive(Debug, PartialEq, Eq, Deserialize, Serialize, JsonSchema)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Deserialize, Serialize, JsonSchema)]
```
tasn: Scary amount of clones. 😬

svix-jplatte: Actually maybe not needed, given I switched to `Arc<[AccessRule]>`.

crates/coyote-proto/src/msgpack_or_json.rs
svix-jplatte: Sure, this was just a shortcut for not having to think about what the rules for namespace mgmt and admin operations would look like. It's trivially backwards-compatible to introduce proper rules for them.

Created a ticket: https://github.com/svix/coyote-private/issues/683

src/core/auth.rs
svix-jplatte: Fortunately this is an implementation detail so we don't need to be optimal immediately. With the perf level we have now, this is surely not going to be a bottleneck.

I created a ticket: https://github.com/svix/coyote-private/issues/682

crates/coyote-proto/src/msgpack_or_json.rs
```diff
@@ -123,6 +123,8 @@ where
             }
         }
 
+        // let permissions = req.extensions().get::<Permissions>().cloned();
```
svix-gabriel, line 126: ?

svix-gabriel: Oh it's part of the `FIXME` below?

svix-gabriel (approved):

crates/coyote-proto/src/msgpack_or_json.rs
svix-jplatte: Yes.

PR #672, opened 2026-03-26
cluster: port from openraft 0.9 to 0.10-alpha

svix-jbrown:
This ports the cluster from openraft 0.9 to 0.10. I don't think these are compatible, but 0.10 is an alpha without much documentation so it's hard to know for sure.

Performance is still a mixed bag; perf when enabling fsync is much better due to the concurrent I/O there, but otherwise it's about the same.

I set up the new runtime-metrics flag (behind our own config flag) that dumps a bunch of stuff to stdout, but haven't really looked at it. I also added more otel metrics.

This is an absolute nightmare of merge conflicts, so if we're going to do it we should do it soon.

Part of #600.


src/core/cluster/logs.rs
```diff
@@ -369,7 +376,7 @@ impl CoyoteLogs {
                 .expect_point_read_hits(true)
         })?;
         let meta_keyspace = db.keyspace("cluster:meta", KeyspaceCreateOptions::default)?;
-        let (flush_tx, flush_rx) = tokio::sync::mpsc::channel(1000);
+        let (flush_tx, flush_rx) = tokio::sync::mpsc::channel(65536);
```
svix-jplatte, line 379: You want `unbounded_channel`?

src/core/metrics.rs
```diff
@@ -32,29 +38,55 @@ impl DbMetrics {
...
+        );
+    }
+
     pub fn bytes_used(&self, bytes: u64, db_type: DbType) {
         self.bytes_used.record(
             bytes,
             &[
-                opentelemetry::KeyValue::new("node_id", self.node_id.to_string()),
-                opentelemetry::KeyValue::new("db_type", db_type),
+                KeyValue::new("node_id", self.node_id.to_string()),
```
svix-jplatte, line 66: We should implement `From<NodeId> for opentelemetry::Value` so all these `.to_string()` calls aren't needed. (I can send a PR after this is merged)

svix-jplatte (approved):

src/core/cluster/logs.rs
svix-jbrown: I was actually matching the internal channel size in openraft (which is 65536 as of openraft 0.10). unbounded_channel scares me for OOMing if fsyncs are too slow

svix-jplatte: Okay. I took a small look at the implementation and it seems like tokio doesn't pre-allocate so fine I guess.

src/core/metrics.rs
svix-jbrown: https://github.com/svix/coyote-private/pull/692 does this

PR #677, opened 2026-03-26
Admin auth: add APIs for policy and role handling + fix rules loading

tasn:


src/v1/endpoints/admin/auth/policy.rs
```diff
@@ -0,0 +1,184 @@
...
+
+// Upsert
+
+#[derive(Clone, Debug, Deserialize, Serialize, Validate, JsonSchema)]
+pub struct AdminAccessPolicyUpsertIn {
+    pub id: AccessPolicyId,
+    pub description: String,
+    #[serde(default)]
+    pub rules: Vec<AccessRule>,
+}
```
svix-jbrown, line 60: this feels like another good thing to provide OCC on; otherwise, concurrent requests could get weird

crates/coyote-admin-auth/src/controller.rs
```diff
@@ -0,0 +1,214 @@
...
+        let db = self.db.clone();
+        let keyspace = self.keyspace.clone();
+        let id = id.clone();
+        spawn_blocking_in_current_span(move || {
+            if RoleRow::fetch(&keyspace, RoleRow::key_for(&id))?.is_none() {
+                return Ok(false);
+            }
+            let mut batch = db.batch();
+            batch.remove_row(&keyspace, RoleRow::key_for(&id))?;
+            batch.commit()?;
```
svix-jbrown, line 159: is there any reason to do a batch here instead of just doing `keyspace.remove(RoleRow::key_for(&id))`?

crates/coyote-admin-auth/src/controller.rs
```diff
@@ -0,0 +1,214 @@
...
+                description: input.description,
+                rules: input.rules,
+                policies: input.policies,
+                context: input.context,
+                created,
+                updated: input.now,
+            };
+            let mut batch = db.batch();
+            batch.insert_row(&keyspace, RoleRow::key_for(&row.id), &row)?;
+            batch.commit()?;
```
svix-jbrown, line 143: is there a reason to do a batch here instead of just doing `keyspace.insert(...)`?

crates/coyote-admin-auth/src/controller.rs
```diff
@@ -0,0 +1,214 @@
...
+            let row = AccessPolicyRow {
+                id: input.id,
+                description: input.description,
+                rules: input.rules,
+                created,
+                updated: input.now,
+            };
+            let mut batch = db.batch();
+            batch.insert_row(&keyspace, AccessPolicyRow::key_for(&row.id), &row)?;
+            batch.commit()?;
```
svix-jbrown, line 211: ditto

crates/coyote-admin-auth/src/controller.rs
```diff
@@ -0,0 +1,214 @@
...
+        let db = self.db.clone();
+        let keyspace = self.keyspace.clone();
+        let id = id.clone();
+        spawn_blocking_in_current_span(move || {
+            if AccessPolicyRow::fetch(&keyspace, AccessPolicyRow::key_for(&id))?.is_none() {
+                return Ok(false);
+            }
+            let mut batch = db.batch();
+            batch.remove_row(&keyspace, AccessPolicyRow::key_for(&id))?;
+            batch.commit()?;
```
svix-jbrown, line 227: ditto

crates/coyote-admin-auth/src/controller.rs
```diff
@@ -0,0 +1,214 @@
...
+        let keyspace = self.keyspace.clone();
+        let id = id.clone();
+        spawn_blocking_in_current_span(move || {
+            let row = AccessPolicyRow::fetch(&keyspace, AccessPolicyRow::key_for(&id))?;
+            Ok(row.map(AccessPolicyModel::from))
+        })
+        .await?
+    }
+
+    pub async fn list_policies(&self) -> Result<Vec<AccessPolicyModel>> {
```
svix-jbrown: it seems limiting to only be able to list all records (without even any pagination); what do you think the long-term goals should be for how to be able to list these records? I think that will decide what indices we need to build up-front.

crates/coyote-admin-auth/src/controller.rs
tasn: I'm just in the habit of always doing batches so that when we add something we don't forget. Doesn't really matter.

crates/coyote-admin-auth/src/controller.rs
tasn: Great timing, I was working on pagination in parallel, incoming!

tasn: It's already in both here and the other place.

src/v1/endpoints/admin/auth/policy.rs
tasn: Yeah, it depends on people doing fetch -> upsert rather than just upsert, and I suspect most uses would be just upsert to start (that's why I just created upsert and not insert/update as separate paths).

tasn:
Merged piecemeal in other PRs.

PR #678, opened 2026-03-27
ListResponse: add support for iterators, limit, and done.

tasn:
Implement it for auth_token.

### DIFF
--- a/crates/auth-token/src/controller.rs
+++ b/crates/auth-token/src/controller.rs
@@ -6,6 +6,7 @@ use fjall::{KeyspaceCreateOptions, KvSeparationOptions};
 use fjall_utils::{TableRow, WriteBatchExt};
 use jiff::Timestamp;
 use serde::{Deserialize, Serialize};
+use std::ops::Bound;
 
 use crate::tables::{AuthTokenEntity, AuthTokenRow, IdIndexRow, OwnerIndexRow};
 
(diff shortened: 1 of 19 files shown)


crates/auth-token/src/controller.rs
```diff
@@ -273,6 +275,25 @@ impl AuthTokenController {
...
+                    Some(idx) => {
+                        tokens = tokens.split_off(idx + 1);
+                    }
+                    None => {
+                        // Iterator ID not found; return empty to avoid returning
+                        // an inconsistent page.
+                        return Ok(Vec::new());
+                    }
+                }
+            }
```
svix-jbrown: this needs to actually do the pagination in fjall

svix-jbrown (commented):
see comment

crates/auth-token/src/controller.rs
tasn: Whoops, fixed!

crates/auth-token/src/controller.rs
```diff
@@ -255,22 +256,32 @@ impl AuthTokenController {
...
+
+            let start_id = iterator.unwrap_or(AuthTokenId::nil());
+            let start = OwnerIndexRow::owner_iter_start(namespace_id, &owner_id, start_id);
             let mut tokens = Vec::new();
-            for item in keyspace.prefix(&prefix) {
+            for item in keyspace.range((Bound::Excluded(start), Bound::Unbounded)) {
                 let key = item.key()?;
+                if !key.starts_with(&prefix) {
+                    break;
+                }
```
svix-jbrown, line 274: okay, this works but let's file a follow-up to add a helper method that does all of these shenanigans (constructing the start key, building teh range, breaking on limit, etc)

svix-jbrown (approved):

crates/auth-token/src/controller.rs
tasn: Yes, I agree. There was just only so much infra I was willing to build, already had to do a random set of things.

PR #684, opened 2026-03-27
Add parsing support for placeholders in access rule key patterns

svix-jplatte:
Part of #68.

### DIFF
--- a/crates/coyote-authorization/src/pattern.rs
+++ b/crates/coyote-authorization/src/pattern.rs
@@ -1,4 +1,8 @@
-use std::{borrow::Cow, fmt, str::FromStr};
+use std::{
+    borrow::Cow,
+    fmt::{self, Write},
+    str::FromStr,
+};
 
 use coyote_id::Module;
 use itertools::Itertools;
@@ -49,8 +53,23 @@ pub enum NamespacePattern {
 pub enum KeyPattern {
     Exactly(String),
     Prefix(String),
+    Segmented(SegmentedKeyPattern),
     Any,
-    // FIXME: Add single star wildcards
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct SegmentedKeyPattern {
+    pub segments: Vec<KeyPatternSegment>,
+    /// Whether the pattern has a trailing `*` segment.
+    ///
+    /// `*` is only allowed at the very end.
+    pub trailing_any: bool,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum KeyPatternSegment {
+    Fixed(String),
+    Placeholder(String),
 }
 
 impl ResourcePattern {
@@ -184,6 +203,7 @@ impl KeyPattern {
             Self::Exactly(k) => key == Some(k),
             Self::Prefix(p) => key.is_some_and(|k| k.starts_with(p)),
             Self::Any => true,
+            Self::Segmented(s) => key.is_some_and(|k| s.matches(k)),
         }
     }
 }
@@ -196,6 +216,7 @@ impl fmt::Display for KeyPattern {
                 f.write_str(s)?;
                 f.write_str("*")
             }
+            Self::Segmented(s) => s.fmt(f),
             Self::Any => f.write_str("*"),
         }
     }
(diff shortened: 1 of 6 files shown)


svix-jplatte:
Change of plans, `**` at the end just becomes `*` and we don't support other wildcards.

crates/coyote-authorization/src/pattern.rs
```diff
@@ -40,8 +44,23 @@ pub enum NamespacePattern {
 pub enum KeyPattern {
     Exactly(String),
     Prefix(String),
```
tasn: What is prefix for? Isn't that dealt with by segemented?

crates/coyote-authorization/src/pattern.rs
```diff
@@ -170,16 +191,126 @@ impl FromStr for KeyPattern {
         }
 
         if s.ends_with("/*") {
+            if s.contains(['$', '{', '}']) {
```
tasn: Seems a bit lax, we should probably just match this regex? `^\${.*}$` no? Probably faster too.

crates/coyote-authorization/src/pattern.rs
```diff
@@ -170,16 +191,126 @@ impl FromStr for KeyPattern {
...
+    type Err = &'static str;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s.contains('*') {
+            return Err("asterisk may only be used as a standalone slash-separated segment");
+        }
+
+        if let Some(rest) = s.strip_prefix("${")
+            && let Some(placeholder) = rest.strip_suffix('}')
+        {
```
tasn, line 302: Regex...

crates/coyote-authorization/src/pattern.rs
```diff
@@ -170,16 +191,126 @@ impl FromStr for KeyPattern {
...
+            // FIXME: probably disallow most characters inside?
+            // Though the pattern will just never match if invalid anyways.
+            return Ok(Self::Placeholder(placeholder.to_owned()));
+        }
+
+        if s.contains(['$', '{', '}']) {
```
tasn, line 308: Regex to disallow `${` not at the start. There's no problem with having `$`, `{`, or `}` otherwise.

crates/coyote-authorization/src/pattern.rs
svix-jplatte: Could be dealt with by segmented, yes. The `Exactly` variant too, actually. Having the stuff from before as more optimized code paths for the common case was easy, but if you prefer I can simplify the code and kick them out.

crates/coyote-authorization/src/pattern.rs
svix-jplatte: I would rather not depend on regex for this, and I really doubt it will help with performance (this shouldn't be that hot of a code path anyways).

crates/coyote-authorization/src/pattern.rs
svix-jplatte: Are you sure we should allow that? Then people could write `foo.${role}` and it would be accepted, but not do what's (probably) expected.

tasn: I'm saying something else, I'm saying: disallow `${` anywhere that's not at the start, and if it's at the start, it has to match the pattern I mentioned above. Though yeah, if you want you can disallow it otherwise if it makes it easier, I guess it's not a problem to not allow these characters at all for now.

crates/coyote-authorization/src/pattern.rs
tasn: It probably is hot path, it will potentially be called on every request no? (not all context is available at construction time - well maybe now, but once we add JWTs no. And maybe even once we add context for auth tokens - e.g. for the RLS use-case there).

Regex is not only faster though, it's also more correct, no? Because this line of code is not what we mean, the regex is.

crates/coyote-authorization/src/pattern.rs
tasn: I don't know if it's really that much faster (maybe), but having divergence in security code feels wrong, as it's more room for bugs and issues.

crates/coyote-authorization/src/pattern.rs
svix-jplatte: Context will be passed in separately, this is for the parsing of the pattern that includes placeholders (not expanded). We won't re-parse to evaluate against context.

crates/coyote-authorization/src/pattern.rs
svix-jplatte: I think we should disallow at least `$` entirely for now.

svix-jplatte:
Rebased and added a `pattern.rs` simplification commit.

crates/coyote-authorization/src/pattern.rs
tasn: Cool, do it. You can also disallow {}

crates/coyote-authorization/src/pattern.rs
tasn: Ah, makes sense.

tasn: Still feels less correct than regex (even without the performance!).

crates/coyote-authorization/src/pattern.rs
```diff
@@ -37,11 +41,18 @@ pub enum NamespacePattern {
...
-    Exactly(String),
-    Prefix(String),
-    Any,
-    // FIXME: Add single star wildcards
+pub struct KeyPattern {
+    pub segments: Vec<KeyPatternSegment>,
+    /// Whether the pattern has a trailing `*` segment.
+    ///
+    /// `*` is only allowed at the very end.
+    pub trailing_any: bool,
```
svix-jbrown, line 58: This doesn't match [the RFC](https://github.com/svix/rfc/blob/main/prds/coyote/auth.md), which says that any segment can have a trailing glob

crates/coyote-authorization/src/pattern.rs
```diff
@@ -166,20 +199,69 @@ impl FromStr for KeyPattern {
...
+            trailing_any,
+        })
+    }
+}
+
+impl KeyPatternSegment {
+    fn matches(&self, key_seg: &str) -> bool {
+        match self {
+            Self::Fixed(s) => s == key_seg,
+            // FIXME: Add support for context stuff
```
svix-jbrown, line 277: what context stuff

crates/coyote-authorization/src/pattern.rs
```diff
@@ -166,20 +199,69 @@ impl FromStr for KeyPattern {
...
+
+impl FromStr for KeyPatternSegment {
+    type Err = &'static str;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s.contains('*') {
+            return Err("asterisk may only be used as a standalone slash-separated segment");
+        }
+
+        if let Some(rest) = s.strip_prefix("${")
```
svix-jbrown, line 300: does this mean that `${` have to become banned characters in any authorizable object, or do we want to support escaping?

crates/coyote-authorization/src/pattern.rs
svix-jplatte: That's what this code already does.

svix-jbrown (commented):
This is the part of the RFC that I'm most skeptical of, because I think it's going to quite quickly turn into an expression-oriented domain-specific programming language and it seems to add a ton of unnecessary complexity. Once you start supporting this kind of interpolation, answering the question of "what resources can I see" becomes effectively impossible.

crates/coyote-authorization/src/pattern.rs
svix-jplatte: I'd rather not have escaping, but I'd say it's up for debate.

crates/coyote-authorization/src/pattern.rs
svix-jplatte: Yeah Tom suggested to change it end of last week and I went along with it. I already did one small RFC update, will have to do more.

crates/coyote-authorization/src/pattern.rs
svix-jplatte: This code has been deleted, if there are any concerns, do you think the code below also should use regex? I'm pretty strongly against regex in auth code tbh.

crates/coyote-authorization/src/pattern.rs
svix-jplatte: Hah. The actual values we want the placeholders to stand for, within a specific request. Probably just a kv map passed by reference into `matches` functions.

crates/coyote-authorization/src/pattern.rs
tasn: @svix-jbrown, the context is that we removed `**` and we now only support `*` and only at the end for simplicity. Because we felt it would be too complex for people to understand `**` and needing to do that, and that `*` doesn't actually catch all if a key just happens to have a `/` in it.

crates/coyote-authorization/src/pattern.rs
tasn: @svix-jplatte, you're very brief. :)

@svix-jbrown, we allow roles/tokens to have additional "variables" that can be used for substitution. So you can do stuff like `foo/${role}/baz`

crates/coyote-authorization/src/pattern.rs
tasn: I also asked for escaping, but @svix-jplatte was against it. I don't think it's the end of the world disallowing $, {, } which is what's currently done (and I also complained about). But I also think it's trivial to just use regex here, and allow these characters (and disallow `${`) which is probably a good idea.

svix-jplatte: I'm happy to implement it if we decide we want to support `${}`, but escapes seem like something regex is especially ill-suited for.

tasn: We can skip escaping for now, yeah.

tasn (approved):

PR #686, opened 2026-03-27
Add coyote-admin-auth crate

svix-jplatte:
Split out from #677 to ease review.
Part of #68.

### DIFF
--- a/crates/coyote-admin-auth/src/controller.rs
+++ b/crates/coyote-admin-auth/src/controller.rs
@@ -55,11 +55,10 @@ pub struct AccessPolicyModel {
     pub created: Timestamp,
     pub updated: Timestamp,
 }
-
-impl From<AccessPolicyRow> for AccessPolicyModel {
-    fn from(row: AccessPolicyRow) -> Self {
+impl AccessPolicyModel {
+    fn new(id: AccessPolicyId, row: AccessPolicyRow) -> Self {
         Self {
-            id: row.id,
+            id,
             description: row.description,
             rules: row.rules,
             created: row.created,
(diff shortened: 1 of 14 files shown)


crates/coyote-admin-auth/src/controller.rs
```diff
@@ -0,0 +1,244 @@
...
+                description: input.description,
+                rules: input.rules,
+                policies: input.policies,
+                context: input.context,
+                created,
+                updated: input.now,
+            };
+            let mut batch = db.batch();
+            batch.insert_row(&keyspace, RoleRow::key_for(&row.id), &row)?;
+            batch.commit()?;
```
svix-jbrown, line 142: same question as on tom's PR last night: is there a benefit to using a one-op batch rather than just calling `keyspace.insert`? Notably, the code for `Keyspace::insert` does not call into the batch code and appears a fair bit simpler.

crates/coyote-admin-auth/src/tables.rs
```diff
@@ -0,0 +1,66 @@
...
+impl AccessPolicyRow {
+    pub fn key_for(id: &AccessPolicyId) -> TableKey<Self> {
+        TableKey::init_key(Self::ROW_TYPE, &[], &[id.as_str()])
+    }
+
+    pub fn decode_fjall_key(key: &fjall::UserKey) -> Result<AccessPolicyId> {
+        assert!(key.len() >= 3);
+        assert!(key[0] == Self::ROW_TYPE);
+
+        let s = str::from_utf8(&key[1..key.len() - 1])
```
svix-jbrown, line 61: does this need  the end bound or could it just be
```suggestion
        let s = str::from_utf8(&key[1..])
```


on a more philosophical note, I hate all of these hard-coded decode functions and would love to make a derive macro that builds encode/decode functions for keys.

crates/coyote-admin-auth/src/controller.rs
svix-jplatte: I'll change it (on Mon) 👍

crates/coyote-admin-auth/src/tables.rs
svix-jplatte: Yes, end bound is to discard the nul byte.

svix-jbrown (approved):

PR #687, opened 2026-03-27
feat(gha): add `COYOTE_ADMIN_TOKEN` as k8s secret

ashayshub:
This PR adds `COYOTE_ADMIN_TOKEN` as k8s secret

Instead of providing the secret through `envVar` this change
adds a k8s secret for the same.

### DIFF
--- a/.github/workflows/svix-cloud-operator.yml
+++ b/.github/workflows/svix-cloud-operator.yml
@@ -11,6 +11,13 @@ on:
         description: "Target environment (svix-cloud-staging-apply or svix-cloud-dev-apply)"
         required: true
         type: string
+      namespace:
+        description: "K8s namespace"
+        type: string
+      coyote_admin_token:
+        description: "Coyote admin token"
+        type: string
+        default: "admin_abcdefghijlmnopqrstuvwxyz012345"
   workflow_call:
     inputs:
       image_tag:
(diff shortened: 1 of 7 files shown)


svix-james:
This looks fine, but I was really hoping we could have a way to supply this to Helm at install time and have the operator create/configure the secret as needed. At least, I think that's probably what would be easiest for external users of the Helm chart.

.github/workflows/svix-cloud-operator.yml
```diff
@@ -23,6 +30,10 @@ on:
       namespace:
         description: "K8s namespace"
         type: string
+      coyote_admin_token:
+        description: "Coyote admin token"
+        type: string
+        default: "admin_abcdefghijlmnopqrstuvwxyz012345"
```
svix-james, line 36: Default not needed here.

.github/workflows/svix-cloud-operator.yml
```diff
@@ -66,6 +77,19 @@ jobs:
...
+      - name: Install kubectl
+        uses: azure/setup-kubectl@v4
+
+      - name: Create COYOTE_ADMIN_TOKEN secret if not present
+        run: |
+          if kubectl get secret coyote-admin-token -n ${{ inputs.namespace }} > /dev/null 2>&1; then
+            echo "Secret coyote-admin-token is already present, skipping creation."
+          else
+            kubectl create secret generic coyote-admin-token \
+              --from-literal=COYOTE_ADMIN_TOKEN=${{ inputs.coyote_admin_token }} \
```
svix-james, line 89: We already have a secret for this that's being used elsewhere in CI, fyi

svix-james:
I would actually rather we not take this approach for now. Since ultimately we want the operator to manage this, let's just wait until we have that so we can dogfood it.

The secret right now is not sensitive, so we can just stick w/ what we currently have for now.

ashayshub:
Closing.

PR #690, opened 2026-03-27
refactor(clusterspec): use k8s-native EnvVar for extraEnv

ashayshub:
This PR ensures we use k8s-native EnvVar for extraEnv

First and foremost, this renames `spec.extraEnv` to `spec.envVar` 
This then adds  support for the fields `spec.envVar` to  extract
data from configmap or secrets (e.g. `valueFrom.secretRef`)
i.e. uses k8s_openapi native `EnvVar` type for `spec.envVar`.

### DIFF
--- a/infra/helm-coyote/charts/crds/crds/coyoteclusters.json
+++ b/infra/helm-coyote/charts/crds/crds/coyoteclusters.json
@@ -143,24 +143,11 @@
                     },
                     "type": "object"
                   },
-                  "extraEnv": {
-                    "default": [],
-                    "description": "Additional environment variables to inject into pods.",
+                  "envVar": {
+                    "description": "Additional environment variables to inject into pods.\nFollows the Kubernetes EnvVar API spec (v1.EnvVar): supports plain `value`\nand `valueFrom` (secretKeyRef, configMapKeyRef, fieldRef, resourceFieldRef)",
                     "items": {
-                      "description": "An environment variable to inject into pods.",
-                      "properties": {
-                        "name": {
-                          "type": "string"
-                        },
-                        "value": {
-                          "nullable": true,
-                          "type": "string"
-                        }
-                      },
-                      "required": [
-                        "name"
-                      ],
-                      "type": "object"
+                      "type": "object",
+                      "x-kubernetes-preserve-unknown-fields": true
                     },
                     "type": "array"
                   },
--- a/infra/operator/src/crd.rs
+++ b/infra/operator/src/crd.rs
@@ -3,7 +3,7 @@
 
 use std::collections::BTreeMap;
 
-use k8s_openapi::api::core::v1::{Affinity, Toleration, TopologySpreadConstraint};
+use k8s_openapi::api::core::v1::{Affinity, EnvVar, Toleration, TopologySpreadConstraint};
 use kube::CustomResource;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
@@ -65,8 +65,11 @@ pub(crate) struct CoyoteClusterSpec {
     pub api_port: u16,
 
     /// Additional environment variables to inject into pods.
-    #[serde(default)]
-    pub extra_env: Vec<EnvVar>,
+    /// Follows the Kubernetes EnvVar API spec (v1.EnvVar): supports plain `value`
+    /// and `valueFrom` (secretKeyRef, configMapKeyRef, fieldRef, resourceFieldRef)
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    #[schemars(schema_with = "env_var_schema")]
+    pub env_var: Vec<EnvVar>,
 
     /// CPU and memory resource requests and limits for the coyote container.
     #[serde(default)]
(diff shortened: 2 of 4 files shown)


infra/helm-coyote/charts/crds/crds/coyoteclusters.json
```diff
@@ -144,23 +144,10 @@
...
-                        "value": {
-                          "nullable": true,
-                          "type": "string"
-                        }
-                      },
-                      "required": [
-                        "name"
-                      ],
-                      "type": "object"
+                      "type": "object",
```
svix-james, line 149: This makes sense, but let's try and update the description to indicate we're effectively delegating to the native k8s type here

infra/helm-coyote/charts/crds/crds/coyoteclusters.json
```diff
@@ -144,23 +144,10 @@
                     "type": "object"
                   },
                   "extraEnv": {
```
svix-james: Maybe we should just rename this field to `envVars` to better match k8s standards? I'm not actually sure if that even is a standard.

infra/helm-coyote/charts/crds/crds/coyoteclusters.json
ashayshub: Done.

infra/operator/src/crd.rs
```diff
@@ -65,8 +65,11 @@ pub(crate) struct CoyoteClusterSpec {
     pub api_port: u16,
 
     /// Additional environment variables to inject into pods.
-    #[serde(default)]
-    pub extra_env: Vec<EnvVar>,
+    /// Follows the Kubernetes EnvVar API spec (v1.EnvVar): supports plain `value`
+    /// and `valueFrom` (secretKeyRef, configMapKeyRef, fieldRef, resourceFieldRef)
+    #[serde(default, rename = "envVar", skip_serializing_if = "Vec::is_empty")]
```
svix-james: We can remove the `rename`

infra/operator/Cargo.toml
```diff
@@ -12,7 +12,7 @@ path = "src/main.rs"
 [dependencies]
 anyhow.workspace = true
 futures-util.workspace = true
-k8s-openapi.workspace = true
+k8s-openapi = { workspace = true, features = ["schemars"] }
```
svix-james: Do we actually need this?

ashayshub: This is what I get

```
 - schemars.workspace = true brings in the schemars crate itself — the library that provides the JsonSchema trait, the SchemaGenerator, json_schema! macro,  
  etc.                                                                                                                                                        
  - k8s-openapi = { features = ["schemars"] } is a feature flag inside the k8s-openapi crate that tells it to compile its own conditional code — specifically 
  the impl JsonSchema for EnvVar, impl JsonSchema for Affinity, etc. Those impls are gated behind #[cfg(feature = "schemars")] inside k8s-openapi's source and
   won't exist at all without it.
                                                                                                                                                              
  So they're orthogonal — one says "I want the schemars library", the other says "I want k8s-openapi to implement schemars traits on its types". You need     
  both:
                                                                                                                                                              
  - Without schemars.workspace = true → no JsonSchema trait, json_schema! macro, or SchemaGenerator at all.                                                   
  - Without k8s-openapi features = ["schemars"] → EnvVar and friends don't implement JsonSchema, so CoyoteClusterSpec can't derive it either (since all fields
   must implement JsonSchema).                                                                     
   ```
   
  I haven't gone into details but apparently this is a `k8s-openapi` specific feature which is required during json schema generation.

File: infra/helm-coyote/charts/crds/crds/coyoteclusters.json (hunk 2 above)
Comment (ashayshub) [reply infra/helm-coyote/charts/crds/crds/coyoteclusters.json]: Done.

File: infra/operator/Cargo.toml (hunk 4 above)
Comment (ashayshub) [reply infra/operator/Cargo.toml]:
```

When schemars processes a field annotated with #[schemars(schema_with = "...")], the derive macro generates a call to your function instead of              
  T::json_schema(). This means it never needs T: JsonSchema to be satisfied at all — the trait bound is bypassed entirely.
                                                                                                                                                              
  So:                                                       
  - Affinity, Toleration, TopologySpreadConstraint all have schema_with → never needed features = ["schemars"]
  - The original EnvVar field had no schema_with → the derive macro called EnvVar::json_schema() directly → required the feature                              
                                                                                                                                
  Which leads to an interesting consequence: now that we added schema_with to envVar as well, there are no longer any k8s-openapi types in this file that     
  require JsonSchema to be implemented on them. The features = ["schemars"] in Cargo.toml may now be unnecessary for this crate — assuming no other files in  
  the crate use k8s-openapi types without schema_with.                                                                                                        
                                                                                                                                                              
  Worth checking the rest of the codebase before removing it though.                      
  ```
Removed, when digged further apparently it is not required because of the presence of `schema_with` for each of those k8s-openapi types.

infra/operator/src/crd.rs
ashayshub: Removed.

svix-james (approved):
This makes sense, and it looks dumb that

PR #692, opened 2026-03-27
server: clean up metrics a bit

svix-jbrown:
This is branched off of #672.

It refactors metrics so that we aren't cloning strings constantly. Specific changes:

 - Implements `From<NodeId>` for `opentelemetry::KeyValue` using an `Arc<str>` so that the clones are cheapish
 - When possible, shares the state (like the Connection Type and the Node ID) between the whole metrics instance instead of recreating it on every metric emission
 - Adds request metrics to the interserver port

### DIFF
--- a/src/core/cluster/node.rs
+++ b/src/core/cluster/node.rs
@@ -1,4 +1,5 @@
 use anyhow::Context;
+use opentelemetry::{KeyValue, Value};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use std::net::SocketAddr;
@@ -124,5 +125,20 @@ impl From<u64> for NodeId {
     fn from(value: u64) -> Self {
         let inner = Uuid::from_u64_pair(value, value);
         Self { inner }
+    }
+}
+
+impl From<NodeId> for Value {
+    fn from(value: NodeId) -> Self {
+        // we want to make this cheaply cloneable since we use it in
+        // metrics all the time, and it's low-cardinality
+        let value: std::sync::Arc<str> = value.to_string().into();
+        Self::String(value.into())
+    }
+}
+
+impl From<NodeId> for KeyValue {
+    fn from(value: NodeId) -> Self {
+        KeyValue::new("node_id", value)
     }
 }
--- a/src/core/metrics.rs
+++ b/src/core/metrics.rs
@@ -1,5 +1,3 @@
-use std::fmt::{self, Display, Formatter};
-
 use http::StatusCode;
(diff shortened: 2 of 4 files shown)


src/core/cluster/node.rs
```diff
@@ -126,3 +127,18 @@ impl From<u64> for NodeId {
         Self { inner }
     }
 }
+
+impl From<NodeId> for Value {
+    fn from(value: NodeId) -> Self {
+        // we want to make this cheaply cloneable since we use it in
+        // metrics all the time, and it's low-cardinality
+        let value: std::sync::Arc<str> = value.to_string().into();
```
svix-jbrown, line 135: something else I thought about here was making it a `Box<str>` and then calling `Box::leak` to make it a `&'static str`, since in most situations it lives forever anyway and there's one per invocation (except tests, I guess). would you prefer that?

svix-jplatte: Definitely not in a trait impl that might be invoked from anywhere implicitly. Even this impl is a bit weird since it would be worse than the naive `String` one in case the resulting `Value` is never cloned, but probably fine.

I doubt the refcounting will ever be relevant, so I'd say let's keep it like this.

src/core/metrics.rs
```diff
@@ -46,27 +50,20 @@ impl DbMetrics {
...
     }
 
     pub fn record_apply(&self, batch_size: usize) {
         self.apply_operations
-            .record(1, &[KeyValue::new("node_id", self.node_id.to_string())]);
-        self.apply_batch_size.record(
-            batch_size as u64,
-            &[KeyValue::new("node_id", self.node_id.to_string())],
-        );
+            .record(1, std::slice::from_ref(&self.node_id_kv.clone()));
```
svix-jplatte:
```suggestion
            .record(1, std::slice::from_ref(&self.node_id_kv));
```

src/core/cluster/node.rs
svix-jbrown: Another dumb idea I had was to add `string_version: &'static str` to the NodeId struct and `Box::leak` on NodeId construction to get the same effect.

Can't add `Arc<str>` because NodeId needs to be Copy.

svix-jplatte (dismissed):
Nice. This is gonna clash with #693, but I can deal with it.

src/core/cluster/node.rs
svix-jplatte: Doesn't sound too bad.

svix-jplatte (approved):

Issue #531, opened 2026-03-19
Copy over missing middlewares from svix-webhooks-private


tasn:
- [ ] There are quite a few there, and I think we need at least some of them (see example below).
- [ ] The intra-cluster and external have very different middlewares attached, we should probably unify.

Example:
```
    let app = app
        .route(
            "/_/cache-stats",
            axum::routing::get(cache_stats).with_state(cache),
        )
        .layer((
            axum::middleware::from_fn(move |req: Request, next: Next| {
                // Make configuration available for the whole middleware stack through a task-local
                svix_server_data::CFG.scope(cfg.clone(), next.run(req))
            }),
            SetResponseHeaderLayer::overriding(CACHE_CONTROL, HeaderValue::from_static("no-store")),
            layer_fn(move |service| IdempotencyService::new(lock_cache.clone(), service, &meter)),
            axum::middleware::from_fn(svix_server_data::tracing::request_id::request_id_middleware),
            TraceLayer::new_for_http()
                .make_span_with(AxumOtelSpanCreator)
                .on_request(DefaultOnRequest::new().level(tracing::Level::TRACE))
                .on_response(AxumOtelOnResponse)
                .on_failure(AxumOtelOnFailure),
            axum::middleware::from_fn_with_state(
                app_state,
                svix_server_data::tracing::otel_spans::middleware::route_metrics,
            ),
            CorsLayer::new()
                .allow_origin(Any)
                .allow_methods(Any)
                .allow_headers(AllowHeaders::mirror_request())
                .max_age(Duration::from_secs(600)),
            CompressionLayer::new(),
            TimeoutLayer::new(
                // This is a catch-all timeout with a slightly larger deadline.
                // Specific routes like `create_message` have a smaller deadline but give a
                // `4xx` response instead of `5xx` within the `BodyLimitHandler` layer.
                BODY_READ_DEADLINE + Duration::from_secs(15),
            ),
            axum::middleware::from_fn(svix_server_core::metrics::collect_route_task_metrics),
        ))
        .fallback(v1::no_route_exists_handler)
        // It is important that this service wraps the router instead of being
        // applied via `Router::layer`, as it would run after routing then.
        .trim_trailing_slash();
```

svix-jplatte:
Closing as per Slack discussion: https://svixhq.slack.com/archives/C0A72988962/p1775553536371099

PR #693, opened 2026-03-27
Refactor server initialization

svix-jplatte:
… and add response compression, which was the only thing from #531 that I felt we definitely should also have here.


svix-jplatte:
Looks like I legit broke CI, will split up and figure out what's wrong tomorrow.

svix-jplatte:
Closing in favor of #706 which now contains all of the same changes (and one more).

PR #694, opened 2026-03-27
Bundle JS library with tsdown

svix-lucho:
Add [tsdown](https://github.com/rolldown/tsdown/) to bundle the JS/TS lib for ESM Node and browsers. 

I decided to use a bundler instead of using plain `tsc` because `tsc` emits `.js` files and keeps extensionless import `./example`. Node’s ESM loader won’t resolve those paths, so the only way to make that work for Node was to add `.js` extensions to all imports in the source files, which seemed very ugly.

`tsdown` seems very simple, and we can configure it to bundle the lib for whatever target we want to quite easily. It's not webpack/vite config hell. And it's built on top of [Rolldown](https://rolldown.rs/), a bundler written in Rust. 

This also cleans up test related code that was unused. When we add tests here, we can add it back.

### DIFF
--- a/z-clients/javascript/package-lock.json
+++ b/z-clients/javascript/package-lock.json
@@ -14,7 +14,76 @@
       "devDependencies": {
         "@biomejs/biome": "2.3.15",
         "mockttp": "^3.15.5",
+        "tsdown": "^0.21.6",
         "typescript": "^5.9.3"
+      }
+    },
+    "node_modules/@babel/generator": {
+      "version": "8.0.0-rc.3",
+      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-8.0.0-rc.3.tgz",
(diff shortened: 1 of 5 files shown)


svix-jbrown (approved):
I've never heard of tsdown before and I have a little intrinsic skepticism of "large project made by one committer in the last 10 months or so" but I guess we can give it a shot

svix-jplatte:
Doesn't tsc have a flag for appending the file endings maybe? ISTR there was something about allowing ts endings in source at least 🤔

svix-lucho:
> I've never heard of tsdown before and I have a little intrinsic skepticism of "large project made by one committer in the last 10 months or so" but I guess we can give it a shot

btw `tsdown` is just 'the new tsup'. That one commiter is the maintainer of `tsup` ([which is not actively maintained anymore](https://github.com/egoist/tsup))

PR #695, opened 2026-03-27
refactor(codegen): use selectattr on field loop

svix-pluiz:
Pull request inspired by https://github.com/svix/coyote-private/pull/620

Apply the `{%- for field in type.fields | selectattr("positional", "false") %}` syntax sugar in java, ts and py structs


svix-jbrown (commented):
please run `cargo codegen` and check in the result

svix-jplatte:
There are none, this is a pure template refactoring. (otherwise CI would be failing)

svix-jplatte (approved):

PR #696, opened 2026-03-27
Fix codegen bug in JS client

svix-lucho:
Fix a bug in the template that generated code that didn't compile because of differences in expected casing.

Also update the github workflow to actually run `npm build` in CI. That's why this never came up 😛


svix-james (approved):

z-clients/javascript/src/apis/msgs.ts
```diff
@@ -43,7 +43,7 @@ export class Msgs {
         request.setBody(
             MsgPublishInSerializer._toJsonObject({
                 ...msgPublishIn,
-                topic,
+                topic: topic,
```
svix-jbrown, line 46: can we suppress this duplication when they're the same?

PR #702, opened 2026-03-27
Add basic metrics to `msgs` module

svix-james:
Basic metrics added to the different operations. For now I've avoided the probably more useful metrics like queue depth/topic lag because I'm not yet sure how to do it efficiently. We'll get them in the next PR.

### DIFF
--- a/crates/msgs/src/metrics.rs
+++ b/crates/msgs/src/metrics.rs
@@ -1,0 +1,167 @@
+use opentelemetry::metrics::{Counter, Meter};
+
+#[derive(Clone)]
+pub struct MsgMetrics {
+    pub(crate) published: Counter<u64>,
+    pub(crate) published_bytes: Counter<u64>,
+
+    pub(crate) queue_received: Counter<u64>,
+    pub(crate) queue_acked: Counter<u64>,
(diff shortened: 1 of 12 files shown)


crates/msgs/src/metrics.rs
```diff
@@ -0,0 +1,167 @@
...
+                .with_description("Stream seek operations")
+                .with_unit("{event}")
+                .build(),
+        }
+    }
+}
+
+impl MsgMetrics {
+    pub(crate) fn record_published(&self, topic: &str, msg_count: u64, bytes: u64) {
+        let attrs = &[opentelemetry::KeyValue::new("topic", topic.to_owned())];
```
svix-jbrown: instead of calling `.to_string()` on TopicName, then passing it as a `&str`, then calling `.to_owned()` on it, why don't we make these take `&TopicName` and `&ConsumerGroup`, then implement `From<&TopicName> for KeyValue` so that we can get some type-safety and reduce the number of string clones by one per item.

svix-james: Done

svix-jplatte (approved):

PR #703, opened 2026-03-27
cluster: improve discovery handling

### LINKED ISSUE #701: make cluster discovery restartable
Right now, if the server shuts down after adding itself as a learner to the cluster but before actually joining as a follower, the only option is to remove it (manually with `coyote cluster-admin remove-node ...`) and restart.

The Discovery process should be able to determine that it's currently a learner and then do the "wait to catch up and then upgrade" flow all by itself.

svix-jbrown:
There's actually not much to do here; openraft already makes it safe to re-call `AddLearner` with an already-learner node, so we just need to make sure that discovery still gets called if we're in the Learner state.

Fixes #701.

### DIFF
--- a/src/core/cluster/discovery.rs
+++ b/src/core/cluster/discovery.rs
@@ -110,8 +110,9 @@ impl Discovery {
         let Some(log_id) = leader_cluster.last_committed_log_id else {
             anyhow::bail!("existing cluster has no logs");
         };
-        // TODO: if any of these steps fail, how do we recover?
         let client = self.network.client_for(leader_node_id, &leader_addr.into());
+        // add_learner and upgrade_learner are both safe to retry if a node shuts down
+        // so on failure, just try try again
         client
             .add_learner(AddLearnerRequest {
                 node_id: self.my_node_id,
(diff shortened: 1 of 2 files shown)


src/core/cluster/discovery.rs
```diff
@@ -110,8 +110,9 @@ impl Discovery {
         let Some(log_id) = leader_cluster.last_committed_log_id else {
             anyhow::bail!("existing cluster has no logs");
         };
-        // TODO: if any of these steps fail, how do we recover?
         let client = self.network.client_for(leader_node_id, &leader_addr.into());
+        // add_learner and upgrade_learner are both safe to retry if a node shuts down
+        // so on failure, just try try again
```
svix-jplatte, line 115:
```suggestion
        // so on failure, just try again
```

svix-jplatte (approved):

src/core/cluster/discovery.rs
svix-jbrown: ah, "try, try again" is an english idiom

PR #706, opened 2026-03-30
Refactor server initialization

svix-jplatte:
… and add response compression, which was the only thing from #531 that I felt we definitely should also have here.


svix-jbrown (dismissed):

svix-jplatte:
Merge conflicts :melting_face:

svix-jbrown (dismissed):

tasn (approved):

PR #710, opened 2026-03-30
Allow access rule action to be a wildcard

svix-jplatte:
Extracted from #677, for separate discussion.

### DIFF
--- a/crates/coyote-authorization/src/lib.rs
+++ b/crates/coyote-authorization/src/lib.rs
@@ -87,7 +87,11 @@ pub struct AccessRule {
 
 impl AccessRule {
     pub fn matches(&self, operation: &RequestedOperation<'_>) -> bool {
-        self.resource.matches(operation) && self.actions.iter().any(|a| a == operation.action)
+        self.resource.matches(operation)
+            && self
+                .actions
+                .iter()
+                .any(|a| a == "*" || a == operation.action)
     }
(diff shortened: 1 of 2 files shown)


crates/coyote-authorization/src/lib.rs
```diff
@@ -87,7 +87,11 @@ pub struct AccessRule {
 
 impl AccessRule {
     pub fn matches(&self, operation: &RequestedOperation<'_>) -> bool {
-        self.resource.matches(operation) && self.actions.iter().any(|a| a == operation.action)
+        self.resource.matches(operation)
+            && self
+                .actions
+                .iter()
+                .any(|a| a == "*" || a == operation.action)
```
svix-jplatte, line 94: This seems a bit weird to me that for the wildcard case, we'd still have a _list_ of actions. On the other hand, supporting both `actions: "*"` and `actions: ["create", "read"]` in the SDKs would definitely require some hand-writing of models.

tasn: `["*"]` is the correct syntax. Which is consistent, as we'll probably want (In the future) to support `["some_prefix.*"]`

svix-jplatte: Okay.

svix-jbrown (approved):

PR #715, opened 2026-03-30
Add auth role API

svix-jplatte:
Extracted out of #677, part of #68.


src/v1/endpoints/admin/auth/role.rs
```diff
@@ -0,0 +1,210 @@
...
+use validator::Validate;
+
+use crate::{
+    AppState,
+    core::cluster::RaftState,
+    error::Result,
+    v1::utils::{ListResponse, ListResponseItem, Pagination, openapi_tag},
+};
+
+fn role_out(model: RoleModel) -> AdminRoleOut {
```
tasn, line 29: Why not a From<> or something? No biggie, just wondering.

tasn (approved):

src/v1/endpoints/admin/auth/role.rs
svix-jplatte: This is straight from #677, but you're right. I'll make a follow-up PR.

Issue #659, opened 2026-03-26
check in (automatically generate?) an example TOML configuration file


svix-jbrown:
It would be nice to have a well-commented example configuration file in the root of the repo. It'd be even better if we could figure out how to generate this from the serde.

We should also clearly document all of the environment variables we support so that prospective users don't need to read the Rust code.

tasn:
Yes, 100%!

svix-jplatte:
> It'd be even better if we could figure out how to generate this from the serde.

How about using the `schemars::JsonSchema` derive? Mentioned before in the previous ticket about this: https://github.com/svix/coyote-private/issues/197

PR #717, opened 2026-03-30
server: add a command to dump out the config as toml

svix-jbrown:
This isn't terribly useful without comments, but at least it exposes the values without making people read the source.

Part of #659

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -60,9 +60,9 @@ fn dump_config(cfg: Configuration, path: Option<PathBuf>) -> anyhow::Result<()> 
     {
         let f = File::open(path)?;
         let mut bf = BufWriter::new(f);
-        write!(bf, "{}", str)?;
+        write!(bf, "{str}")?;
     } else {
-        print!("{}", str);
+        print!("{str}");
     }
     Ok(())
 }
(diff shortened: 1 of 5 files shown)


tasn:
I think this is a good step forward, but 100% in favor of doing what we do in webhooks-private, and have the default values be loaded from a config file instead of default.rs. Feels like a huge regression.

svix-jbrown:
> I think this is a good step forward, but 100% in favor of doing what we do in webhooks-private, and have the default values be loaded from a config file instead of default.rs. Feels like a huge regression.

Even if we change it, I still think we should land this to have a way to render what the effective config is, inclusive of all overrides.

server/src/main.rs
```diff
@@ -5,9 +5,17 @@
...
 use clap::{Parser, Subcommand};
-use coyote::{cfg, run};
+use coyote::{
+    cfg::{self, Configuration},
+    run,
+};
 use dotenvy::dotenv;
 use mimalloc::MiMalloc;
+use std::{
+    fs::File,
```
svix-jplatte: Let's use `fs_err::File`.

svix-jplatte (approved):

PR #721, opened 2026-03-30
feat(clusterspec): add `adminToken` to the clusterspec

ashayshub:
This PR adds `adminToken` to coyote clusterspec

Currently it accepts the admin token through env variable.
This ensures CRD has `adminToken` to coyote clusterspec CRD.
This allows `adminToken` to accept a secretRef or a String value. 
This also ensures that envVar accepts both `value` and `value_from` 
fields if available in the CR.

### DIFF
--- a/infra/helm-coyote/charts/crds/crds/coyoteclusters.json
+++ b/infra/helm-coyote/charts/crds/crds/coyoteclusters.json
@@ -43,10 +43,752 @@
               "spec": {
                 "description": "A Coyote cluster deployment.",
                 "properties": {
+                  "adminToken": {
+                    "anyOf": [
+                      {
+                        "required": [
+                          "value"
+                        ]
+                      },
+                      {
+                        "required": [
+                          "valueFrom"
+                        ]
+                      }
+                    ],
+                    "description": "Admin token for privileged API access.\nSet either `value` for a plain string or `valueFrom` to pull from a Kubernetes Secret.\nUsing a plain string value is only recommended for testing. Use `valueFrom` in all other cases.",
+                    "nullable": true,
+                    "properties": {
+                      "value": {
+                        "type": "string"
+                      },
+                      "valueFrom": {
+                        "description": "SecretKeySelector selects a key of a Secret.",
+                        "properties": {
+                          "key": {
+                            "description": "The key of the secret to select from.  Must be a valid secret key.",
+                            "type": "string"
+                          },
+                          "name": {
+                            "description": "Name of the referent. This field is effectively required, but due to backwards compatibility is allowed to be empty. Instances of this type with an empty value here are almost certainly wrong. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names",
+                            "type": "string"
+                          },
+                          "optional": {
+                            "description": "Specify whether the Secret or its key must be defined",
+                            "type": "boolean"
+                          }
+                        },
+                        "required": [
+                          "key",
+                          "name"
+                        ],
+                        "type": "object"
+                      }
+                    },
+                    "type": "object"
+                  },
                   "affinity": {
                     "description": "Affinity rules for advanced pod scheduling (node affinity, pod affinity/anti-affinity).",
-                    "type": "object",
-                    "x-kubernetes-preserve-unknown-fields": true
+                    "nullable": true,
+                    "properties": {
+                      "nodeAffinity": {
+                        "description": "Describes node affinity scheduling rules for the pod.",
+                        "properties": {
+                          "preferredDuringSchedulingIgnoredDuringExecution": {
+                            "description": "The scheduler will prefer to schedule pods to nodes that satisfy the affinity expressions specified by this field, but it may choose a node that violates one or more of the expressions. The node that is most preferred is the one with the greatest sum of weights, i.e. for each node that meets all of the scheduling requirements (resource request, requiredDuringScheduling affinity expressions, etc.), compute a sum by iterating through the elements of this field and adding \"weight\" to the sum if the node matches the corresponding matchExpressions; the node(s) with the highest sum are the most preferred.",
+                            "items": {
+                              "description": "An empty preferred scheduling term matches all objects with implicit weight 0 (i.e. it's a no-op). A null preferred scheduling term matches no objects (i.e. is also a no-op).",
+                              "properties": {
+                                "preference": {
+                                  "description": "A node selector term, associated with the corresponding weight.",
+                                  "properties": {
+                                    "matchExpressions": {
+                                      "description": "A list of node selector requirements by node's labels.",
+                                      "items": {
+                                        "description": "A node selector requirement is a selector that contains values, a key, and an operator that relates the key and values.",
+                                        "properties": {
+                                          "key": {
+                                            "description": "The label key that the selector applies to.",
+                                            "type": "string"
+                                          },
+                                          "operator": {
+                                            "description": "Represents a key's relationship to a set of values. Valid operators are In, NotIn, Exists, DoesNotExist. Gt, and Lt.",
+                                            "type": "string"
+                                          },
+                                          "values": {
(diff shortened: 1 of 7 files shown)


infra/helm-coyote/charts/crds/crds/coyoteclusters.json
```diff
@@ -43,6 +43,69 @@
               "spec": {
                 "description": "A Coyote cluster deployment.",
                 "properties": {
+                  "adminToken": {
```
svix-james, line 46: Am I going to see this value in the spec if I do a `kubectl describe coyotecluster {}`?

ashayshub: Not if it is being set through secretref/configmapref . Yes if it is being set as string value. It is the user discretion. We can recommend keeping it as secretRef without forcing. string is useful in unit testing for example.

infra/svix-values.yaml
```diff
@@ -50,3 +50,4 @@ cluster:
         operator: Equal
         value: coyote
         effect: NoSchedule
+    adminToken: "helloWorld"
```
svix-james: Why are we using this instead of `COYOTE_ADMIN_TOKEN` value?

ashayshub: Fixed.

svix-james:
FYI, looks like you need to do a `cargo fmt`.

But, I'm wondering -- can we compare this to how some other operator handles this sort of thing. I just don't have a good feel for whether this is the right way to approach it. Overall, it looks good, but I think it would be ideal if it were possible to not have to worry about this appearing in the spec even in the case of it being a raw string.

ashayshub:
> FYI, looks like you need to do a `cargo fmt`.
> 
> But, I'm wondering -- can we compare this to how some other operator handles this sort of thing. I just don't have a good feel for whether this is the right way to approach it. Overall, it looks good, but I think it would be ideal if it were possible to not have to worry about this appearing in the spec even in the case of it being a raw string.

Using SecretKeyRef _only_ seems to be the standard for credentials everywhere. I have removed other options (string, ConfigmapRef) for `adminToken`.  Ref: [Prometheus basicAuth](https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.BasicAuth). This way it won't appear in the `GET` call as raw string even by accident.

svix-james:
I still think we should allow by raw string. Let's just make it clear in description that we don't recommend it and should only be used for testing.

ashayshub:
> I still think we should allow by raw string. Let's just make it clear in description that we don't recommend it and should only be used for testing.

Done.

infra/operator/src/crd.rs
```diff
@@ -218,6 +225,29 @@ pub(crate) enum Phase {
     Degraded,
 }
 
+/// Source for the admin token. Mirrors the Kubernetes EnvVar pattern:
+/// set either `value` for a plain string or `valueFrom` to pull from a Secret.
+/// Setting both is invalid.
+#[derive(Deserialize, Serialize, Clone, Debug, JsonSchema, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct AdminTokenSource {
```
svix-james: This should be an enum, not a struct. You get mutual exclusivity for free.

infra/svix-values.yaml
```diff
@@ -43,11 +43,13 @@ cluster:
       - name: COYOTE_OPENTELEMETRY_ADDRESS
         value: grpc://datadog-agent.datadog.svc.cluster.local:4317
       - name: COYOTE_ADMIN_TOKEN
```
svix-james: Why are we keeping this?

infra/operator/src/resources/statefulset.rs
```diff
@@ -221,15 +217,38 @@ fn build_env(
...
-        env.push(EnvVar {
-            name: extra.name.clone(),
-            value: extra.value.clone(),
-            ..Default::default()
-        });
+    env.extend(spec.env_var.iter().cloned());
+
+    // Admin token — pushed last so spec.admin_token always takes precedence over
+    // any COYOTE_ADMIN_TOKEN set via spec.env_var. If both are set, spec.env_var value is silently ignored.
+    if let Some(admin_token) = &spec.admin_token {
```
svix-james, line 230: This kind of complex logic is not needed if you use an enum.

infra/svix-values.yaml
ashayshub: To test if the env variable is correctly overridden by the adminToken attribute.

infra/operator/src/crd.rs
ashayshub: I had this as enum at first but Coyote suggested although it is rust idiomatic way to use an enum here the k8s openapi idiomatic way is to go with struct here. It would be easy for k8s programmers to read this according to it. Its argument is since "k8s_openapi::EnvVar itself uses the struct pattern" this closely follows it apart from other schema relataed issues with enum.

infra/svix-values.yaml
svix-james: Can we test that in Dev and not leave it here?

ashayshub: Removed.

infra/operator/src/crd.rs
svix-james: The code is worse because we're trying to represent this as a struct. We should change it.

ashayshub: I have used the enum for `AdminTokenSouce` now.

infra/operator/src/resources/statefulset.rs
ashayshub: Added adminTokenSource as enum now.

infra/operator/src/crd.rs
```diff
@@ -18,10 +20,6 @@ fn default_nodes() -> i32 {
     1
 }
 
-fn default_secret_key() -> String {
```
ashayshub, line 21: @svix-james: Do we need a default secret key function here? I have used the k8s openapi SecretKeySelector but that type does not accept a default function. Let me know. Removed it in order to use uniform k8s openapi types rather than building one.

ashayshub: Nvm, Kubernetes itself supports field defaults via the CRD schema using default in the OpenAPI spec. Added it using schema_with.

svix-james: yep, it's total garbage. Good to remove it.

infra/operator/src/crd.rs
```diff
@@ -272,3 +282,19 @@ fn tolerations_schema(_gen: &mut schemars::SchemaGenerator) -> schemars::Schema
 fn affinity_schema(_gen: &mut schemars::SchemaGenerator) -> schemars::Schema {
     schemars::json_schema!({ "type": "object", "x-kubernetes-preserve-unknown-fields": true })
 }
+
+fn secret_key_selector_schema(_gen: &mut schemars::SchemaGenerator) -> schemars::Schema {
```
svix-james: Do we need these schemas? I've already forgotten whether we can rely on defaults provided by the native types here.

ashayshub: The native types don't implement JsonSchema which is what these functions do. That helps us during validation and `k explain` queries for the CRD on these types.

infra/operator/src/crd.rs
ashayshub: Removed.

infra/operator/src/crd.rs
svix-james: the `schemars` feature in the k8s openapi crate can't do that for us?

ashayshub: Perfect, using the schemars from k8s openapi did it. Removed all the schema_with fn stuff.

svix-james:
Ok it looks like that goddam typo linter doesn't like the CRD spec. Maybe we can make that linter just ignore this particular file?

ashayshub:
> Ok it looks like that goddam typo linter doesn't like the CRD spec. Maybe we can make that linter just ignore this particular file?

Damn, that check was working on sub-word level (AND-ed)

svix-james (approved):

PR #723, opened 2026-03-30
Config: load default config from file at compile time

tasn:
This removes src/cfg/default.rs and replaces it with a top-level file called config.default.toml.

Advantages:
1. No need to manually implement (annoying) default implementations for all config values.
2. We can document this config file and use it as an example for us and others.
3. It's very easy to know what configs we have and how to override them.

Follow up to #717


svix-jplatte (commented):
I'm a bit wary of semantics-unaware TOML merging before deserialization. Right now, the only impact this has is that a config file that declares a filename for `persistent_db` or `ephemeral_db`, but not a `path`, is now valid when it wasn't before. Probably okay? But it could have unintended consequences down the line.

tasn:
> I'm a bit wary of semantics-unaware TOML merging before deserialization. Right now, the only impact this has is that a config file that declares a filename for persistent_db or ephemeral_db, but not a path, is now valid when it wasn't before. Probably okay? But it could have unintended consequences down the line.

Why would this be a problem? Sounds like it should be valid.

And anyway, we don't care about the config file or the default config files being fully valid, what we care about is the config being valid, after you've taken into account the default config, loaded config, and environment variables, no? And that's already validated?

config.default.toml
```diff
@@ -0,0 +1,32 @@
...
+seed_nodes = []
+auto_initialize = true
+discovery_timeout_ms = 30000
+startup_discovery_delay_ms = 10
+log_index_interval_ms = 600000
+log_sync_interval_commits = 0
+log_sync_interval_ms = 10000
+log_ack_immediately = true
+snapshot_after_ms = 900000
+shut_down_on_go_away = true
```
svix-jbrown: this is only for tests; I don't think it makes sense to expose in this file

config.default.toml
```diff
@@ -0,0 +1,32 @@
...
+log_format = "default"
+opentelemetry_metrics_use_http = false
+opentelemetry_metrics_period_seconds = 60
+opentelemetry_service_name = "coyote"
+
+[persistent_db]
+path = "./db"
+
+[ephemeral_db]
+path = "./db"
```
svix-jbrown, line 12: shouldn't these both have filenames?

config.default.toml
tasn: Sorry, was just the output of the new command you added. Will remove!

config.default.toml
tasn: It was just the output of the new command you added, and that's what it printed.

Though tbh I want to change this config anyway. No use letting people control the filename, it's too low-level and not useful.

tasn: Ah yeah, it's optional, and anyway feel low-level, do we really need it? I don't think it's useful for people.

config.default.toml
tasn: OK, I removed it but tests are failing. The reason is that the value is now not set, which makes sense, as this is what populates defaults.

I have a few ideas curious to hear what you think:
1. Have a section at the end of the config file that is for "internal" stuff that people shouldn't touch, and we can trim it before putting in the docs, (or just leave it there). The problem with this is that if we ever want to change defaults for these things it's going to be a pain, as customers may copy the example config file and start from that. I guess it's a problem with all values, including cluster configs.
2. For internal values, we can just have a default defined in code.
3. We can for our own convenience have a default config file (and I can move it to be under `src/cfg/default.toml`) and we can have a separate file we just copy-paste and edit (or automatically generate by commenting out all the configs so that they start commented out). <-- probably my favorite.
4. Give up on this PR. :(

What I'm trying to solve here is:
1. Make it easy for us to share a config file that's documented with customers.
2. Make it easy for us to know what configs there are and how to edit them - it was a big pain in the ass for me earlier.

svix-jplatte:
> Why would this be a problem? Sounds like it should be valid.

I'm thinking of 
- tables that are more like a map than a struct (i.e. allows arbitrary keys) - you couldn't unset all default keys for that at once (might be okay, might be a problem, depends on the situation)
- tagged union / struct enum type things, where there's a type field and associated fields that only apply for certain types - overriding the type should mean none of the associated fields from the default config are used imo, but that goes against the generic map-merge approach

config.default.toml
svix-jbrown: I just think that for things that are purely internal (which is to say, mostly just this one setting so far), we should keep a `serde` default and not include them in this file.

config.default.toml
svix-jbrown: I think the idea was that if the persistent and ephemeral DBs are on different volumes, you might want to do something like

```toml
[persistent_db]
path = "/data"
filename = "coyote_db"

[ephemeral_db]
path = "/instance-data"
filename = "coyote_db"
```

I don't have strong feelings about it, though

tasn: Yeah, I remember, though I think this achieves the same:
filename for persistent is always `persistent` and for ephemeral it's always `ephemeral`.
And then people just set the path to whatever the way (including the same).

svix-jbrown:
(this needs to be rebased and updated with some of the newly-added config options)

tasn:
Seems like people don't like it. :(

PR #735, opened 2026-03-30
Msgs: make retention optional, and rename values to more descriptive.

tasn:


crates/msgs/src/entities.rs
```diff
@@ -458,27 +458,8 @@ impl JsonSchema for ConsumerGroup {
...
-    }
-}
-
-pub fn default_retention_ms() -> DurationMs {
-    DurationMs::from_hours(24 * 30)
-}
-
-pub fn default_retention_bytes() -> NonZeroU64 {
-    NonZeroU64::new(1_000_000_000_000).expect("constant is non-zero")
+    pub period_ms: Option<DurationMs>,
```
svix-gabriel, line 463: I'm fine with this, but I just want to call out this goes against the original design you asked for of
```
{
   "retention": {
      "size": ...,
      "bytes": ...
   }
}

svix-gabriel (dismissed):

crates/msgs/src/entities.rs
tasn: Yeah, I made a mistake there. It just looks too opaque.

svix-gabriel (approved):

PR #736, opened 2026-03-30
Stop creating non-deterministic uuid-v7 inside raft 

svix-lucho:
This
```
Self::from_uuid(Uuid::new_v7(uuid::Timestamp::from_unix(
    uuid::NoContext,
    now.as_second() as u64,
    now.subsec_nanosecond() as u32,
)))
```
adds random bits to the generated ID, so we can't use it as-is inside raft operations. They have to be completely deterministic. 

This updates the `Id::new` method to receive the timestamp and 10 random bytes to generate the uuidv7 id, which makes generation deterministic. The random bytes are then generated at the API level, when the operations are created. 

This makes me think that maybe the random bytes should come from the raft layer, instead of having to add the random bytes to each operation. Not sure though. In any case, it should be easy to refactor this version to get the random bytes from raft. 

Helps with https://github.com/svix/coyote-private/issues/738

### DIFF
--- a/crates/coyote-id/src/lib.rs
+++ b/crates/coyote-id/src/lib.rs
@@ -13,16 +13,20 @@ use self::marker::{IdMarker, PublicIdMarker};
 pub use self::{module::Module, public::Public};
 
 const V7_RANDOM_BYTES_LEN: usize = 10;
-pub type UuidV7RandomBytes = [u8; V7_RANDOM_BYTES_LEN];
+
+#[derive(Serialize, Deserialize, Clone, Copy, Debug)]
+#[serde(transparent)]
+pub struct UuidV7RandomBytes([u8; V7_RANDOM_BYTES_LEN]);
+
+impl UuidV7RandomBytes {
+    pub fn new_random() -> Self {
+        Self(rand::random())
+    }
+}
 
 pub type AuthTokenId = Id<m::AuthToken>;
 pub type NamespaceId = Id<m::Namespace>;
 pub type TopicId = Id<m::Topic>;
-
-pub fn random_v7_bytes() -> UuidV7RandomBytes {
-    // FIXME: maybe this should come from the raft state? it could also guarantee sub millisecond ordering.
-    rand::random()
-}
 
 // Don't want these types to be nameable,
 // so they're defined in a private submodule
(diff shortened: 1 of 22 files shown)


crates/coyote-id/src/lib.rs
```diff
@@ -12,10 +12,18 @@ mod module;
...
 
+const V7_RANDOM_BYTES_LEN: usize = 10;
+pub type UuidV7RandomBytes = [u8; V7_RANDOM_BYTES_LEN];
+
 pub type AuthTokenId = Id<m::AuthToken>;
 pub type NamespaceId = Id<m::Namespace>;
 pub type TopicId = Id<m::Topic>;
 
+pub fn random_v7_bytes() -> UuidV7RandomBytes {
+    // FIXME: maybe this should come from the raft state? it could also guarantee sub millisecond ordering.
```
tasn: We could potentially pass a random seed in the raft context. Dunno if a good idea though.

crates/coyote-id/src/lib.rs
```diff
@@ -12,10 +12,18 @@ mod module;
 use self::marker::{IdMarker, PublicIdMarker};
 pub use self::{module::Module, public::Public};
 
+const V7_RANDOM_BYTES_LEN: usize = 10;
+pub type UuidV7RandomBytes = [u8; V7_RANDOM_BYTES_LEN];
```
svix-jbrown: I'm a little scared having a type alias like this; can we make this a newtype instead?

```rust
#[serde(transparent)]
pub struct UuidV7RandomBytes([u8; V7_RANDOM_BYTES_LEN])

impl UuidV7RandomBytes {
  fn new_random() -> Self {
    Self(rand::random())
  }
}
```

that way it's harder to accidentally cast one thing to another

crates/coyote-id/src/lib.rs
svix-jbrown: I don't think using deterministically-seeded RNGs is a good idea; too easy to either (a) use them in a cryptographically-sensitive place and break everything, or (b) call into them in a mildly-nondeterministic fashion and get them out of sync.

crates/coyote-id/src/lib.rs
svix-lucho: yeah. I don't know why I didn't do that

svix-lucho: Updated

svix-jbrown (approved):

PR #737, opened 2026-03-30
Optimize msgs.*.receive to avoid waiting longer than necessary for batch_wait

svix-gabriel:
Currently, if there aren't enough messages to satisfy a batch, the receive method simply sleeps for the batch_wait specified by the user. This is simple, but means we risk waiting longer than necessary. For long batch_wait times, that can be killer.

Instead of reading from fjall in a loop, this adds a central way to register `tokio::*::Notify`ers so that publish calls will "Wake up" the appropriate receive calls. There's some synchronization/mutex overhead, but overall much less overhead than we'd expect from reading from `fjall`.

The `TopicNotifier` abstraction we use here is shared by both the Axum handlers (AppState) _and_ accessible by the State in the raft layer, so we can call `TopicNotifier` from within the `publish` operation. This is needed so the logic works across all nodes, not just if reads and writes on the same node.

### DIFF
--- a/crates/msgs/src/lib.rs
+++ b/crates/msgs/src/lib.rs
@@ -15,6 +15,9 @@ pub mod entities;
 pub mod metrics;
 pub mod operations;
 pub(crate) mod tables;
+mod topic_publish_notifier;
+
+pub use topic_publish_notifier::*;
 
 pub const MSG_KEYSPACE: &str = "mod_msgs";
 pub const METADATA_KEYSPACE: &str = "mod_msgs_metadata";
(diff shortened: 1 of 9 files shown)


crates/msgs/src/lib.rs
```diff
@@ -110,13 +126,17 @@ pub fn estimate_available_stream_messages(
     topic: &TopicName,
     consumer_group: &ConsumerGroup,
     now: Timestamp,
-) -> Result<u64> {
+) -> Result<StreamEstimate> {
     let Some(topic_row) = TopicRow::fetch(metadata_tables, TopicRow::key_for(namespace_id, topic))?
     else {
-        return Ok(0);
+        return Ok(StreamEstimate {
```
svix-james: Maybe a Default impl instead?

svix-james (dismissed):

svix-james (approved):

Issue #716, opened 2026-03-30
Test Suite for the SDKs


svix-pluiz:
Create an automated test suite to check if the SDKs are working as expected

PR #741, opened 2026-03-30
SDK test suite

svix-pluiz:
Issue: https://github.com/svix/coyote-private/issues/716

- Add integration tests for all SDKs that run against a real coyote server
- New Github action `sdk-integration.yml` that builds the server image once then run each SDK tests in parallel
- Add a local test runner script `scripts/run-sdk-integration-tests.sh` that builds the docker image, starts the server, run selected SDKs and print a unified summary
- Delete obsolete `kv.rs` test


.github/workflows/sdk-integration.yml
```diff
@@ -0,0 +1,237 @@
...
+  build-server:
+    name: Build Server Image
+    runs-on: ubicloud-standard-4-ubuntu-2404
+    steps:
+      - uses: actions/checkout@v6
+      - name: Build Docker image
+        run: docker build -t coyote-server .
+      - name: Save Docker image
+        run: docker save coyote-server | gzip > /tmp/coyote-server.tar.gz
+      - uses: actions/upload-artifact@v7
```
svix-jbrown, line 22: please note that artifact uploading/downloading from ubicloud is incredibly slow (usually limited to around 10Mbps).

Some options:

 1. Run the build on regular github actions workers
 2. Store the build by pushing it to a GHCR repo rather than uploading it as an artifact

.github/workflows/sdk-integration.yml
```diff
@@ -0,0 +1,237 @@
...
+          retention-days: 1
+
+  test-rust:
+    name: Rust SDK
+    needs: build-server
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/download-artifact@v4
```
svix-jbrown, line 39: fyi current is v8

.github/workflows/sdk-integration.yml
```diff
@@ -0,0 +1,237 @@
...
+env:
+  COYOTE_TOKEN: admin_abcdefghijlmnopqrstuvwxyz012345
+
+jobs:
+  build-server:
+    name: Build Server Image
+    runs-on: ubicloud-standard-4-ubuntu-2404
+    steps:
+      - uses: actions/checkout@v6
+      - name: Build Docker image
```
svix-jbrown: is there an advantage to building a full docker image instead of just the binary?

.github/workflows/sdk-integration.yml
```diff
@@ -0,0 +1,237 @@
...
+        run: |
+          docker run -d --name coyote-test \
+            -p 8050:8050 \
+            -e COYOTE_LISTEN_ADDRESS=0.0.0.0:8050 \
+            -e COYOTE_ENVIRONMENT=dev \
+            -e COYOTE_CLUSTER_AUTO_INITIALIZE=true \
+            -e COYOTE_ADMIN_TOKEN=${{ env.COYOTE_TOKEN }} \
+            coyote-server
+          for _i in $(seq 1 60); do
+            if curl -sf http://localhost:8050/api/v1.health.ping > /dev/null 2>&1; then
```
svix-jbrown: this isn't actually sufficient to determine if it's ready

.github/workflows/sdk-integration.yml
```diff
@@ -0,0 +1,237 @@
...
+              echo "Server is ready"
+              exit 0
+            fi
+            sleep 1
+          done
+          echo "Server failed to start"
+          docker logs coyote-test
+          exit 1
```
svix-jbrown: the copy-pasta here is rough. if we continue down this route, we should make this a script in this repository (`/.github/launch-test-server` or something)

.github/workflows/sdk-integration.yml
```diff
@@ -0,0 +1,237 @@
...
+      - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: false
+      - name: Run Rust SDK integration tests
+        run: cargo test --package coyote-client -- --ignored
+      - name: Stop server
+        if: always()
+        run: docker stop coyote-test || true
+
+  test-python:
```
svix-jbrown, line 57: this one (and probably some of the others) should be matrix builds running all the supported versions of the interpreter

scripts/run-sdk-integration-tests.sh
```diff
@@ -0,0 +1,264 @@
...
+            local name="${BASH_REMATCH[1]}"
+            local result="${BASH_REMATCH[2]}"
+            if [[ "$result" == "ok" ]]; then
+                echo "PASS $name"
+            else
+                echo "FAIL $name"
+            fi
+        fi
+    done <<< "$output"
+}
```
svix-jbrown: please don't try to parse human-readable test output this way; this extremely fragile and I don't want to deal with accidentally giving ourselves bash injection because someone wrote a test with weird characters in the name.

Let's just look at the exit status and be done with it. If any of the runners do not set the exit status, let's fix that.

scripts/run-sdk-integration-tests.sh
```diff
@@ -0,0 +1,264 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```
svix-jbrown: overcomplicated; just use `SCRIPT_DIR="$(dirname "$0")"` for regular posix compliance

scripts/run-sdk-integration-tests.sh
```diff
@@ -0,0 +1,264 @@
...
+ROOT_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
```
svix-jbrown: ditto; `cd` in a subshell is dinky and non-posix

scripts/run-sdk-integration-tests.sh
```diff
@@ -0,0 +1,264 @@
...
+for sdk in "${SDKS[@]}"; do
+    results_file="$RESULTS_DIR/${sdk}.results"
+    if [[ ! -f "$results_file" ]]; then
+        continue
+    fi
+
+    content=$(<"$results_file")
+
+    if [[ "$content" == "ERROR" ]]; then
+        echo "  $sdk: ERROR (build or setup failed, no tests ran)"
```
svix-jbrown: errors go to stderr

```suggestion
        echo >&2 "  $sdk: ERROR (build or setup failed, no tests ran)"
```

scripts/run-sdk-integration-tests.sh
```diff
@@ -0,0 +1,264 @@
...
+        echo "$content" | grep "^FAIL " | while read -r _ name; do
+            echo "    - FAIL: $name"
+        done
+    fi
+
+    echo "$content" | grep "^PASS " | while read -r _ name; do
+        echo "    - PASS: $name"
+    done
+
+    echo ""
```
svix-jbrown: I think all of this is way overcomplicated. I wouldn't do any of this parsing; I would just show the output as-is.

z-clients/python/tests/test_integration.py
```diff
@@ -0,0 +1,135 @@
...
+    get_resp = client.kv.get(key, KvGetIn())
+    assert get_resp.value == value
+
+    # Delete
+    del_resp = client.kv.delete(key, KvDeleteIn())
+    assert del_resp.success is True
+
+    # Verify deleted
+    get_resp = client.kv.get(key, KvGetIn())
+    assert get_resp.value is None
```
svix-jbrown: these (and the rest of the file) should be using regular pytest fixtures and DI rather than free function calls
```suggestion

@pytest.fixture
def client() -> Coyote:
    return Coyote(TOKEN, CoyoteOptions(server_url=SERVER_URL))

@pytest.fixture
def async_client() -> CoyoteAsync:
    return CoyoteAsync(TOKEN, CoyoteOptions(server_url=SERVER_URL))


# --- Sync tests ---


def test_health_ping(client):
    client.health.ping()


def test_kv_set_get_delete(client):
    key = "python-sync-kv-key"
    value = b"python-sync-kv-value"

    # Set
    set_resp = client.kv.set(key, KvSetIn(value=value))
    assert set_resp.success is True

    # Get
    get_resp = client.kv.get(key, KvGetIn())
    assert get_resp.value == value

    # Delete
    del_resp = client.kv.delete(key, KvDeleteIn())
    assert del_resp.success is True

    # Verify deleted
    get_resp = client.kv.get(key, KvGetIn())
    assert get_resp.value is None
```

z-clients/rust/tests/integration.rs
```diff
@@ -0,0 +1,89 @@
...
+use coyote_client::{
+    CoyoteClient,
+    models::{CacheDeleteIn, CacheGetIn, CacheSetIn, KvDeleteIn, KvGetIn, KvSetIn},
+};
+
+const TOKEN: &str = "admin_abcdefghijlmnopqrstuvwxyz012345";
+
+fn client() -> CoyoteClient {
+    CoyoteClient::new(TOKEN.to_owned(), None)
+}
```
svix-jbrown, line 16: any reason that the token and bind address aren't being read from the environment? I don't love hardcoding them in here

z-clients/python/tests/test_integration.py
```diff
@@ -0,0 +1,135 @@
...
+    CacheDeleteIn,
+    CacheGetIn,
+    CacheSetIn,
+    KvDeleteIn,
+    KvGetIn,
+    KvSetIn,
+)
+
+TOKEN = "admin_abcdefghijlmnopqrstuvwxyz012345"
+SERVER_URL = "http://localhost:8050"
```
svix-jbrown: these should be read from the environment

z-clients/javascript/tests/integration.test.mjs
```diff
@@ -0,0 +1,65 @@
+import { describe, it } from "node:test";
+import assert from "node:assert/strict";
+import { Coyote } from "../dist/index.mjs";
+
+const TOKEN = "admin_abcdefghijlmnopqrstuvwxyz012345";
+const SERVER_URL = "http://localhost:8050";
```
svix-jbrown: these should be read from the environment

z-clients/java/src/test/java/com/svix/coyote/IntegrationTest.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.List;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+@Tag("integration")
+class IntegrationTest {
+
+    private static final String TOKEN = "admin_abcdefghijlmnopqrstuvwxyz012345";
+    private static final String SERVER_URL = "http://localhost:8050";
```
svix-jbrown: these should be read from the environment

z-clients/go/integration_test.go
```diff
@@ -0,0 +1,112 @@
...
+package coyote_test
+
+import (
+	"context"
+	"testing"
+
+	coyote "github.com/svix/coyote/z-clients/go"
+)
+
+const token = "admin_abcdefghijlmnopqrstuvwxyz012345"
```
svix-jbrown: this should be read from the environment

scripts/run-sdk-integration-tests.sh
```diff
@@ -0,0 +1,264 @@
+#!/usr/bin/env bash
```
svix-jbrown, line 1: I'm not sure the point of this file if you aren't going to call it from the github workflow. Just for people to run locally? We should strive to always use exactly the same code to invoke tests locally and in CI so that CI is appropriately debuggable.

.github/workflows/sdk-integration.yml
svix-pluiz: > is there an advantage to building a full docker image instead of just the binary?

I chose to go with the docker image because it's the same image we ship to prod, however, it would save a lot of time to just use the binary

.github/workflows/sdk-integration.yml
svix-pluiz: > please note that artifact uploading/downloading from ubicloud is incredibly slow (usually limited to around 10Mbps).

Oh, I didn't know that.

But if we go by the "just upload the binary" route this wouldn't be a problem, right?

.github/workflows/sdk-integration.yml
svix-pluiz: > this one (and probably some of the others) should be matrix builds running all the supported versions of the interpreter

Nice catch, by this way I'll definitely split this into different yml files for each test

scripts/run-sdk-integration-tests.sh
svix-pluiz: > please don't try to parse human-readable test output this way; this extremely fragile and I don't want to deal with accidentally giving ourselves bash injection because someone wrote a test with weird characters in the name.

That's fair, I think I can just remove this part

svix-pluiz:
@svix-jbrown This draft was the general idea for the suite, I'll update this draft soon to use environment variables, remove all bash hacky stuff that I wrote on the local tester and update the python test suite. Thanks for the review!!
I'll probably rewrite the `sdk-integration.yml` from scratch and split across different files for each lang, as we are gonna test multiple versions of the same lang

.github/workflows/sdk-integration.yml
svix-pluiz: Hey, could you tell me the appropriate way to know if the server is ready?

z-clients/rust/tests/integration.rs
```diff
@@ -0,0 +1,93 @@
...
+    let client = client();
+    let key = "rust-integration-cache-key".to_string();
+    let value = b"rust-integration-cache-value".to_vec();
+
+    // Set (ttl_ms = 60000)
+    client
+        .cache()
+        .set(
+            key.clone(),
+            CacheSetIn::new(value.clone(), Duration::new(60, 0)),
```
svix-jplatte:
```suggestion
            CacheSetIn::new(value.clone(), Duration::from_secs(60)),
```

.github/workflows/sdk-integration.yml
svix-jplatte: Also for simple SDK tests that would only do a handful of calls, I think we shouldn't do release builds, which take a lot longer than dev.

It unfortunately seems like our dev binaries are huge (128MB) though, and we already disabled debug-info. If James meant 10 mega-*bits* per second, that'd be around 100s for uploading.

(release binary is 56M right now, which is also bigger than I would have expected but makes sense in relation to dev)

z-clients/go/integration_test.go
```diff
@@ -0,0 +1,124 @@
+//go:build integration
```
svix-jplatte, line 1: Is this necessary? We don't do it in the Go SDK tests in `svix-webhooks`.

.github/workflows/sdk-integration.yml
```diff
@@ -0,0 +1,179 @@
...
+      - name: Start Coyote server
+        run: .github/scripts/start-coyote-server.sh
+        env:
+          COYOTE_ADMIN_TOKEN: ${{ env.COYOTE_TOKEN }}
+      - uses: actions/setup-go@v5
+        with:
+          go-version: ${{ matrix.go-version }}
+      - name: Run Go SDK integration tests
+        working-directory: z-clients/go
+        run: go test -tags=integration -v -count=1 ./...
```
svix-jplatte: What do these arguments do? The trailing `./...` looks escpecially weird.

.github/workflows/sdk-integration.yml
```diff
@@ -0,0 +1,179 @@
...
+
+  test-java:
+    name: Java SDK (${{ matrix.java-version }})
+    needs: build-server
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    strategy:
+      fail-fast: false
+      matrix:
+        java-version: ['17', '21']
```
svix-jplatte: We support down to Java 8 (unfortunately still far from end-of-life), I'd say we should test it.

z-clients/python/tests/test_integration.py
svix-pluiz: https://github.com/svix/coyote-private/pull/741/commits/45b46fd683fd1f636048603f514c810e9f5a012d

z-clients/javascript/tests/integration.test.mjs
svix-pluiz: https://github.com/svix/coyote-private/pull/741/commits/45b46fd683fd1f636048603f514c810e9f5a012d

z-clients/java/src/test/java/com/svix/coyote/IntegrationTest.java
svix-pluiz: https://github.com/svix/coyote-private/pull/741/commits/45b46fd683fd1f636048603f514c810e9f5a012d

scripts/run-sdk-integration-tests.sh
svix-pluiz: https://github.com/svix/coyote-private/pull/741/commits/a5c3ecfdd846db2562926d8c87b1b1c86df58c9c

scripts/run-sdk-integration-tests.sh
svix-pluiz: https://github.com/svix/coyote-private/pull/741/commits/a5c3ecfdd846db2562926d8c87b1b1c86df58c9c

z-clients/python/tests/test_integration.py
svix-pluiz: https://github.com/svix/coyote-private/pull/741/commits/336d4509f00ff7ff21ed857158af09a4de28b09a

scripts/run-sdk-integration-tests.sh
svix-pluiz: >  Just for people to run locally? 

Yes

> We should strive to always use exactly the same code to invoke tests locally and in CI so that CI is appropriately debuggable.

I'll update it to be the 'source of truth' of the CI

z-clients/go/integration_test.go
svix-pluiz: I read that it's common to separate 'slow' tests in go using tags [here](https://mickey.dev/posts/go-build-tags-testing/)

.github/workflows/sdk-integration.yml
svix-pluiz: `./...` is specific for Go, it means "run all tests files from all subdirectories, you can read more here https://pkg.go.dev/cmd/go#hdr-Package_lists_and_patterns

.github/workflows/sdk-integration.yml
svix-pluiz: > We support down to Java 8 

image

svix-pluiz: No problem, going to add as well

svix-pluiz: added here
https://github.com/svix/coyote-private/pull/741/commits/b120acde89cdd4a3f7490f1ab1a25c6b52b41a07

.github/workflows/sdk-integration.yml
```diff
@@ -0,0 +1,160 @@
...
+        go-version: ['1.22', '1.23']
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/download-artifact@v4
+        with:
+          name: coyote-server-binary
+          path: /tmp
+      - name: Start Coyote server
+        run: scripts/start-coyote-server.sh /tmp/coyote-server
+      - uses: actions/setup-go@v5
```
svix-jplatte: TIL there's a first-party Go setup action.

.github/workflows/ci.yml
```diff
@@ -141,6 +141,9 @@ jobs:
           else
             echo 'all good.'
           fi
+  sdk-integration:
+    uses: ./.github/workflows/sdk-integration.yml
+    secrets: inherit
```
svix-jplatte: Do we actually need this?

scripts/start-coyote-server.sh
```diff
@@ -0,0 +1,52 @@
...
+    if curl -sf "http://localhost:$PORT/api/v1.health.ping" > /dev/null 2>&1; then
+        echo "Server is ready (pid $(cat "$PID_FILE"))"
+        exit 0
+    fi
+    if [[ $i -eq 60 ]]; then
+        echo "Server failed to start"
+        exit 1
+    fi
+    sleep 1
+done
```
svix-jplatte:
```suggestion
done

echo "Server failed to start"
```

.github/workflows/sdk-integration.yml
```diff
@@ -0,0 +1,160 @@
...
+          name: coyote-server-binary
+          path: /tmp
+      - name: Start Coyote server
+        run: scripts/start-coyote-server.sh /tmp/coyote-server
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ matrix.rust }}
+      - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: false
```
svix-jplatte: This is equivalent to just not using the action at all. Why did you put `save-if: false`?

.github/workflows/sdk-integration.yml
svix-jplatte: wtf :laughing:

.github/workflows/ci.yml
svix-pluiz: > Do we actually need this?

We could either put sdk-integration individually in `on-pr.yml` and `on-push.yml` or put this under `ci.yml`.
I put under `ci.yml` because it's 'semantically' part of the CI process

svix-jplatte: Sorry, I specifically meant `secrets: inherit`

.github/workflows/sdk-integration.yml
svix-pluiz: This actually means it's only restoring from the cache, but not saving it. I think it's fine to just get the cache of compiled dependencies from the main branch without polluting the cache with builds from integration test runs

svix-jplatte: The cache key includes the job name I'm pretty sure, so it will never find anything. Also this might be the only job that builds all of the SDK's dependencies. I think it should use the same `save-if` as other jobs. It shouldn't take much cache space (restoring a cache entry for all of the server's dependencies could be pretty costly, actually, if that was how this worked).

.github/workflows/ci.yml
svix-pluiz: > Sorry, I specifically meant secrets: inherit

Oh, I wasn't sure if this would need inherited secrets. We can test this without inheriting to check if it works

.github/workflows/sdk-integration.yml
svix-jplatte: From Slack: since this file hasn't gotten huge with the matrix stuff, so leaving it as a single workflow (https://svixhq.slack.com/archives/C0A72988962/p1775127109848229?thread_ts=1775122485.930729&cid=C0A72988962).

.github/workflows/sdk-integration.yml
svix-pluiz: > The cache key includes the job name I'm pretty sure, so it will never find anything. Also this might be the only job that builds all of the SDK's dependencies. I think it should use the same save-if as other jobs. 

Oh, if that's the case definitely it makes sense to also write it, going to update in a moment

svix-pluiz: done https://github.com/svix/coyote-private/pull/741/commits/0c4722ccb27b9d6105741b0e312b734cbd74fc1f

svix-jplatte (approved):
Please make sure to squash when merging, there are so many random fix commits here ^^

PR #742, opened 2026-03-30
Enable NLB zonal affinity

svix-james:
I'm not positive this will work, but it's supposed to enable zone affinity, which means internal DNS lookups of the main NLB address should return an IP in the same AZ as the requester.


svix-jbrown (approved):
worth experimenting with

PR #744, opened 2026-03-30
rust: fix TTLs

svix-jbrown:
The Rust client library has been broken since #722 because it is passing the wrong type for TTLs. 

Before this change, the `cache.set` benchmark crashed with a msgpack decode error, and now it does not. 🎉


svix-jplatte (approved):
Wow I'm such an idiot. We need a lint for unreferenced Rust source files 🥲

PR #748, opened 2026-03-31
Add partition `end_offset` metrics

svix-james:


crates/msgs/src/lib.rs
```diff
@@ -153,14 +153,15 @@ impl AllNodesWorker {
...
-        .await
-        {
-            Ok(Ok(())) => {}
-            Ok(Err(err)) => tracing::warn!(?err, "failed to collect stream lag metrics"),
-            Err(err) => tracing::warn!(?err, "stream lag metrics task panicked"),
+
+        let mut tasks = tokio::task::JoinSet::new();
+        tasks.spawn_blocking(move || record_topic_lag_metrics(&state));
+        let state = self.state.clone();
+        tasks.spawn_blocking(move || record_end_offsets(&state));
```
svix-jplatte, line 160: This is often written as:

```suggestion
        tasks.spawn_blocking({
            let state = self.state.clone();
            move || record_end_offsets(&state)
        });
```

which I think makes the intent of the clone a little clearer. Though if you accept this suggestion, please do it for both `spawn_blocking` calls (I couldn't do a larger suggestion for whatever reason).

crates/msgs/src/metrics.rs
```diff
@@ -190,6 +196,19 @@ impl MsgMetrics {
...
 
+    pub(crate) fn record_topic_end_offset(
+        &self,
+        topic: &TopicName,
+        partition: Partition,
+        end_offset: u64,
+    ) {
+        let attrs = &[
+            topic.into(),
+            opentelemetry::KeyValue::new("partition", partition.get() as i64),
```
svix-jplatte, line 207: Since we have a nice strongly-typed `Partition` here, we can use the same `impl From<Thing> for opentelemetry::KeyValue` as elsewhere here, right?

svix-jplatte (approved):
Feedback is non-blocking.

Issue #156, opened 2026-02-19
Make codegen use nice types for durations


svix-jplatte:
Once #155 is done, it shouldn't be a lot of work to incorporate it into codegen, though it probably needs updated in both openapi-codegen (update the type mapping) and templates (adding imports in languages where fully-qualified syntax without imports is not available).

tasn:
1. We need to make sure durations are correctly marked throughout the API (all the modules).
2. We are going to probably have `_millis` suffix in the API, so we would need to trim that for SDKs that support durations (and internally translate between the two).

svix-jplatte:
Oh, I thought seeing previous discussions where you said you don't want a suffix like `_millis`. If we consistently do `_millis`, we can also just have this as a template hack - strip `_millis` from field names (but not serialized names of course) and substitute the type in the template. This would be less clean than doing it as I was thinking originally (w/ #155), but we don't have any infrastructure for field name rewriting in `openapi-codegen` so doing it "properly" is likely going to be a fair bit harder.

svix-jplatte:
Last piece of this was done in #923.

PR #751, opened 2026-03-31
Some time-duration improvements

svix-jplatte:
I went to review whether all the durations were `_ms`-suffixed in serialized form for https://github.com/svix/coyote-private/issues/156, and found that they already were, but there were a couple other improvements to be made.

Last commit is a **breaking change** for raft because the rate-limit's `LimitResponseData` field `retry_after` which I think is serialized as an object is removed, and instead we get `retry_after_ms` which is just an integer. It's an optional property though, so I don't think we have to kill any instances over this.

### DIFF
--- a/crates/coyote-core/src/types/duration_ms.rs
+++ b/crates/coyote-core/src/types/duration_ms.rs
@@ -1,7 +1,7 @@
 use std::{
     borrow::Cow,
     fmt,
-    ops::{Add, AddAssign},
+    ops::{Add, AddAssign, Mul},
     time::Duration,
 };
 
@@ -68,6 +68,12 @@ impl From<u64> for DurationMs {
     }
 }
 
+impl From<DurationMs> for jiff::TimestampArithmetic {
+    fn from(value: DurationMs) -> Self {
+        Duration::from(value).into()
+    }
+}
+
 impl fmt::Debug for DurationMs {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         self.0.fmt(f)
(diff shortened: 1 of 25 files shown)


crates/coyote-core/src/types/duration_ms.rs
```diff
@@ -90,6 +96,22 @@ impl AddAssign<DurationMs> for jiff::Timestamp {
...
+    }
+}
+
+impl Mul<u64> for DurationMs {
+    type Output = DurationMs;
+
+    fn mul(self, rhs: u64) -> DurationMs {
+        DurationMs(self.0 * rhs)
+    }
+}
```
svix-jbrown, line 113: do you think we should use `.saturating_mul` here instead to prevent weird behavior for large values?

crates/coyote-core/src/types/duration_ms.rs
```diff
@@ -90,6 +96,22 @@ impl AddAssign<DurationMs> for jiff::Timestamp {
     }
 }
```
svix-jbrown, line 97: similarly, I feel like these should use saturating arithmetic

openapi.json
```diff
@@ -6295,7 +6294,6 @@
             "description": "Milliseconds until enough tokens are available (only present when allowed is false)",
             "type": "integer",
             "format": "uint64",
-            "minimum": 0,
```
svix-jbrown, line 7275: why did we lose all of the `minimum: 0` values here? I guess it's implied by `uint64`, although I think it's probably *also* worth setting `minimum` and `maximum` for languages that ignore the format.

crates/coyote-core/src/types/duration_ms.rs
svix-jplatte: Sure, sounds like a good idea.

openapi.json
svix-jplatte: Yeah it's implied. Also our codegen doesn't read `minimum` and `maximum`. But I'll add it back, it doesn't hurt.

crates/coyote-core/src/types/duration_ms.rs
svix-jplatte: jiff's `saturating_add` is fallible...

svix-jbrown: wat

svix-jbrown (approved):

svix-jbrown:
very perplexed how `saturating_add` can be fallible but whatevs

PR #755, opened 2026-03-31
feat(gha): add `COYOTE_ADMIN_TOKEN` as k8s secret

ashayshub:
This PR adds  `COYOTE_ADMIN_TOKEN` as a k8s secret

The secret is fed to the `adminToken` in the coyoteclusterspec CR
which then sets the  `COYOTE_ADMIN_TOKEN`  env var.


ashayshub:
No longer required.

PR #759, opened 2026-03-31
Assign access rules to all routes

### LINKED ISSUE #683: Add access rules for all routes
Initially, I only added access rules for the "main" modules, not the namespace management and admin APIs. Those are currently represented in the authorization code to be admin-only, but they should have their own rules like everything else. This is a backwards-compatible change.

svix-jplatte:
Part of #68.
Closes #683.

### DIFF
--- a/crates/coyote-id/src/module.rs
+++ b/crates/coyote-id/src/module.rs
@@ -9,11 +9,11 @@ pub enum Module {
     Msgs = 5,
     AuthToken = 6,
 
-    AdminCluster = 192,
-    AdminNamespace = 193,
-    AdminAuthToken = 194,
-    AdminRole = 195,
-    AdminAccessPolicy = 196,
+    AdminCluster = 200,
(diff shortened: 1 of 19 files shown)


crates/coyote-id/src/module.rs
```diff
@@ -8,6 +8,12 @@ pub enum Module {
     RateLimit = 4,
     Msgs = 5,
     AuthToken = 6,
+
+    AdminCluster = 192,
+    AdminNamespace = 193,
+    AdminAuthToken = 194,
+    AdminRole = 195,
+    AdminAccessPolicy = 196,
```
tasn: Why 192? Maybe at least do 200?

tasn (approved):

svix-jbrown (commented):
can we make something that dumps all referenced resources and actions to stdout so we can easily see what they all are? it's hard to review in the macro form

PR #763, opened 2026-03-31
CI: Kick off lambda tests after deploy

svix-james:
Kick off staging e2e tests after coyote deploy


svix-jbrown (commented):
for my edification, which PR added  permissions for this IAM user to invoke this lambda? I couldn't find it in monorepo-private

svix-james:
> for my edification, which PR added permissions for this IAM user to invoke this lambda? I couldn't find it in monorepo-private

Updated -- we already have AWS env vars set up in the repo right now for staging/dev envs. Updated the step to use those, which should have permissions. If not I'll manually add them later.

svix-jbrown (approved):

PR #765, opened 2026-03-31
cluster: make default config crash-safe

### LINKED ISSUE #732: Change default fsync behavior to fsync synchronously
Need to check the exact implications.

svix-jbrown:
This changes the default config (and the benchmarking config) to be crash-safe by refusing to acknowledge until data has been synced out, which by default happens every 10ms.

This subtantially increases latency (making the mean latency for writes 5ms, for obvious reasons), but means that the log is guaranteed to be flushed to persistent storage if a crash occurs.

The throughput impact is less-stark; for very-low concurrency benchmarking (e.g., 1 writer) it's obviously awful and drops throughput to ~100 ops; however, there's essentially no impact for concurrent benchmarks (e.g., at 500 concurrent clients, `msgs.stream.receive` goes from 10936 ops to 9093 ops (about a 15% decrease) and `msgs.queue.receive (batch=1) - receive` goes from 9484 ops to 11032 ops -- mostly just showing that there's a ton of variance between otherwise-identical test runs!

Fixes #732

### DIFF
--- a/config.bench.toml
+++ b/config.bench.toml
@@ -10,5 +10,5 @@ opentelemetry_sample_ratio=0.1
 
 [cluster]
 log_sync_interval_commits = 0
-log_sync_interval_ms = 100
-log_ack_immediately = true
+log_sync_interval_ms = 10
+log_ack_immediately = false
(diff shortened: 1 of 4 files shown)


svix-james:
bruh you tellin me our benchmarks gonna get even fuckin worse? 

I mean, I know you actually said that benchmarks are not super-affected, but 5ms min latency is scary.

svix-james (approved):

PR #770, opened 2026-03-31
cluster: manually persist journal once per batch

svix-jbrown:
The main change in here is to change it so that instead of writing the journal on every operation, we do it at most once per batch.

This also adds the option (disabled by default) to fsync the primary data, which we've never supported before.

Branched off of and depends on #769.

### DIFF
--- a/config.bench.toml
+++ b/config.bench.toml
@@ -1,3 +1,5 @@
+# This is a configuration used for benchmarking, not for general use
+
 listen_address = "0.0.0.0:8050"
 log_level="info"
 environment = "dev"
@@ -6,9 +8,13 @@ opentelemetry_metrics_use_http=true
 opentelemetry_service_name="coyote"
 opentelemetry_address="http://localhost:4317"
 opentelemetry_metrics_address="http://localhost:9090/api/v1/otlp/v1/metrics"
-opentelemetry_sample_ratio=0.1
+opentelemetry_sample_ratio=0.001
+# for the primary data, write to the OS but do not fsync
+sync_mode = "buffer"
 
 [cluster]
+# fsync very infrequently
 log_sync_interval_commits = 0
-log_sync_interval_ms = 10
+log_sync_interval_ms = 10000
+# acknowledge commits before we fsync
 log_ack_immediately = false
--- a/config.bench.toml
+++ b/config.bench.toml
@@ -17,4 +17,4 @@ sync_mode = "buffer"
 log_sync_interval_commits = 0
 log_sync_interval_ms = 10000
 # acknowledge commits before we fsync
-log_ack_immediately = false
+log_ack_immediately = true
--- a/src/cfg/mod.rs
+++ b/src/cfg/mod.rs
@@ -101,6 +101,15 @@ impl Serialize for PeerAddr {
     }
 }
 
+impl From<SyncMode> for fjall::PersistMode {
+    fn from(value: SyncMode) -> Self {
+        match value {
+            SyncMode::Buffer => fjall::PersistMode::Buffer,
+            SyncMode::Sync => fjall::PersistMode::SyncAll,
+        }
+    }
+}
+
 #[derive(Clone, Debug, Default, Deserialize, EnvOverridable, Serialize)]
 pub struct DatabaseConfig {
     /// Directory in which this database is stored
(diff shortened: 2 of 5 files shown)


src/cfg/mod.rs
```diff
@@ -682,15 +710,23 @@ impl fmt::Display for Environment {
...
-            Self::Trace => Level::TRACE,
-        }
-        .fmt(f)
+        Level::from(*self).fmt(f)
     }
 }
 
+/// How data is synchronized to the underlying database
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, Default)]
+#[serde(rename_all = "lowercase")]
```
tasn: Should this be kebab? I know it doesn't matter at the moment, but would matter for the next value we put in if it's multi-word?

src/core/cluster/state_machine.rs
```diff
@@ -496,6 +499,16 @@ impl Store {
         if changed_log_id || changed_membership {
             self.record_ids_().await.context("recording updated IDs")?;
         }
+        if num_normal_entries > 0 {
+            // TODO: this should handle ephemeral if we ever bring ephemeral back.
```
tasn: I believe we use ephemeral for rate-limit and cache? So we need to handle that too.

config.bench.toml
```diff
@@ -6,9 +8,13 @@ opentelemetry_metrics_use_http=true
...
+opentelemetry_sample_ratio=0.001
+# for the primary data, write to the OS but do not fsync
+sync_mode = "buffer"
 
 [cluster]
+# fsync very infrequently
 log_sync_interval_commits = 0
-log_sync_interval_ms = 10
-log_ack_immediately = false
+log_sync_interval_ms = 10000
```
tasn, line 18: I think we should benchmark the default values, and not override to benchmark friendly values. What do you think?

src/core/cluster/state_machine.rs
```diff
@@ -455,6 +455,7 @@ impl Store {
         let mut changed_log_id = false;
```
tasn, line 455: Making a comment here to have a discussion thread: I don't understand how this change achieves what the description says. As in, I don't see anywhere where we changed how we persist log_sync?

src/cfg/mod.rs
svix-jbrown: ok

src/core/cluster/state_machine.rs
svix-jbrown: this PR has nothing to do with logging. It changes the fjall configuration so that the fjall journal isn't peristed on every `commit()` call and is instead persisted once per bulk-`apply` call.

We disable automatic-persistence-on-commit by calling `.manual_journal_persist(true)` in the DatabaseBuilder call and then we added the new `.persist()` call in the state machine applier.

config.bench.toml
svix-jbrown: I think that's true for general benchmarks, but for day-to-day testing it's a lot more useful if you don't have to fight the entire running system to get fsync time for a benchmark.

src/core/cluster/state_machine.rs
svix-jbrown: ok

src/core/cluster/handle.rs
```diff
@@ -57,6 +57,24 @@ pub enum Request {
...
+                | Self::Msgs(_)
+                | Self::AuthToken(_)
+                | Self::AdminAuth(_)
+        )
+    }
+
+    pub fn affects_ephemeral(&self) -> bool {
+        matches!(self, Self::Cache(_) | Self::RateLimit(_))
+    }
+}
```
tasn, line 81:
```suggestion
    pub fn is_persistent(&self) -> bool {
        match self {
            Self::ClusterInternal(_)
                | Self::Kv(_)
                | Self::Idempotency(_)
                | Self::Msgs(_)
                | Self::AuthToken(_)
                | Self::AdminAuth(_) => true
            Self::Cache(_) | Self::RateLimit(_)) => false
        }
    }
```
So that we know we get full coverage here (compiler error) when a new module is added.

The code that calls it can be changed to always write to one or the other.

If the problem is that this has more than just modules (e.g. internal cluster stuff), then we should probably make it return an enum or something, dunno. Though we want the full coverage to get the compiler error for sure.

src/core/cluster/state_machine.rs
tasn: Ah, journal being fjall journal, not the raft log. Ignore me, it's been a long day. :)

src/core/cluster/handle.rs
svix-jbrown: let me know what you think of the latest commit

tasn (approved):

PR #775, opened 2026-04-01
authorization: Enable checking of access token permissions

### LINKED ISSUE #68: Add ACL for the modules

svix-jplatte:
Closes https://github.com/svix/coyote-private/issues/68.
There's still follow-up work left to do before the release, but I'd rather have clearly-defined follow-up tickets than stuffing ever more stuff into that one.

### DIFF
--- a/crates/coyote-authorization/src/lib.rs
+++ b/crates/coyote-authorization/src/lib.rs
@@ -90,6 +90,32 @@ pub struct AccessRule {
 }
 
 impl AccessRule {
+    pub fn admin_rules() -> Arc<[Self]> {
+        static RULES: OnceLock<Arc<[AccessRule]>> = OnceLock::new();
+        RULES
+            .get_or_init(|| {
+                [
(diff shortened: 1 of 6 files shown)


crates/coyote-authorization/src/lib.rs
```diff
@@ -90,6 +90,32 @@ pub struct AccessRule {
 }
 
 impl AccessRule {
+    pub fn admin_rules() -> Arc<[Self]> {
```
tasn, line 93: I know I talked about having generic rules for operator/admin, but we can also just have the ACL check check if role == admin, allow.

tasn (approved):

crates/coyote-authorization/src/lib.rs
svix-jplatte: Makes sense to me. But not before making sure people can't create a normal role with the same name ^^

PR #787, opened 2026-04-01
operator: configure a startup probe for the server

svix-jbrown:
60 seconds is not necessarily enough time to stream over GB of data.

### DIFF
--- a/infra/operator/src/resources/statefulset.rs
+++ b/infra/operator/src/resources/statefulset.rs
@@ -281,7 +281,7 @@ fn build_container(
             }),
             initial_delay_seconds: Some(5),
             period_seconds: Some(10),
-            failure_threshold: Some(6),
+            failure_threshold: Some(2),
             success_threshold: Some(1),
             ..Default::default()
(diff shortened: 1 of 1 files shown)


infra/operator/src/resources/statefulset.rs
```diff
@@ -291,9 +291,21 @@ fn build_container(
...
+        startup_probe: Some(Probe {
+            http_get: Some(HTTPGetAction {
+                path: Some(CLUSTER_HEALTH_ENDPOINT.into()),
+                port: IntOrString::Int(cluster_port as _),
+                ..Default::default()
+            }),
+            initial_delay_seconds: Some(15),
             period_seconds: Some(10),
-            failure_threshold: Some(6),
+            failure_threshold: Some(120), // TODO: this should come from the helm chart
```
svix-james, line 308: Note that CI will "fail" if it takes this long. Not sure there's anything to do there for now, but worth noting.

svix-james (approved):

PR #789, opened 2026-04-01
Operator: Add default topology constraints

svix-james:
It seems like a reasonable default to try and spread pods across zones, so add this as the default spread constraint. User can override/add to them if needed.

Note: Also add the `PreferSameZone` annotation to the service, which won't have an effect on pre-1.35 versions of k8s, but should prefer to route requests to service endpoints in same-AZ in later versions.


svix-jbrown (dismissed):

svix-james:
Had to rebase.

svix-jbrown (approved):

PR #792, opened 2026-04-01
db: add some more persistence calls

svix-jbrown:


src/core/cluster/serialized_state_machine.rs
```diff
@@ -232,6 +232,7 @@ pub(crate) fn load_from_file<F: Read + Seek>(dbs: &Databases, f: &mut F) -> anyh
             );
             let keyspace = db.keyspace(&keyspace_name, KeyspaceCreateOptions::default)?;
             deserialize_keyspace(&mut z, &keyspace, chunks)?;
+            db.persist(fjall::PersistMode::SyncAll)?;
```
svix-james, line 235: When does this happen otherwise?

svix-james (approved):

src/core/cluster/serialized_state_machine.rs
svix-jbrown: the next time the keyspace gets written to by a regular log

PR #797, opened 2026-04-02
KV: add OCC to the delete operation.

tasn:


crates/kv/src/kvcontroller.rs
```diff
@@ -245,13 +245,25 @@ impl KvController {
...
+                let current_version = current.as_ref().map(|m| m.version).unwrap_or(0);
+                if current_version != expected {
+                    return Err(Error::bad_request("version_mismatch", "version mismatch"));
+                }
+            }
+
+            let db = db.lock("kvcontroller::delete");
 
             if let Some(data) = KvPairRow::fetch(&keyspace, KvPairRow::key_for(namespace_id, key))?
             {
```
svix-jbrown: this doesn't seem quite right; we're fetching the row twice? once on line 256 and once on line 268

tasn: Sorry, fixed.

svix-jbrown (approved):

PR #798, opened 2026-04-02
Auth token: remove kv-separation in storage.

tasn:
It was a bad copy-paste. We don't actually need that because values are always small. So this is actually the opposite of what we need.


svix-jbrown (approved):
this will conflict with #794 but whatevs

PR #799, opened 2026-04-02
cluster: improve background ticks to only happen when idle

svix-jbrown:


svix-jplatte (dismissed):
Definitely an improvement, but why make the interval bigger than the threshold?

The loop should be super cheap when it doesn't meet the threshold condition, so why not make time-from-last-op-to-tick more consistent by lowering the loop period to somewhere between 5ms / 50ms (and increasing the threshold if consistent 200-250ms ticks are too noisy)?

svix-jplatte (approved):

PR #800, opened 2026-04-02
cluster: make replication_lag_threshold configurable

svix-jbrown:


src/cfg/mod.rs
```diff
@@ -394,6 +394,16 @@ pub struct ClusterConfiguration {
     #[serde(default = "defaults::default_true")]
     #[env_overridable(skip)]
     pub shut_down_on_go_away: bool,
+
+    /// How far behind must the current node be to be considered "lagging" and eligible for
+    /// re-snapshotting?
```
svix-jplatte: I first thought "how far behind" was meant to be a duration, should probably clarify that here with whatever is correct. Probably something like "in number of raft operations"?

svix-jplatte (approved):

PR #804, opened 2026-04-02
codegen/python: More cleanup

svix-jplatte:
Ahead of the duration change (https://github.com/svix/coyote-private/issues/156).


svix-jbrown (approved):
much nicer. appreciate using the >=py3.10 optional syntax

PR #813, opened 2026-04-02
feat(operator): inject labelselector when none provided by user

ashayshub:
This PR injects labelselectors for `podaffinity` and `topologySpreadConstraint`

When user does not provide any `labelSelectors` for 
`podAffinity` and `topologySpreadConstraint` this code
injects the default values to the `podSpec`. By design, this PR 
does not make any default decisions on behalf of the user 
about the `podaffinity` or `topolgySpreadConstraint`.

### DIFF
--- a/infra/operator/src/resources/statefulset.rs
+++ b/infra/operator/src/resources/statefulset.rs
@@ -4,9 +4,10 @@ use k8s_openapi::{
     api::{
         apps::v1::{StatefulSet, StatefulSetSpec},
         core::v1::{
-            Container, ContainerPort, EnvVar, EnvVarSource, HTTPGetAction, ObjectFieldSelector,
-            PersistentVolumeClaim, PersistentVolumeClaimSpec, PodSecurityContext, PodSpec,
-            PodTemplateSpec, Probe, VolumeMount, VolumeResourceRequirements,
+            Affinity, Container, ContainerPort, EnvVar, EnvVarSource, HTTPGetAction,
+            ObjectFieldSelector, PersistentVolumeClaim, PersistentVolumeClaimSpec,
+            PodSecurityContext, PodSpec, PodTemplateSpec, Probe, Toleration,
+            TopologySpreadConstraint, VolumeMount, VolumeResourceRequirements,
         },
     },
     apimachinery::pkg::{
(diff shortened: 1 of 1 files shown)


infra/operator/src/crd.rs
```diff
@@ -23,15 +24,43 @@ fn default_nodes() -> i32 {
...
-        when_unsatisfiable: "ScheduleAnyway".into(),
+/// Hard: never place two CoyoteCluster pods on the same node.
+/// Soft: prefer pods in different zones (weight 100).
+///
+/// The operator automatically injects the cluster's pod labelSelector into
+/// each term at reconcile time, matching the same pattern as topology spread constraints.
+fn default_affinity() -> Option<Affinity> {
+    let label_selector = Some(LabelSelector {
+        match_labels: Some(BTreeMap::from([
+            ("app.kubernetes.io/name".into(), "coyote".into()),
```
svix-james: Can we populate this from the `label::selector` function so we're not duplicating these everywhere (not to mention hard-coding the name, which IIRC is user-configurable)

ashayshub: Done.

infra/operator/src/crd.rs
```diff
@@ -23,13 +23,30 @@ fn default_nodes() -> i32 {
...
-    vec![TopologySpreadConstraint {
-        max_skew: 1,
-        topology_key: "topology.kubernetes.io/zone".into(),
-        when_unsatisfiable: "ScheduleAnyway".into(),
+/// Hard: never place two CoyoteCluster pods on the same node.
+/// Soft: prefer pods in different zones (weight 100).
+///
+/// The operator automatically injects the cluster's pod labelSelector into
+/// each term at reconcile time, matching the same pattern as topology spread constraints.
+fn default_affinity() -> Option<Affinity> {
```
svix-james: So, I think we should probably remove the default here, but let's definitely keep the selector label injection.

ashayshub: Removed. Now, we are not making any default decisions on behalf of the users about either the `podAaffinity` or the `topologySpreadConstraint`. Only if the user does not add the labelselectors to his CR, we will add them on their behalf.

ashayshub:
No longer required.

PR #814, opened 2026-04-02
[draft 1] add idempotency to msgs publish

svix-lucho:
Using KvController inside msgs


crates/msgs/src/operations/publish.rs
```diff
@@ -45,23 +56,96 @@ impl PublishOperation {
...
+        log_index: u64,
+    ) -> Result<Option<PublishResponseData>> {
+        let Some(key) = self.idempotency_key.as_ref() else {
+            return Ok(None);
+        };
+
+        match state
+            .idempotency
+            .try_start(
+                self.namespace_id,
```
svix-lucho, line 108: wrong: this namespace should be internal, not the namespace of the msgs operation. Same for the other idempotency calls.

PR #816, opened 2026-04-03
Refactor Operator/CRD and add Helm README

svix-james:
While working on the README, I decided to try and simplify the configurable fields that the CRD exposes. A lot of the cluster-specific config params that were added initially don't need to be first-class items in the CRD for now, so I've removed them entirely. They can still be added as env vars if they need to be overridden.

Otherwise, I've tried to reorganize how the cluster spec is defined in the CRD to better isolate Coyote-specific configurations (currently in the `CoyoteSpec` struct) from more general k8s things like pod affinities and resource requests/limits.

The README has been updated with the relevant fields that the operator and CRD expose directly. I think this is a good start at something that's publicly shareable.

### DIFF
--- a/infra/helm-coyote/charts/crds/crds/coyoteclusters.json
+++ b/infra/helm-coyote/charts/crds/crds/coyoteclusters.json
@@ -56,7 +56,7 @@
                         ]
                       }
                     ],
-                    "description": "Admin token for privileged API access.\nSet either `value` for a plain string or `valueFrom` to pull from a Kubernetes Secret.\nUsing a plain string value is only recommended for testing. Use `valueFrom` in all other cases.",
+                    "description": "Admin token for privileged API access.\n\nSet either `value` for a plain string or `valueFrom` to pull from a Kubernetes Secret.\nUsing a plain string value is only recommended for testing. Use `valueFrom` in all other cases.",
                     "nullable": true,
                     "properties": {
                       "value": {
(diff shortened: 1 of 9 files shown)


infra/operator/src/crd.rs
```diff
@@ -156,43 +127,74 @@ pub(crate) struct VolumeSpec {
...
-    pub election_timeout_min_ms: Option<u64>,
+    pub log_format: Option<String>,
 
-    /// Maximum election timeout in milliseconds.
+    /// The OpenTelemetry address to send events to if given.
+    ///
+    /// Currently only GRPC exports are supported.
     #[serde(default)]
-    pub election_timeout_max_ms: Option<u64>,
+    pub opentelemetry_address: Option<String>,
```
svix-james, line 164: I'm honestly not clear yet whether the OTEL fields (or heck, even log-level/format) should be exposed as fields directly in the CRD, or if we should just leave it to users to discover the appropriate env vars from our code/docs and supply them themselves. 

For now, though, these params seem important enough to me that making them part of the CRD directly, where they can be more easily discovered, is appropriate. We can reverse this later if needed.

infra/helm-coyote/charts/crds/crds/coyoteclusters.json
```diff
@@ -1018,11 +986,26 @@
                   },
                   "nodes": {
                     "default": 1,
-                    "description": "Number of Coyote nodes. Should be odd for quorum (1, 3, 5...).",
+                    "description": "Number of Coyote nodes. Should be an odd number.",
```
tasn: I'd probably add: "recommended value is 3, or 1 if you want single node" (just better phrased).
I don't think we should recommend 5 or 7 or whatever.

infra/operator/src/resources/statefulset.rs
```diff
@@ -82,7 +82,7 @@ pub(crate) fn build(cluster: &CoyoteCluster, ns: &str) -> Result<StatefulSet> {
             ..Default::default()
         },
         spec: Some(StatefulSetSpec {
-            replicas: Some(spec.nodes),
+            replicas: Some(spec.coyote.nodes),
```
tasn, line 85: Should we call it `coyote` or just `server` (something generic)

The problem with using the project name, is that changing it is even more of a pain. :)

svix-james: This field is flattened anyway and doesn't affect the actual API, so I think it's fine for now. It will just be part of whatever larger PR we do when we rename. I don't think it's a big deal.

tasn (approved):

PR #824, opened 2026-04-03
Document access control.

tasn:


docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
+# Access Control
+
+Coyote offers granular access control to let you configure the access control exactly as you need it.
```
svix-james: "Access control" twice sounds weird here.

Maybe something like? 🤷 
```suggestion
Coyote offers highly (finely?) granular controls to let you configure access control exactly as you need it.
```

docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
...
+
+The access control functionality was designed to be easy to start with, but powerful for advanced users. See the [examples](#examples) section below, for advanced use-cases including frontend-friendly tokens and sharing tokens with untrusted 3rd parties.
```
svix-james:
```suggestion
The access control functionality was designed to be easy to start with, but powerful for advanced users. See the [Examples](#examples) section below for advanced use-cases including frontend-friendly tokens and sharing tokens with untrusted 3rd parties.
```

docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
...
+
+## Entities and hierarchy
+
+- [Role](#role): each action on Coyote is associated with a specific role.
+  - A role has a list of access control rules or policies attached to it.
+  - Roles can have additional context that can be used in path replacements and in conditionals (in the future when we support it).
+- [Policy](#policy): a policy is a reusable set of access rules. The same policy can be attached to multiple roles.
+- [Access rules](#access-rules): rules describe what operations are allowed or disallowed for a groups of resources.
```
svix-james:
```suggestion
- [Access rules](#access-rules): rules describe what operations are allowed or disallowed for a group of resources.
```

docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
...
+- [Auth tokens](#auth-tokens): tokens are how users authenticate with Coyote. Each token has a role associated with it, which is what the token authenticate to.
```
svix-james:
```suggestion
- [Auth tokens](#auth-tokens): tokens are how users authenticate with Coyote. Each token has a role associated with it, which is what the token authenticates to.
```

docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
...
+      "another-policy",
+    ],
+    "context": {
+        "tier": "enterprise",
+        "category": "my-category",
+    },
+}
+```
+
+Rules and resource identifiers are covered in the [access rules](#access-rules) section below.
```
svix-james:
```suggestion
Rules and resource identifiers are covered in the [Access Rules](#access-rules) section below.
```

docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
...
+
+The policies is a list of policies, which are explained in the next section.
```
svix-james:
```suggestion
`policies` is a list of policies, which are explained in the next section.
```

docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
...
+    ],
+}
+```
+
+
+### Access rules
+
+Access control rules are the specific rules that allow/deny access to resources or groups of resources.
+
+The way access control rules work is as follows: everything is denied by default, you can allow actions by adding an allow rule, and deny actions by adding a deny rule. The order of precedent is: `deny rule -> allow rule -> implicit deny all`, or in other words, if a deny rule matches, the request will be denied no matter if there's an allow before or after it.
```
svix-james:
```suggestion
The way access control rules work is as follows: everything is denied by default. You can allow actions by adding an allow rule and deny actions by adding a deny rule. The order of precedent is: `deny rule -> allow rule -> implicit deny all`, or in other words, if a deny rule matches, the request will be denied no matter if there's an allow before or after it.
```

docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
...
+
+Effect is either `allow` or `deny`.
+
+The resource field is the set of resources (may be more than one due to globbing) this rule should apply to. Main properties of a resource:
+- Resource names are structured like this: `module:namespace:key/subkey/subkey`. For example: `cache:store123:foo/bar`
+- The [namespace](./namespaces.mdx) part is optional, and when missing it matches to the default namespace. For example, `cache::foo/bar` means `foo/bar` in the default namespace. It does *NOT* mean wildcard matching.
+- Variable substitution is supported for both built-in variables (such as `${role}`) and context passed generic properties like `${context.my_variable}`.
+- The key part can have nesting using `/` which splits it to different sections.
+- Globbing is supported
+  - For the entire namespace section: `cache:*:foo/bar`
```
svix-james, line 154: Unless I'm misunderstanding, this sounds clearer to me:

```suggestion
  - For all namespaces: `cache:*:foo/bar`
```

docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
...
+  - For the entire module section: `*::foo/bar`
```
svix-james, line 155:
```suggestion
  - For all modules: `*::foo/bar`
```

docs/content/access-control.mdx
svix-james: Oh n/m

docs/content/access-control.mdx
svix-james: Ignore

docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
...
+```
+coyote admin auth_token create '{"name": "token-name", "role": "frontend-role", "context": {"user": "their-user-id"}}'
+# Returns:
+{
+    "id": "tok_123",
+    "token": "sk_1234",
+}
+```
+
+Alternatively, if JWT auth is also configured, people can just generate their own config, and pass the role as `role` (or maybe the correct JWT one for this), and context as a sub object called `context`.
```
svix-james:
```suggestion
Alternatively, if JWT auth is also configured, people can just generate their own config and pass the role as `role` (or maybe the correct JWT one for this) and context as a sub object called `context`.
```

docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
...
+
+
+## Admin auth token
+
+You can create roles, policies, and auth tokens using the Coyote API. However, you need an auth token in order to interact with the API.
+
+To solve that, Coyote lets you configure an admin auth token in the configuration file, which you can use to create policies, roles, and tokens. The admin auth token assumes the role `admin` which has access to everything on the cluster.
+
+To configure an admin auth token set the `COYOTE_ADMIN_TOKEN` environment variable to you wanted token, for example: `zpsNuhPMeMbg9q4cG2ZsJn`, and then you'll be able to use it for authenticating to that specific node. It's recommended that you only use the admin auth token for bootstrapping, and remove the configuration once done.
```
svix-james:
```suggestion
To configure an admin auth token set the `COYOTE_ADMIN_TOKEN` environment variable to the desired token, for example: `zpsNuhPMeMbg9q4cG2ZsJn`, and then you'll be able to use it for authenticating to that specific node. It's recommended that you only use the admin auth token for bootstrapping and remove the configuration once done.
```

docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
...
+
+Each token can have additional context associated with it which can be used for the policy variable substitution, observability and other use-cases.
+
+There are two ways to create tokens:
+- Creating them in Coyote using an API.
+- Using JWTs (coming soon).
+
+Creating a new token using the API looks something like this:
+```
+coyote admin auth_token create '{"name": "token-name", "role": "frontend-role", "context": {"user": "their-user-id"}}'
```
svix-james: Is it actually `auth_token` here?

docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
...
+                "resource": "cache::foo/*",
+                "actions": [
+                    "cache.get"
+                ],
+            },
+            {
+                "effect": "deny",
+                "resource": "cache::foo/bar",
+                "actions": [
+                    "cacheg.te"
```
svix-james:
```suggestion
                    "cache.get"
```

docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
...
+                "resource": "cache::foo/*",
+                "actions": [
+                    "cache.get"
+                ],
+            },
+            {
+                "effect": "deny",
+                "resource": "cache::foo/bar",
+                "actions": [
+                    "cacheg.get"
```
svix-james:
```suggestion
                    "cache.get"
```

docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
...
+                    "kv.get"
+                ],
+            },
+        ],
+    }'
+```
+
+### Configuring authentication tokens
+
+To create a new token attached to a specific role you can use the following API. You can also set an optional expiry if the token is meant to be temporary.
```
svix-james:
```suggestion
To create a new token attached to a specific role, you can use the following API. You can also set an optional expiry if the token is meant to be temporary.
```

docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
...
+
+Tokens also support a variety of other APIs including list, rotate, and expire. Please refer to the API reference docs for more information.
+
+## Examples
+
+Here are some example use-cases and how to achieve them.
+
+### Multi-tenant access to the same cluster
+
+First they would create a policy and a role (or we can have a default one, at least a policy) that looks something like this:
```
svix-james:
```suggestion
First you would create a policy and a role (or you can have a default one, at least a policy) that looks something like this:
```

docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
...
+    "id": "my-role",
+    "description": "The role-owned data.",
+    "policies": [
+      "cache-owned-data",
+      "kv-owned-reader",
+    ],
+}
+```
+
+Then any key with that role, would be limited to their own data, e.g. `cache::owned/my-role/*`
```
svix-james:
```suggestion
Then any key with that role would be limited to their own data, e.g. `cache::owned/my-role/*`
```

docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
...
+
+
+### Frontend friendly tokens
+
+You can create tokens that can be safely shared with the frontend (and thus with your users) by using the context and scoping the token to resources under their username.
+
+First they would create a role (or we can have a default one) that looks something like this:
```
svix-james: Same as above.... using both "they" and "we" here is confusing to me ... should it instead be "you"?

docs/content/access-control.mdx
```diff
@@ -0,0 +1,443 @@
...
+```
+
+Then they will only have access to `cache::frontend/their-user-id/*`
+
+
+### 3rd party access to cluster
+
+Similarly to the frontend friendly tokens above, you can also create tokens that are suitable for sharing with untrusted 3rd parties.
+
+For example, you can create the following role, to allow a trusted 3rd party to publish to a stream that's specific to them:
```
svix-james:
```suggestion
For example, you can create the following role to allow a trusted 3rd party to publish to a stream that's specific to them:
```

docs/content/access-control.mdx
tasn:
```suggestion
coyote admin auth-token create '{"name": "token-name", "role": "frontend-role", "context": {"user": "their-user-id"}}'
```
Whoops.

docs/content/access-control.mdx
tasn: Yes.

```suggestion
First you would create a role (or we can have a default one) that looks something like this:
```

svix-james (approved):

PR #825, opened 2026-04-03
Auth: implement support for JWT tokens.

### LINKED ISSUE #133: Support JWT authentication

tasn:
We verify exp, nbf (when there), and aud/iss (when configured). We support all common signature schemes.

I can see how it would maybe make sense to only validate aud/iss when set in the JWT but ignore when unset. Though at the moment we require these fields to be set, and set to one of the values we require.

Fixes #133

### DIFF
--- a/src/cfg/mod.rs
+++ b/src/cfg/mod.rs
@@ -553,6 +553,7 @@ pub struct ConfigurationInner {
     /// (object with string values) that is forwarded to internal coyote handlers.
     #[serde(default)]
     #[env_overridable(skip)]
+    #[validate(nested)]
     pub jwt: Option<JwtConfig>,
 }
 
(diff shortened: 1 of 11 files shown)


src/cfg/mod.rs
```diff
@@ -549,6 +545,67 @@ pub struct ConfigurationInner {
...
     #[serde(default)]
     pub admin_token: Option<String>,
+
+    /// Configuration for verifying JWT bearer tokens.
+    ///
+    /// When set, bearer tokens in JWT format are verified using this configuration.
+    /// The JWT must contain a `role` claim (string) and may contain a `context` claim
+    /// (object with string values) that is forwarded to internal coyote handlers.
+    #[serde(default)]
+    #[env_overridable(skip)]
```
svix-gabriel, line 555: What does `env_overridable` do?

src/cfg/mod.rs
```diff
@@ -549,6 +545,67 @@ pub struct ConfigurationInner {
...
+/// omitted, the respective claims are not checked (and may be absent from the
+/// token).
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct JwtConfig {
+    #[serde(flatten)]
+    pub key: JwtKey,
+    /// Expected `aud` values. When set, the token must contain one of these
+    /// values in its `aud` claim. When absent, `aud` is not validated.
+    #[serde(default)]
+    pub audience: Option<Vec<String>>,
```
svix-gabriel, line 578: Why use `Option<Vec<...>>` over `Vec<...>` for audience and issuer? Is there a meaningful difference between `null`/`none` and `[]`?

src/cfg/mod.rs
tasn: Ah yes, I need @svix-jplatte's help here. His mechanism for env -> cfg doesn't like `Option<T>`. I need his help with fixing it.

src/cfg/mod.rs
tasn: I can fix that, good call.

Actually, I think I like it this way. I'll add validation that array size is at least 1, but empty array means something else to empty value.

svix-gabriel (dismissed):

svix-gabriel (approved):

PR #839, opened 2026-04-03
feat(codegen): update Java SDK to use Duration instead of Long

svix-pluiz:


svix-jplatte (dismissed):
Looks good, just needs a rebase.

svix-jplatte (approved):

PR #849, opened 2026-04-06
codegen/python: Use timedelta for durations

svix-jplatte:
Based on #848.
Part of #156.

### DIFF
--- a/codegen/templates/python/types/macros.py.jinja
+++ b/codegen/templates/python/types/macros.py.jinja
@@ -1,11 +1,33 @@
-{% macro field_decl(field) %}
-    {%- if field.name | to_snake_case != field.name %}
+{% macro field_name(field) -%}
+    {%- set field_name = field.name | to_snake_case -%}
+    {%- if field_name != field.name %}
         {%- do panic("coyote API does not use non-snake_case field names") %}
     {%- endif %}
-    {{ field.name | to_snake_case }}: {{ field.type.to_python() }}
-    {%- if not field.required %} | None = None
-    {%- elif field.nullable %} | None
+    {%- if field_name is endingwith "_ms" -%}
+        {%- if not field.type.is_u64() %}{% do panic("_ms type should be uint64") %}{% endif -%}
+        {%- set field_name = field_name | strip_trailing_str("_ms") -%}
+    {%- endif -%}
+    {{ field_name }}
+{%- endmacro -%}
+{% macro field_type(field) -%}
+    {% set field_ty = field.type.to_python() -%}
+    {% if field.name is endingwith "_ms" -%}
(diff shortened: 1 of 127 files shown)


svix-pluiz (dismissed):

z-clients/python/coyote/internal/types.py
```diff
@@ -0,0 +1,25 @@
...
+
+def validate(v: object) -> timedelta:
+    if isinstance(v, timedelta):
+        # used by constructor calls
+        return v
+    if isinstance(v, int):
+        # used by model_validate
+        return timedelta(milliseconds=v)
+    else:
+        raise ValueError(f"Expected integer, got {type(v)}")
```
svix-jbrown:
```suggestion
        raise ValueError(f"Expected integer or timedelta, got {type(v)}")
```

z-clients/python/coyote/internal/types.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def serialize(td: timedelta) -> int:
+    return td.seconds * 1000 + int(td.microseconds / 1000)
```
svix-jbrown: as written, this will do truncation instead of rounding; if you want rounding, you have to do `int(round(td.microseconds / 1000))`

codegen/templates/python/types/macros.py.jinja
```diff
@@ -1,11 +1,33 @@
...
+    {%- set field_name = field.name | to_snake_case -%}
+    {%- if field_name != field.name %}
         {%- do panic("coyote API does not use non-snake_case field names") %}
     {%- endif %}
-    {{ field.name | to_snake_case }}: {{ field.type.to_python() }}
-    {%- if not field.required %} | None = None
-    {%- elif field.nullable %} | None
+    {%- if field_name is endingwith "_ms" -%}
+        {%- if not field.type.is_u64() %}{% do panic("_ms type should be uint64") %}{% endif -%}
+        {%- set field_name = field_name | strip_trailing_str("_ms") -%}
```
svix-jbrown, line 8: doing name-based stuff works, I guess, but feels so fragile.

I wonder if we could add something like `x-subtype: "TimeDeltaMs"` to the openapi doc to make it less ambiguous?

z-clients/python/coyote/internal/types.py
svix-jplatte: Yeah, I suppose that's better.

codegen/templates/python/types/macros.py.jinja
svix-jplatte: Made a ticket: https://github.com/svix/coyote-private/issues/853

svix-jbrown (approved):

PR #854, opened 2026-04-06
rust: fix rust client to properly decode error bodies

svix-jbrown:
Before this change:

```
Error: Http error (status=500 Internal Server Error) None
```

After this change:

```
Error: Http error (status=500 Internal Server Error) Some(StandardHttpError { code: "NOT_READY", detail: "this node has not yet finished bootstrapping" })
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1221,6 +1221,7 @@ name = "coyote-client"
 version = "0.0.1"
 dependencies = [
  "form_urlencoded",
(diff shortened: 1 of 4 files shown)


svix-jplatte (dismissed):
Ah, should have updated this when adding msgpack support for errors. Might affect other SDKs as well.

svix-jplatte (approved):

Issue #841, opened 2026-04-03
Update TypeScript SDK to use Uint8Array


svix-pluiz:
For example, TypeScript is using `number[]` instead of Uint8Array. We should investigate and update all SDKs that are not using the correct value

Context: https://svixhq.slack.com/archives/C0A72988962/p1775247009579609

svix-jplatte:
Go is fine, `[]byte` seems most commonly used, we emit `[]uint8` but that's the same type.
Rust is fine, `Vec<u8>` is probably the most common (could consider using some other easy-to-clone type in the future but doesn't seem terribly important).
Python is fine, we already special-cased "list of u8" from the OpenAPI spec to be `bytes` on the Python side.

Java is the only other one that likely needs a different type, currently uses `List<Byte>`.

svix-pluiz:
> Java is the only other one that likely needs a different type, currently uses List<Byte>.

I think we could change Java to use `byte[]`?

svix-jplatte:
Have googled a bit and that indeed seems like the most obvious choice.

svix-jplatte:
Done via #855, #865.

PR #855, opened 2026-04-06
Update TS SDK to use Uint8Array

svix-pluiz:
part of https://github.com/svix/coyote-private/issues/841

### DIFF
--- a/codegen/templates/javascript/api_resource.ts.jinja
+++ b/codegen/templates/javascript/api_resource.ts.jinja
@@ -33,7 +33,7 @@ export class {{ resource_type_name }} {
         {% set body_schema = api.types[op.request_body_schema_name] -%}
         {% set positional_fields = body_schema.fields | selectattr("positional") -%}
         {% for field in positional_fields -%}
-            {{ field.name | to_snake_case }}: {{ field.type.to_js() }},
+            {{ field.name | to_snake_case }}: {% if field.type.is_bytes() %}Uint8Array{% else %}{{ field.type.to_js() }}{% endif %},
         {% endfor -%}
 
         {# body parameter interface -#}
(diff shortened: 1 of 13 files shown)


svix-pluiz:
@svix-jplatte Probably the "best fix" is to update `openapi-codegen/src/api/types.rs` function `to_js_typename` to generate as a UInt8Array instead of number[].

Here is the current implementation

```rust
    fn to_js_typename(&self) -> Cow<'_, str> {
        match self {
            Self::Bool => "boolean".into(),
            Self::Int16
            | Self::UInt16
            | Self::Int32
            | Self::UInt32
            | Self::Int64
            | Self::UInt64 => "number".into(),
            Self::String | Self::Uri => "string".into(),
            Self::DateTime => "Date".into(),
            Self::JsonObject => "any".into(),
            Self::List { inner } | Self::Set { inner } => {
                format!("{}[]", inner.to_js_typename()).into()
            }
            Self::Map { value_ty } => {
                format!("{{ [key: string]: {} }}", value_ty.to_js_typename()).into()
            }
            Self::SchemaRef { name, .. } => filter_schema_ref(name, "any"),
            Self::StringConst { .. } => "string".into(),
        }
    }
```

svix-jplatte:
> @svix-jplatte Probably the "best fix" is to update `openapi-codegen/src/api/types.rs` function `to_js_typename` to generate as a UInt8Array instead of number[].

I agree. You can copy the `bytes` special-case from Python:

```rust
Self::List { inner } if matches!(**inner, Self::UInt8) => "bytes".into(),
```

and just push to the `jplatte/coyote` branch, then update the git rev here.

svix-jplatte (approved):
Though if you prefer merging this first, feel free to.

PR #858, opened 2026-04-06
Add optional idempotency to msgs publish operation

svix-lucho:
Adds an optional `idempotency_key` field in the `msgs.publish` body.

To store the idempotency key, a new table row is added in the msgs module. We return a 409 Conflict if the idempotency key has been used - simply to avoid having to store the responses.

The cleanup process for the idempotency table is still TBD. An idea is to do it by sampling keys and removing the expired ones, so we don't have to keep a secondary index for expirations. We can even consider doing the cleanup using [compaction filters](https://fjall-rs.github.io/post/fjall-3-1/), which might work well for this use case, where it's ok to defer cleanup for a while. I wrote a sample cleanup function just as a placeholder.


crates/fjall-utils/src/table_row.rs
```diff
@@ -40,10 +40,11 @@ pub trait TableRow: Sized + Serialize + DeserializeOwned {
         Ok(())
     }
 
-    fn iter<K: ReadableKeyspace>(keyspace: &K) -> impl Iterator<Item = (fjall::Slice, Self)> {
+    fn iter<K: ReadableKeyspace>(keyspace: &K) -> impl Iterator<Item = (TableKey<Self>, Self)> {
```
svix-lucho, line 49: updated this because it made no sense to return a `fjall::Slice`

tests/it/msgs/publish.rs
```diff
@@ -266,6 +268,85 @@ async fn publish_keyless_same_partition() -> TestResult {
     Ok(())
 }
 
+#[tokio::test]
+async fn publish_rejects_reused_idempotency_key() -> TestResult {
```
svix-gabriel, line 272: Can we add a test that does concurrent publishes with the same idempotency key?

crates/msgs/src/operations/publish.rs
```diff
@@ -23,10 +27,18 @@ pub struct PublishOperation {
     partition: Option<Partition>,
     msgs: Vec<MsgIn>,
     topic_id_random_bytes: UuidV7RandomBytes,
+    idempotency_key: Option<MsgsIdempotencyKey>,
 }
 
+const IDEMPOTENCY_TTL_SECS: u64 = 120;
```
tasn: Way too short, no? Should probably be at least an hour. Also, we do everything in milliseconds it's weird to have seconds all of a sudden.

crates/msgs/src/operations/publish.rs
```diff
@@ -54,13 +67,36 @@ impl PublishOperation {
...
+            if let Some(idempotency_key) = self.idempotency_key {
+                let existing = IdempotencyRow::fetch(
+                    &state.metadata_tables,
+                    IdempotencyRow::key_for(self.namespace_id, &idempotency_key),
+                )?;
+
+                if let Some(existing) = existing
+                    && existing.expiry > now
+                {
+                    return Err(Error::operation(
```
tasn: We don't use this error type, and we don't use status code conflict anyway.

This should be `Error::conflict`

crates/msgs/src/operations/publish.rs
tasn: I also see you use it as a duration later on, so let's just init a duration here instead of using a u64. `Duration::from_hours(1)`

crates/msgs/src/entities.rs
```diff
@@ -248,6 +249,22 @@ impl JsonSchema for TopicIn {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct MsgsIdempotencyKey(String);
```
tasn: It should be bytes and use the sha256 bytes, not a string.

crates/msgs/src/lib.rs
```diff
@@ -160,6 +163,38 @@ pub fn estimate_available_stream_messages(
...
+#[tracing::instrument(skip_all, fields(cleared))]
+pub fn clear_expired_in_background(state: &State, now: Timestamp) -> Result<()> {
+    let mut cleared = 0;
+    let mut iterated_keys = 0;
+
+    tracing::trace!("msgs: clearing expired idempotency keys");
+
+    let mut batch = state.db.batch();
+
+    // FIXME: this is inefficient and doesn't expire keys across different namespaces fairly.
```
tasn: Just use compaction filter to clean up the idempotency.

crates/msgs/Cargo.toml
```diff
@@ -13,12 +13,15 @@ coyote-namespace.workspace = true
 coyote-operations.workspace = true
 fjall.workspace = true
 fjall-utils.workspace = true
+hex.workspace = true
+hyper.workspace = true
```
tasn:
```suggestion
```
Not needed (see other comments).

crates/msgs/src/entities.rs
```diff
@@ -248,6 +249,22 @@ impl JsonSchema for TopicIn {
...
+
+impl MsgsIdempotencyKey {
+    pub fn new(key: &str) -> Self {
```
tasn: It should probably use more than just the key. E.g. should at least be topic name + idempotency key.

svix-lucho: ahh right I forgot about the topic

crates/msgs/Cargo.toml
svix-lucho: removed

crates/msgs/src/entities.rs
svix-lucho: fixed

tests/it/msgs/publish.rs
svix-lucho: Added

crates/msgs/src/operations/publish.rs
svix-lucho: made it one hour, and got rid of the u64

crates/msgs/src/entities.rs
svix-lucho: updated

crates/msgs/src/operations/publish.rs
svix-lucho: Updated

crates/msgs/src/operations/publish.rs
```diff
@@ -23,10 +26,18 @@ pub struct PublishOperation {
     partition: Option<Partition>,
     msgs: Vec<MsgIn>,
     topic_id_random_bytes: UuidV7RandomBytes,
+    idempotency_key: Option<MsgsIdempotencyKey>,
 }
 
+static IDEMPOTENCY_TTL: LazyLock<DurationMs> = LazyLock::new(|| DurationMs::from_hours(1));
```
svix-jbrown: #897 adds a const constructor here so you can avoid locking all the time

crates/msgs/src/entities.rs
```diff
@@ -248,6 +249,24 @@ impl JsonSchema for TopicIn {
...
+#[serde(transparent)]
+pub struct MsgsIdempotencyKey([u8; 32]);
+
+impl MsgsIdempotencyKey {
+    pub fn new(key: &str, topic: &TopicName) -> Self {
+        let mut hasher = Sha256::new();
+        hasher.update(topic.as_bytes());
+        hasher.update(b":");
+        hasher.update(key.as_bytes());
+        Self(hasher.finalize().into())
```
svix-jbrown, line 264: we don't want to mix the user or anything else into it? in production, we include the full auth token hashed value in the idempotency key so that users can't ever mess with one another.

crates/msgs/src/operations/publish.rs
```diff
@@ -54,13 +66,33 @@ impl PublishOperation {
...
+                let existing = IdempotencyRow::fetch(
+                    &state.metadata_tables,
+                    IdempotencyRow::key_for(self.namespace_id, &idempotency_key),
+                )?;
+
+                if let Some(existing) = existing
+                    && existing.expiry > now
+                {
+                    return Err(Error::conflict("idempotency key already used".to_owned()));
+                }
```
svix-jbrown, line 85: it might be nice to include the expiry here, and turn it into a `Retry-After` header in the `IntoResponse` impl

crates/msgs/src/lib.rs
svix-lucho: Done

crates/msgs/src/operations/publish.rs
svix-lucho: I did it

crates/msgs/src/operations/publish.rs
svix-lucho: nice. Updated it

crates/msgs/src/entities.rs
svix-lucho: I added the auth token to the mix, like we do in webhooks

crates/msgs/src/compaction.rs
```diff
@@ -0,0 +1,135 @@
...
+impl CompactionFilter for IdempotencyExpiryFilter {
+    fn filter_item(&mut self, item: ItemAccessor<'_>, _ctx: &Context) -> CompactionFilterResult {
+        let key = item.key();
+        if key.first().copied() != Some(IdempotencyRow::ROW_TYPE) {
+            return Ok(Verdict::Keep);
+        }
+
+        let value = item.value()?;
+        let Ok(row) = IdempotencyRow::from_fjall_value(value) else {
+            return Ok(Verdict::Keep);
```
tasn: The else here means that it failed to parse? So we keep on failure to parse? Shouldn't we remove + print an error?

svix-lucho: this is because I'm using the same keyspace as other non-idempotency TableRows - we have to ignore those. The alternative is to add a keyspace just for idempotency, but it seemed unnecessary just for this.

tasn: But you already check the row type, no? That's what that's for. This would error only on real errors.

svix-lucho: ah yeah you're right. Changed it to remove + print an error

svix-gabriel (approved):

PR #859, opened 2026-04-06
cli: teach the cli to allow strings for values

svix-jbrown:
This is maybe an off-the-wall idea I had, but it would be nice to be able to use the `kv` / `cache` / etc modules without hand-crafting binary strings.

Without this branch (a.k.a. on main):

```
$ coyote kv set foo bar
error: invalid value 'bar' for '[VALUE]...': invalid digit found in string
$ coyote kv set foo 49 50 51
{
    "success": true,
    "version": 19225
}
$ coyote kv get foo
{
    "value": "123",
    "version": 19225
}
```

With this branch:

```
$ coyote kv set foo bar
{
    "success": true,
    "version": 18282
}
$ coyote kv get foo
{
    "value": "bar",
    "version": 18282
}
```

### DIFF
--- a/codegen/templates/cli/api_resource.rs.jinja
+++ b/codegen/templates/cli/api_resource.rs.jinja
@@ -8,6 +8,9 @@
 use coyote::CoyoteClient;
 use clap::{Args, Subcommand};
 
+#[allow(unused)]
+use crate::prelude::*;
+
 {% for _, sub in resource.subresources | items -%}
 use super::{{ sub.name | to_upper_camel_case }}Args;
 {% endfor %}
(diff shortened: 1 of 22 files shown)


svix-jplatte:
Oh, this broke when values were made positional. We already special-case the bytestrings in JSON.

codegen/templates/cli/api_resource.rs.jinja
```diff
@@ -65,7 +68,12 @@ pub enum {{ resource_type_name }}Commands {
                     {% if field.description is defined -%}
                         {{ field.description | to_doc_comment(style="rust") }}
                     {% endif -%}
-                    {{ field.name | to_snake_case }}: {{ field.type.to_rust() }},
+                    {% if field.type.to_rust() == "Vec<u8>" -%}
```
svix-jplatte:
```suggestion
                    {% if field.type.is_bytes() -%}
```

codegen/templates/cli/api_resource.rs.jinja
```diff
@@ -121,7 +129,11 @@ impl {{ resource_type_name }}Commands {
                                     | selectattr("positional", "false")
                                     | length == 0 %}
                                 {% for field in body_schema.fields | selectattr("positional") -%}
-                                    {{ field.name | to_snake_case }},
+                                    {% if field.type.to_rust() == "Vec<u8>" -%}
```
svix-jplatte: Same

svix-jplatte (dismissed):

tasn (approved):

PR #865, opened 2026-04-06
Update Java SDK to use byte[] instead of List<byte>

svix-pluiz:
Part of https://github.com/svix/coyote-private/issues/841
Surprisingly, I had to create a custom serializer/deserializer to support byte[] conversion

### DIFF
--- a/z-clients/java/src/main/java/com/svix/coyote/ByteArrayAsIntArrayDeserializer.java
+++ b/z-clients/java/src/main/java/com/svix/coyote/ByteArrayAsIntArrayDeserializer.java
@@ -1,0 +1,22 @@
+package com.svix.coyote;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.databind.DeserializationContext;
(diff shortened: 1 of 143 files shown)


z-clients/java/src/main/java/com/svix/coyote/ByteArrayAsIntArrayDeserializer.java
```diff
@@ -0,0 +1,22 @@
+package com.svix.coyote;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+
+public class ByteArrayAsIntArrayDeserializer extends JsonDeserializer<byte[]> {
```
svix-jplatte, line 10: I was briefly confused about "as int array", then remembered that we still use JSON rather than msgpack here. I'm guessing this will have to change again once we switch to msgpack.

svix-jplatte (approved):

PR #875, opened 2026-04-07
Update Java SDK to use msgpack

### LINKED ISSUE #869: Update Java SDK to use msgpack for requests, responses
Add msgpack encoding / decoding, make sure to set `content-type` and `accept` header.

Make sure to also decode msgpack errors (ideally, honor the `Content-Type` header and parse errors as JSON or MsgPack appropriately). See #854 for an example in the Rust client library and #857 for the Python client library.

svix-pluiz:
closes https://github.com/svix/coyote-private/issues/869


z-clients/java/src/main/java/com/svix/coyote/Utils.java
svix-jplatte, line 26: Do we still need this?

svix-jplatte (approved):

z-clients/java/src/main/java/com/svix/coyote/Utils.java
svix-pluiz: We have macros in `struct.java.jinja` and `struct_enum.java.jinja` that generates `fromJson` and `toJson` methods in several different Classes, I'm not sure if we can remove

svix-jplatte: Remove it and see. Please also have to make sure one of those struct enums (if there's even more than one now) is used by integration tests.

PR #876, opened 2026-04-07
go: Use custom type for durations

svix-jplatte:
Based on #874 (now merged).
Part of #156.


svix-jplatte:
Maybe this isn't much better than the situation before. Though Go offers few features that would make this type convenient to use. I suppose a constructor from the standard duration might be useful, but not even sure what to name it.

svix-gabriel (commented):
Why a custom type and not the native golang duration? https://pkg.go.dev/time#Duration

svix-jplatte:
Because the only way to customize type serialization is through a _global_ encoder / decoder registry, and I don't want to affect people's use of the msgpack library independently of the SDK. (also I'm not even sure whether that customization point works for types that are natively supported by the library)

svix-gabriel (approved):

PR #881, opened 2026-04-07
Simplify tracing http attributes

svix-jplatte:
Should fix traces in datadog.

### DIFF
--- a/src/core/otel_spans.rs
+++ b/src/core/otel_spans.rs
@@ -77,10 +77,7 @@ impl<B> MakeSpan<B> for AxumOtelSpanCreator {
             .get(header::HOST)
             .and_then(|header| header.to_str().ok());
 
-        let http_route = request
-            .extensions()
-            .get::<MatchedPath>()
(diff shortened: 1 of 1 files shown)


svix-jbrown (commented):
is `request.uri().path()` guaranteed to be normalized the same way as in routing?

e.g., of someone make a request to `/foo/../bar/..//api/v1.admin.cluster`, I would expect us to route that to `/api/v1.admin.cluster`, but for `uri.path()` to be `/foo/../bar/..//api/v1.admin.cluster`...

C-f "normalize" on <https://docs.rs/http/1.3.1/http/uri/struct.Uri.html#method.path> has no results

svix-jplatte:
Good point. I'll defer figuring that out and get this PR into a landable state regardless.

svix-jplatte:
Removed the second commit and moved one simplification from the second commit that's unrelated to the comment above into a separate PR at #882.

svix-jbrown (approved):

PR #882, opened 2026-04-07
Exclude some middlewares for non-routable paths

svix-jplatte:
Cleaner this way.


svix-jbrown (approved):
TIL about `.route_layer`.

PR #887, opened 2026-04-08
Use DurationMs for configuration

svix-jplatte:
A little more convenient than `serde` helpers.


svix-gabriel (dismissed):

crates/cache/src/lib.rs
```diff
@@ -52,7 +50,7 @@ pub struct CacheModel {
 pub struct AllNodesWorker {
     state: State,
     time: Monotime,
-    cleanup_interval: Duration,
+    cleanup_interval: DurationMs,
```
svix-jbrown: rather than promulgate `DurationMs` everywhere, why use `.into()` when constructing this? I don't think it makes sense to make every module use this custom type.

svix-jplatte: If the sub-millisecond bits are always zero, why have them around? The custom type should be just as easy to use as the std type, for most things. Don't mind changing it if you prefer, though.

svix-jbrown (approved):

Issue #809, opened 2026-04-02
Use ms-since-unix-epoch for timestamp representation in msgpack


svix-jplatte:
Right now it's a string, and the Go SDK is failing to interact with it (because it tries to use a msgpack extension for it instead).

PR #889, opened 2026-04-08
Remove datetime types from Java and JavaScript SDK

svix-pluiz:
Part of https://github.com/svix/coyote-private/issues/809


z-clients/javascript/src/request.ts
```diff
@@ -88,7 +88,7 @@ export function makeRequestContext(token: string, options: CoyoteOptions) {
   };
 }
 
-type QueryParameter = string | boolean | number | Date | string[] | null | undefined;
+type QueryParameter = string | boolean | number | string[] | null | undefined;
```
svix-jplatte, line 91: We should remove this whole concept of query parameters, we don't use them here.

svix-pluiz: > We should remove this whole concept of query parameters, we don't use them here.

I'm going to open a separate PR for that

PR #891, opened 2026-04-08
Rename coyote-proto => diom-proto

svix-jplatte:
Let's get this rename train started.


svix-jbrown:
Do we have to? Actually renaming everything is going to make all of the git history totally useless.

If we're going to rename every module, we should at least have a fake user do the commit so that it's really obvious in the commit history.

svix-jplatte:
It's not that bad in my experience, but we can try other strategies for implementing it. I don't think not renaming is an option :smile:

tasn (approved):

svix-jbrown (commented):
I think renaming the modules internally is not a very good use of our time, since it's just an implementation detail and means people will be cursed to pass `--follow` on every subsequent git invocation for all time, but if we feel like branding is more important than future devex, who am I to argue.

tasn:
What I shared [on Slack](https://svixhq.slack.com/archives/C0A72988962/p1775741398811639?thread_ts=1775741140.346319&cid=C0A72988962):

> I see the `--follow` concern, but I think we need to do it because:
> - It's used in macros and stuff, e.g. in the rust_log macros, and in the actual log lines. So it's not really internal.
> - We want to rename msgs and the likes to have the prefix anyway (the directory) so we need to rename quite a few anyway.
> Either way, let's fix env vars, binary names, etc.

> I guess for the 1st one we could potentially just rename the crate but keep the directory, which doesn't make people need to --follow, but that's confusing, no?

tasn:
Doing the history rewrite instead.

PR #899, opened 2026-04-08
Add new traits for Fjall table keys, use traits with MsgRow

svix-gabriel:
This tackles the key side of https://github.com/svix/diom-private/issues/374.

There are a few goals we're trying to address here

- Construction of keys should be trivial/easy/mostly-automatic. You shouldn't have to manually write logic to parse a `fjall::UserKey` into a struct, and vice versa.
- Key construction should be "as performant as possible". We want to minimize (de)serialization overhead and avoid unnecessary allocations
- It should be easy to do other QoL things with keys, like creating key ranges, or extracting specific fields without parsing the hole key. Especially with composite keys.
- Balance all of these things with the fact that versioning keys is basically impossible, so we have to make breaking changes to keys explicit (so we can catch them in PR review), and allow for enough flexibility that we don't lock ourselves out of future patterns.

To that end, this adds a `FjallKeyAble` trait (maybe there's a better name) and derive macro, to make converting structs into fjall keys.

```rust
#[derive(FjallKeyAble)]
#[table_key(prefix = 0)]
struct ExampleCompositeKey {
    #[key(0)]
    id: u32,
    #[key(1)]
    group: String,
}
```

When you call `ExampleCompositeKey::fjall_key()`, this will generate a byte key that looks like
```
   <prefix byte> | <key-0-field> | <key-1-field> | ...
```

The only limitation here is that

* Variable sized structs (String, Vec) have to be the last field in the key
* You cannot have more than one variable sized field in the key

### Compile Errors

If you misconfigure a key, you get helpful errors. e.g. you use more than one dynamic key

Screenshot 2026-04-10 at 16 34 00

etc

### How to Review

This PR is split into two-3 commits.

The first just adds the trait and derive macro, with tests showing how it works. I've never actually written a derive macro before, and did rely quite a bit on Claude here for the derive macro, for full disclosure. I recommend looking at the tests here since they show a lot of the macro functionality.

The last commit actually changes the MsgRows to use this new pattern. I wanted to migrate at least one struct to show this works, and to demonstrate how the patterns are helpful.

The big value add is that code like this can be completely removed
```rust
    pub(crate) fn key_for(
        topic_id: TopicId,
        partition: Partition,
        offset: Offset,
    ) -> TableKey<Self> {
        TableKey::init_key(
            Self::ROW_TYPE,
            &[
                topic_id.as_bytes(),
                &partition.get().to_be_bytes(),
                &offset.to_be_bytes(),
            ],
            &[],
        )
    }
```
Eventually we can get rid of `TableKey::init_key` too, once all tables are migrated.

The other important diff is this
```diff
                let key = kv.key()?;
-                let offset = u64::from_be_bytes(
-                    key[key.len().saturating_sub(SIZE_U64)..]
-                        .try_into()
-                        .expect("We know the size is right"),
-                );
+                let offset = MsgKey::extract_offset(&key).expect("valid MsgKey in msg table");
                Ok(offset + 1)
```

Instead of manually parsing the offset from the key, we can rely entirely on the generated/automatic `extract_offset(...)` method the derive macro gives us. The macro gives us an `extract_<field>` method for each field in the key, and returns reference types for things like String and Vec.

Finally, this does put us in the weird limbo state where `TableKey<T>` (the struct) exists side by side with `FjallKeyAble`. The end goal here is to kill `TableKey<T>`, and possibly even add strict type associations with Keys and Value structs. However, I wanted to keep the diff smaller (we can migrate individual tables individually), and I didn't want to fiddle with the Value traits/serialization since I know JB is touching those separately.

### Performance

I did rerun`msg` benchmarks locally, and there wasn't a noticeable change. The new key format doesn't do anything involving complex deserialization formats or allocations, so this doesn't suprise me.


crates/diom-derive/src/fjall_key.rs
```diff
@@ -0,0 +1,269 @@
...
+) -> TokenStream {
+    let field_refs: Vec<TokenStream> = fields.iter().map(&field_expr).collect();
+    let _num_fields = fields.len();
+
+    quote! {
+        let __prefix = #prefix.as_bytes();
+        // prefix + null after prefix + (field bytes + null) per field
+        let __total_len = __prefix.len() + 1
+            #(+ ::fjall_utils::KeyComponent::key_len(&#field_refs) + 1)*;
+        let mut __buf = vec![0u8; __total_len];
```
svix-gabriel: There's potentially an optimization here where we use a stack allocated array of `[u8; FIXED_BUFFER_SIZE]`.

It adds a bit of complexity since it means we need to switch to a `Vec` buffer for keys larger than the FIXED_BUFFER_SIZE.

This wouldn't be a breaking change, so I opted to _not_ do the optimization here and then revisit later.

crates/diom-derive/src/fjall_key.rs
```diff
@@ -0,0 +1,269 @@
+use proc_macro2::TokenStream;
```
tasn, line 1: Need to benchmark to make sure we aren't slowing things down. I'd benchmark on kv, as that's what we've been benching on. I know you said you benchmarked, but how did you benchmark, can you share the methodology and results?

I don't think the extra nul bytes would be that bad, but maybe the separate copies would suck a bit.

crates/diom-derive/src/fjall_key.rs
```diff
@@ -0,0 +1,269 @@
...
+use quote::quote;
```
tasn, line 2: What do you mean by composite key? I'm not sure I follow the terminology.

crates/diom-derive/src/fjall_key.rs
tasn, line 1: Given that we already support annotating the key fields, can we potentially mark items as fixed sized and then we don't need the null delimiter (and it will also not break things)?

svix-gabriel: Yup I mentioned this in the PR description

> ## Performance
> I did rerunmsg benchmarks locally, and there wasn't a noticeable change. The new key format doesn't do anything involving complex deserialization formats or allocations, so this doesn't suprise me.

crates/diom-derive/src/fjall_key.rs
svix-gabriel: Just a key with multiple logical fields. e.g. `id` vs `(group, id)`.

So the msg rows, which are key'd on topic + partition + offset, would be a composite key

crates/diom-derive/src/fjall_key.rs
tasn, line 1: Way add the string prefix instead of using the enum prefix that we currently use?

Having a fixed size prefix means we can jump to specific parts of the key without having to scan it. E.g. you can just do `key[offset..end]` because you know the size, so you know how long to skip.

crates/fjall-utils/src/fjall_key_able.rs
```diff
@@ -0,0 +1,249 @@
...
+    fn key_len(&self) -> usize;
+
+    /// Writes key bytes into `buf`, returning the number of bytes written.
+    ///
+    /// `buf` must be at least [`key_len`](Self::key_len) bytes long.
+    fn write_to_key(&self, buf: &mut [u8]) -> usize;
+
+    /// Parses a value from the beginning of `buf`. Returns the parsed value
+    /// and the number of bytes consumed (not including any trailing delimiter).
+    fn read_from_key(buf: &[u8]) -> Result<(Self, usize), Cow<'static, str>>
```
tasn, line 132: Maybe I'm missing something, but I don't get how this can work?

If you serialize bytes, e.g. consider `4u32` it'll have three `\0` in it. You can't rely on nul delimiting for the byte parts.

Take a look at how it's currently done. There's a reason why it separates fixed-size and non-fixed sized parts, because you have to use the compile-time known size.

crates/diom-derive/src/fjall_key.rs
svix-gabriel: Potentially, I think it would have to be manual though. Like `#[key(0, variablesize)]` or something.

My very strong hunch here is that the overhead from null delimiter's is so trivial that it's not worth the added complexity.

svix-gabriel: The string prefix is fixed size too, so you can still jump to specific fields without doing an entire scan. At least in theory

That being said, one of the goals here is that we _shouldn't_ be handwriting code to parse specific fields from the fjall::UserKey, since it's very error prone. I call this out in the PR description, where I point out this diff

```diff
                let key = kv.key()?;
-                let offset = u64::from_be_bytes(
-                    key[key.len().saturating_sub(SIZE_U64)..]
-                        .try_into()
-                        .expect("We know the size is right"),
-                );
-                Ok(offset + 1)
+                let parsed = MsgRowKey::from_fjall_key(key).or_internal_error()?;
+                Ok(parsed.offset + 1)
```

Fundamentally the (de)serialization logic for these keys is really simple, it's just copying bytes with null delimiters. I think the simplicity of the interface and reduced risk of misparsing keys worth the trivial cost of parsing the entire key.

tasn: > I know you said you benchmarked, but how did you benchmark, can you share the methodology and results?

Yeah, I know, I asked the above too ^

crates/fjall-utils/src/fjall_key_able.rs
svix-gabriel: I think you're overcomplicating it.

For fixed size types we don't bother scanning for `\0` bytes. You can look at https://github.com/svix/diom-private/pull/899/changes/BASE..8ca4325a3b9f6adfecd7be5ed8f694f931147d1b#diff-d978b6e29c2e846b2bf6abb36e54bdcfa854c7fa5624998c9a64bc7480a57e14R87 to verify. We know it's fixed size, so we just read the bytes, like you said.

The null byte delimiting only breaks down for types that are BOTH variable sized AND can contain null bytes. So this does preclude using something like `Vec<u8>` in a key.

crates/diom-derive/src/fjall_key.rs
tasn: Gotcha. So composite is the common path.

crates/diom-derive/src/fjall_key.rs
svix-gabriel: I just did the CLI benchmark. I'll redo it again, but I think you're being extraordinarily pessimistic here. We aren't doing any more "separate copies" than we were before

tasn: Sorry, I wasn't clear, but the comment wasn't about performance.

I added more about the reasoning here https://github.com/svix/diom-private/pull/899/changes#r3059417124

Though essentially the reason is that sometimes you want to use fixed size in the normal "fixed size" section, and sometimes in the variable place (we have a key somewhere that does it).

crates/fjall-utils/src/fjall_key_able.rs
tasn: Yeah, that clarifies it, thanks! Sorry, very hard to follow the macro stuff.

We can't use `Vec<u8>` for keys for sure.

crates/diom-derive/src/fjall_key.rs
tasn: I spent 12hrs yesterday shaving single digit %, and a lot is just random array copies. So definitely don't want to regress.

tasn: My point is that we probably want to be able to parse parts of the key, not the whole key. In fact, we already do that in some places (I think in kv expiry?).

So if the helper can just jump to the right offset without having to scan, search for nulls, skip variable length parts, etc. That's much much better. Not saying we write it manually, but the helper function can have it.

tasn: Looking at your comment on my other comment, it looks like you already know what's fixed sized and what's variable sized, no? So you already can do what's suggested here?

svix-gabriel: Ok, here's the benchmarking results. Using --release, with a clean db each time.

Stream (pre change)
```
stream_single_server/publish_batch_100
                        time:   [266.25 µs 290.96 µs 318.36 µs]
Found 5 outliers among 60 measurements (8.33%)
  1 (1.67%) high mild
  4 (6.67%) high severe
stream_single_server/receive_commit_batch_100
                        time:   [112.12 µs 112.78 µs 113.45 µs]
Found 9 outliers among 60 measurements (15.00%)
  1 (1.67%) low severe
  1 (1.67%) low mild
  6 (10.00%) high mild
  1 (1.67%) high severe

stream_cluster/publish_batch_100
                        time:   [532.11 µs 616.37 µs 715.38 µs]
Found 13 outliers among 60 measurements (21.67%)
  3 (5.00%) high mild
  10 (16.67%) high severe
stream_cluster/receive_commit_batch_100
                        time:   [265.27 µs 266.35 µs 267.42 µs]
Found 3 outliers among 60 measurements (5.00%)
  1 (1.67%) high mild
  2 (3.33%) high severe
```

Stream (post change)
```
stream_single_server/publish_batch_100
                        time:   [265.41 µs 294.20 µs 327.63 µs]
                        change: [−17.609% +0.1863% +21.349%] (p = 0.99 > 0.05)
                        No change in performance detected.
Found 4 outliers among 60 measurements (6.67%)
  3 (5.00%) high mild
  1 (1.67%) high severe
stream_single_server/receive_commit_batch_100
                        time:   [112.56 µs 113.23 µs 113.88 µs]
                        change: [−1.0309% +0.1725% +1.3135%] (p = 0.78 > 0.05)
                        No change in performance detected.
Found 5 outliers among 60 measurements (8.33%)
  3 (5.00%) high mild
  2 (3.33%) high severe

stream_cluster/publish_batch_100
                        time:   [520.34 µs 601.80 µs 698.62 µs]
                        change: [−19.033% −2.3385% +18.291%] (p = 0.81 > 0.05)
                        No change in performance detected.
Found 12 outliers among 60 measurements (20.00%)
  2 (3.33%) high mild
  10 (16.67%) high severe
stream_cluster/receive_commit_batch_100
                        time:   [264.77 µs 265.73 µs 266.71 µs]
                        change: [−1.4980% −0.3431% +0.7726%] (p = 0.57 > 0.05)
                        No change in performance detected.
Found 3 outliers among 60 measurements (5.00%)
  2 (3.33%) high mild
  1 (1.67%) high severe
```



Queue (pre change)
```
queue_single_server/receive_ack_batch_100
                        time:   [113.66 µs 115.01 µs 116.58 µs]
Found 6 outliers among 60 measurements (10.00%)
  3 (5.00%) high mild
  3 (5.00%) high severe
queue_single_server/receive_only_batch_100
                        time:   [2.0160 ms 2.0401 ms 2.0601 ms]
Found 12 outliers among 60 measurements (20.00%)
  9 (15.00%) low mild
  2 (3.33%) high mild
  1 (1.67%) high severe

queue_cluster/receive_ack_batch_100
                        time:   [264.93 µs 266.43 µs 267.92 µs]
Found 3 outliers among 60 measurements (5.00%)
  1 (1.67%) low mild
  1 (1.67%) high mild
  1 (1.67%) high severe
queue_cluster/receive_only_batch_100
                        time:   [2.4308 ms 2.4682 ms 2.4996 ms]
Found 4 outliers among 60 measurements (6.67%)
  1 (1.67%) low mild
  2 (3.33%) high mild
  1 (1.67%) high severe
```

Queue (post change)
```
queue_single_server/receive_ack_batch_100
                        time:   [113.00 µs 113.55 µs 114.10 µs]
                        change: [−2.2177% −0.7244% +0.6114%] (p = 0.34 > 0.05)
                        No change in performance detected.
Found 4 outliers among 60 measurements (6.67%)
  2 (3.33%) high mild
  2 (3.33%) high severe
queue_single_server/receive_only_batch_100
                        time:   [1.9851 ms 2.0080 ms 2.0262 ms]
                        change: [−6.1358% −1.3619% +3.2679%] (p = 0.59 > 0.05)
                        No change in performance detected.
Found 6 outliers among 60 measurements (10.00%)
  3 (5.00%) low severe
  1 (1.67%) low mild
  2 (3.33%) high severe

queue_cluster/receive_ack_batch_100
                        time:   [264.66 µs 265.88 µs 267.14 µs]
                        change: [−0.6934% +0.2576% +1.3266%] (p = 0.62 > 0.05)
                        No change in performance detected.
Found 2 outliers among 60 measurements (3.33%)
  1 (1.67%) high mild
  1 (1.67%) high severe
queue_cluster/receive_only_batch_100
                        time:   [2.4633 ms 2.4997 ms 2.5297 ms]
                        change: [−5.8846% +0.8042% +7.3680%] (p = 0.83 > 0.05)
                        No change in performance detected.
Found 4 outliers among 60 measurements (6.67%)
  1 (1.67%) high mild
  3 (5.00%) high severe
```

I don't know how much noise we typically see, and I can do larger batch sizes if you think it'll give a better signal.

tasn: Can you run the CLI tester (both the server and the CLI as release), I feel like it's more reliable.

Here is what I run:
```
cargo run --release --package=coyote-cli -- benchmark --filter 'msgs.' --concurrency 20 --batch-size 10 --iterations 2000
```

Hard to get useful info when it's measured in microseconds.

crates/msgs/src/tables.rs
```diff
@@ -266,6 +264,24 @@ impl TableRow for QueueConfigRow {
     const ROW_TYPE: u8 = RowType::QueueConfig as u8;
 }
 
+#[derive(FjallKeyAble)]
+#[table_key(prefix = "_msg_")]
```
tasn: Why have the prefix as a string and not the row type? Or rather, if you want to drop the row type, why use a string and not a u8? (well varint u8)

Strings are longer so they take more space and are slower to compare. Though they also feel worse than a compile-time verified value?

crates/diom-derive/src/fjall_key.rs
tasn, line 1: Another comment about the null delimiter: even if they don't add a lot of overhead, they still don't serve a purpose after fixed-size elements, so I don't understand why put them there. This is one of our hottest code paths, and also something we can't ever change afterwards, so getting it to what we want feels beneficial.

You only need a null following a string, if the type is of a known size you don't need a null. Since we are now generating with code (rather than what we did manually before), it's easy to get this right.
1. If type is `Vec<T>`: error.
2. If type is of known size (u64, `[u8; 32]`, whatever) just put the bytes.
3. If type is a String put the bytes + a null.

crates/diom-derive/src/fjall_key.rs
tasn: I think we only have one area where it's not composite, or even just one place, which is the raft log. And even then, I suspect it may change soon as we want to introduce multi-raft and it will be composite too. So if it simplifies things, just supporting that is probably fine.

crates/diom-derive/src/fjall_key.rs
```diff
@@ -0,0 +1,269 @@
...
+    ident: Ident,
+    index: usize,
+}
+
+fn parse_key_index(field: &Field) -> Result<Option<usize>, syn::Error> {
+    for attr in &field.attrs {
+        if attr.path().is_ident("key") {
+            let lit: LitInt = attr.parse_args()?;
+            return Ok(Some(lit.base10_parse()?));
+        }
```
svix-jbrown, line 16: it seems a little weird to use the path `key`; I think most libraries try to namespace their attributes with the name of the derive-macro so that it's clear who owns what (e.g., for the `env-overridable` derive macro, it's `env_overridable(skip)` or `env_overridable(nest_with_prefix("foo"))`)

crates/diom-derive/src/fjall_key.rs
```diff
@@ -0,0 +1,269 @@
...
+}
+
+fn parse_prefix(input: &DeriveInput) -> Result<LitStr, syn::Error> {
+    for attr in &input.attrs {
+        if attr.path().is_ident("table_key") {
+            let mut prefix = None;
+            attr.parse_nested_meta(|meta| {
+                if meta.path.is_ident("prefix") {
+                    let value = meta.value()?;
+                    prefix = Some(value.parse::<LitStr>()?);
```
svix-jbrown: this seems to tie us to using stringly-typed prefixes; I feel like we had intentionally chosen to use u8-enum prefixes to save space before?

crates/diom-derive/src/fjall_key.rs
tasn: One more thing about it: we probably want in general zero-copy extraction here rather than parsing the whole key into a struct every time we want a part from a key.

E.g.

```
    pub(crate) fn extract_ts_from_fjall_key(key: &fjall::UserKey) -> Result<Timestamp> {
        let offset = 1 /* row_type */;
        let array: [u8; 8] = (&key[offset..=size_of::<i64>()])
            .try_into()
            .or_internal_error()?;
        let millis = i64::from_be_bytes(array);
        Timestamp::from_millisecond(millis).or_internal_error()
    }
```

Because we don't want to parse the string part, and copy the byte parts, we just want the part that we want.

crates/diom-derive/src/fjall_key.rs
```diff
@@ -0,0 +1,269 @@
...
+        "missing #[table_key(prefix = \"...\")] attribute",
+    ))
+}
+
+fn collect_fields(input: &DeriveInput) -> Result<Vec<KeyField>, syn::Error> {
+    let syn::Data::Struct(data) = &input.data else {
+        return Err(syn::Error::new(
+            input.ident.span(),
+            "FjallKeyAble can only be derived for structs",
+        ));
```
svix-jbrown, line 50: jonas's feedback in #728 was that it's preferable to return `compile_error!` rather than a spanned error for this

crates/diom-derive/src/fjall_key.rs
```diff
@@ -0,0 +1,269 @@
...
+                    index: decl_order,
+                });
+            }
+        }
+    }
+
+    if any_indexed && any_unindexed {
+        return Err(syn::Error::new(
+            input.ident.span(),
+            "either all fields must have #[key(N)] or none",
```
svix-jbrown: I kind of feel like we should just require all; otherwise, adding a new field in the middle is a silent breaking change

crates/diom-derive/src/fjall_key.rs
```diff
@@ -0,0 +1,269 @@
...
+
+        #({
+            let __written = ::fjall_utils::KeyComponent::write_to_key(&#field_refs, &mut __buf[__pos..]);
+            __pos += __written;
+            __buf[__pos] = 0x00;
+            __pos += 1;
+        })*
+
+        debug_assert_eq!(__pos, __total_len);
+        ::fjall::UserKey::from(__buf)
```
svix-jbrown: you're generating an isolated function; you don't actually need to name all your variables with `__`. I think we should optimize for readaibility here.

crates/diom-derive/src/fjall_key.rs
```diff
@@ -0,0 +1,269 @@
...
+                        )))?;
+                __pos += __consumed;
+                if __bytes.get(__pos).copied() != ::std::option::Option::Some(0u8) {
+                    return ::std::result::Result::Err(::std::borrow::Cow::Owned(::std::format!(
+                        "missing null delimiter at position {} after field `{}`",
+                        __pos, #field_name
+                    )));
+                }
+                __pos += 1;
+            }
```
svix-jbrown, line 143: same as above; no need to add the `__` everywhere

crates/diom-derive/src/fjall_key.rs
```diff
@@ -0,0 +1,269 @@
...
+            let local = quote::format_ident!("__field_{}", ident);
+            quote! { #ident: #local }
+        })
+        .collect();
+
+    quote! {
+        let __bytes: &[u8] = &__key;
+        let __prefix = #prefix.as_bytes();
+        if !__bytes.starts_with(__prefix) {
+            return ::std::result::Result::Err(::std::borrow::Cow::Owned(::std::format!(
```
svix-jbrown, line 159: I get the defensiveness against someone renaming `Result` or `Ok` or whatever, but it really is hard to read with the fully-qualified names everywhere

crates/diom-derive/src/fjall_key.rs
```diff
@@ -0,0 +1,269 @@
...
+                "key does not start with expected prefix {:?} (for {})",
+                __prefix, ::std::stringify!(#struct_name)
```
svix-jbrown: can't we stringify this in the macro instead of in the generated code?

crates/diom-derive/src/fjall_key.rs
```diff
@@ -0,0 +1,269 @@
...
+            fn range_inclusive(start: Self, end: Self) -> ::std::ops::RangeInclusive<::fjall::UserKey> {
+                let __start = { #range_build_start };
+                let __end = { #range_build_end };
+                __start..=__end
+            }
+
+            fn range_end_inclusive(end: Self) -> ::std::ops::RangeToInclusive<::fjall::UserKey> {
+                let __end = { #range_build_end_only };
+                ..=__end
+            }
```
svix-jbrown: note that these are only valid for keys where every field is monotonically-ordered with its byte-encoding (notably, they do not work for signed integers using two's complement).

It would be nice if we could detect that in the macro and only generate this if all keys are types that implement some trait. I already have a `MonotonicTableKey` for this.

I'm not sure what the best way to do this is, but one option would be:

 1. Implement `MonotonicTableKey` for the types we care about (u64, String, Uuid, etc)
 2. Implement `MonotonicTableKey` for tuples of `MonotonicTableKey`
 3. Add a new `KeyFamily` associated type to `FjallKeyAble` (or create a new trait containing only that associated type if we want to hide this implementation details)
 4. In your derive macro, populate the `KeyFamily` associated type with a tuple of all of the types found in the struct.
 5. Create a new trait that has the range methods and is only implemented where there's an appropriate tablekeyfamily.

Put together this would look something like

```rust
impl MonotonicTableKey for u64 {}

impl MonotonicTableKey for String {}

macro_rules! impl_monotonic_table_key_for_tuple {
    ( $( $N:literal )+ ) => {
        paste! {
            impl<$([< T $N >],)+> MonotonicTableKey for ($([< T $N >],)+)
                where $([< T $N >]: MonotonicTableKey,)+ {}
        }
    };
}

impl_monotonic_table_key_for_tuple! {0}
impl_monotonic_table_key_for_tuple! {0 1}
impl_monotonic_table_key_for_tuple! {0 1 2}
impl_monotonic_table_key_for_tuple! {0 1 2 3}
impl_monotonic_table_key_for_tuple! {0 1 2 3 4}
impl_monotonic_table_key_for_tuple! {0 1 2 3 4 5}
impl_monotonic_table_key_for_tuple! {0 1 2 3 4 5 6}
impl_monotonic_table_key_for_tuple! {0 1 2 3 4 5 6 7}
impl_monotonic_table_key_for_tuple! {0 1 2 3 4 5 6 7 8}
impl_monotonic_table_key_for_tuple! {0 1 2 3 4 5 6 7 8 9}


trait FjallRangeKeyable {
  fn range(start: Self, end: Self) -> ::std::ops::Range<::fjall::UserKey>;
}

impl<T> FjallRangeKeyable for T
  where T: FjallUserKeyAble,
    T::KeyFamily: MonotonicTableKey {
  fn range(start: Self, end: Self) -> ::std::ops::Range<::fjall::UserKey> {
    // ..
  }
}
```

I think that would make it a little harder to "hold it wrong" and get weird behavior.

crates/fjall-utils/src/fjall_key_able.rs
```diff
@@ -0,0 +1,249 @@
...
+        buf[0] = *self;
+        1
+    }
+
+    fn read_from_key(buf: &[u8]) -> Result<(Self, usize), Cow<'static, str>> {
+        buf.first()
+            .copied()
+            .map(|b| (b, 1))
+            .ok_or(Cow::Borrowed("key buffer empty when reading u8"))
+    }
```
svix-jbrown, line 160: It might be nice for ergonomics to have a newtype here instead of a raw `Cow`.

crates/fjall-utils/src/fjall_key_able.rs
```diff
@@ -0,0 +1,249 @@
...
+                    }
+                    let bytes: [u8; LEN] = buf[..LEN].try_into().expect("checked length");
+                    Ok((<$ty>::from_be_bytes(bytes), LEN))
+                }
+            }
+        )*
+    };
+}
+
+impl_key_component_int!(u16, u32, u64, i8, i16, i32, i64);
```
svix-jbrown: these `i`'s right here give me pause in light of the above

crates/diom-derive/src/fjall_key.rs
tasn, line 1: Zero copy (well, less copies) serialization and deserialization:
1. Deserialization: as [mentioned here](https://github.com/svix/diom-private/pull/899/changes#r3060007195) we only want to get what we want, we don't want to have to copy the key size to access on field, as we almost always just access on field.
2. Serialization: serialization needs to allocate to create the key, but at the moment at least, we also need to allocate to create this pre-structure. Because it takes ownership, so every part would need to be copied and if a String allocated separately.

This code is in the hot-path, and very much so, so we want to establish patterns in the codebase that support performance and minimize copies and allocations.

crates/fjall-utils/src/table_row.rs
```diff
@@ -27,22 +27,29 @@ pub trait TableRow: Sized + Serialize + DeserializeOwned {
             .or_internal_error()
     }
 
-    fn fetch<K: ReadableKeyspace>(keyspace: &K, key: TableKey<Self>) -> Result<Option<Self>> {
+    fn fetch<K: ReadableKeyspace>(
+        keyspace: &K,
+        key: impl Into<TableKey<Self>>,
+    ) -> Result<Option<Self>> {
```
svix-jbrown: having a named type parameter (`T`) and also an anonymous type parameter (`impl Into<...>`) in the same method is kind of a code smell and can make for some funky errors. I would consider rewriting this (and all of the others) as

```rust
fn fetch<K, Y>(keyspace: &K, key: Y) -> Result<Option<Self>>
   where K: ReadableKeyspace,
         Y: Into<TableKey<Self>> {
   // ...
```

It also makes me wonder if we actually want to be taking references to the key type everywhere instead of structs...

crates/msgs/src/entities.rs
```diff
@@ -39,6 +39,23 @@ impl Partition {
...
 
+impl fjall_utils::KeyComponent for Partition {
+    fn key_len(&self) -> usize {
+        size_of::<Self>()
+    }
+
+    fn write_to_key(&self, buf: &mut [u8]) -> usize {
+        let bytes = self.0.to_be_bytes();
+        buf[..bytes.len()].copy_from_slice(&bytes);
+        bytes.len()
```
svix-jbrown: this can just delegate too, right?
```suggestion
        self.0.write_to_key(buf)
```

crates/diom-derive/src/fjall_key.rs
```diff
@@ -0,0 +1,269 @@
...
+            )));
+        }
+
+        ::std::result::Result::Ok(Self { #(#field_constructs),* })
+    }
+}
+
+pub(crate) fn derive(input: TokenStream) -> Result<TokenStream, syn::Error> {
+    let input: DeriveInput = syn::parse2(input)?;
+    let prefix = parse_prefix(&input)?;
```
svix-jbrown, line 181: it would be cool if we also built a global inventory of these and then could have a validator that makes sure they're globally unique. [inventory-rs](https://docs.rs/inventory/latest/inventory/) is a crate I was looking at for doing this in the context of svix/rfc#54.

crates/diom-derive/src/fjall_key.rs
tasn: I wonder if we should limit to only use unsigned numbers for now (at least in the serialization) because you do want lexographical ordering to match other it's hell because you essentially have two separate ordered halves in a range.

crates/diom-derive/src/fjall_key.rs
svix-gabriel: Sure. Here's the output on the main branch

```
╭──────────────────────────────────────────┬─────────────┬─────────┬────────┬───────┬───────┬───────┬────────┬────────┬────────┬────────┬───────────┬──────────────╮
│ op                                       ┆ concurrency ┆ ops/sec ┆ mean   ┆ ±     ┆ min   ┆ p50   ┆ p75    ┆ p99    ┆ p99.9  ┆ max    ┆ bytes/sec ┆ entities/sec │
╞══════════════════════════════════════════╪═════════════╪═════════╪════════╪═══════╪═══════╪═══════╪════════╪════════╪════════╪════════╪═══════════╪══════════════╡
│ msgs.queue.publish (batch=10)            ┆ 20          ┆ 962     ┆ 10.4ms ┆ 5.3ms ┆ 2.6ms ┆ 9.0ms ┆ 12.6ms ┆ 30.2ms ┆ 43.6ms ┆ 56.1ms ┆ 27.27 MB  ┆ 9620         │
│ msgs.queue.receive (batch=10) - ack      ┆ 20          ┆ 1373    ┆ 7.3ms  ┆ 2.1ms ┆ 2.4ms ┆ 7.1ms ┆ 8.4ms  ┆ 12.0ms ┆ 21.1ms ┆ 55.6ms ┆ --        ┆ 13730        │
│ msgs.queue.receive (batch=10) - receive  ┆ 20          ┆ 1310    ┆ 7.6ms  ┆ 2.0ms ┆ 2.5ms ┆ 7.5ms ┆ 8.8ms  ┆ 12.6ms ┆ 18.4ms ┆ 55.8ms ┆ 37.13 MB  ┆ 13100        │
│ msgs.stream.publish (batch=10)           ┆ 20          ┆ 874     ┆ 11.4ms ┆ 6.3ms ┆ 3.3ms ┆ 9.9ms ┆ 14.1ms ┆ 35.4ms ┆ 61.3ms ┆ 91.0ms ┆ 24.78 MB  ┆ 8740         │
│ msgs.stream.receive (batch=10) - commit  ┆ 20          ┆ 1902    ┆ 5.3ms  ┆ 1.2ms ┆ 2.3ms ┆ 5.0ms ┆ 5.8ms  ┆ 9.2ms  ┆ 11.1ms ┆ 15.2ms ┆ --        ┆ 19020        │
│ msgs.stream.receive (batch=10) - receive ┆ 20          ┆ 1861    ┆ 5.4ms  ┆ 1.2ms ┆ 2.5ms ┆ 5.1ms ┆ 5.9ms  ┆ 9.3ms  ┆ 11.5ms ┆ 15.1ms ┆ 52.76 MB  ┆ 18610        │
│ (COMBINED)                               ┆             ┆ 1322.6  ┆ 7.6ms  ┆       ┆ 2.3ms ┆       ┆        ┆        ┆        ┆ 91.0ms ┆ --        ┆              │
```

and output on this PR's branch
```
╭──────────────────────────────────────────┬─────────────┬─────────┬────────┬───────┬───────┬───────┬────────┬────────┬────────┬────────┬───────────┬──────────────╮
│ op                                       ┆ concurrency ┆ ops/sec ┆ mean   ┆ ±     ┆ min   ┆ p50   ┆ p75    ┆ p99    ┆ p99.9  ┆ max    ┆ bytes/sec ┆ entities/sec │
╞══════════════════════════════════════════╪═════════════╪═════════╪════════╪═══════╪═══════╪═══════╪════════╪════════╪════════╪════════╪═══════════╪══════════════╡
│ msgs.queue.publish (batch=10)            ┆ 20          ┆ 904     ┆ 11.1ms ┆ 5.7ms ┆ 3.0ms ┆ 9.6ms ┆ 13.6ms ┆ 33.4ms ┆ 47.1ms ┆ 57.9ms ┆ 25.64 MB  ┆ 9040         │
│ msgs.queue.receive (batch=10) - ack      ┆ 20          ┆ 1408    ┆ 7.1ms  ┆ 2.0ms ┆ 2.2ms ┆ 6.9ms ┆ 8.3ms  ┆ 11.9ms ┆ 19.2ms ┆ 43.6ms ┆ --        ┆ 14080        │
│ msgs.queue.receive (batch=10) - receive  ┆ 20          ┆ 1350    ┆ 7.4ms  ┆ 2.1ms ┆ 2.4ms ┆ 7.2ms ┆ 8.6ms  ┆ 12.3ms ┆ 20.3ms ┆ 43.5ms ┆ 38.27 MB  ┆ 13500        │
│ msgs.stream.publish (batch=10)           ┆ 20          ┆ 1001    ┆ 10.0ms ┆ 5.5ms ┆ 2.8ms ┆ 8.5ms ┆ 12.1ms ┆ 32.1ms ┆ 52.4ms ┆ 89.5ms ┆ 28.38 MB  ┆ 10010        │
│ msgs.stream.receive (batch=10) - commit  ┆ 20          ┆ 1852    ┆ 5.4ms  ┆ 1.3ms ┆ 1.9ms ┆ 5.1ms ┆ 6.3ms  ┆ 9.1ms  ┆ 10.8ms ┆ 29.1ms ┆ --        ┆ 18520        │
│ msgs.stream.receive (batch=10) - receive ┆ 20          ┆ 1800    ┆ 5.6ms  ┆ 1.3ms ┆ 2.0ms ┆ 5.2ms ┆ 6.4ms  ┆ 9.4ms  ┆ 11.0ms ┆ 29.2ms ┆ 51.04 MB  ┆ 18000        │
│ (COMBINED)                               ┆             ┆ 1337.9  ┆ 7.5ms  ┆       ┆ 1.9ms ┆       ┆        ┆        ┆        ┆ 89.5ms ┆ --        ┆              │
╰──────────────────────────────────────────┴─────────────┴─────────┴────────┴───────┴───────┴───────┴────────┴────────┴────────┴────────┴───────────┴──────────────╯
```

crates/diom-derive/src/fjall_key.rs
svix-gabriel: It wasn't really any added complexity to support the non-composite key, imo.

crates/diom-derive/src/fjall_key.rs
svix-gabriel: Ok, so we can do that, it just means changing where the null delimiter is inserted.

1. Right now the null delimiter is inserted by `FjallKeyAble::fjall_key` (which is generated from the macro). https://github.com/svix/diom-private/pull/899/changes#diff-d747a6872fe260c7f372a1c448d3faae2fe8cef7a0672cab6373be0213b494c8R125

2. We could instead optionally insert the null delimiter in the `KeyComponent::write_to_key` function, which is handwritten for each of the native types (u8, u16, String, etc). So String::write_to_key would insert a null delimiter at the end, but u16::write_to_key, u32::write_to_key, etc, wouldn't write the null byte.

I think 1 is a bit simpler, and less error prone, since it means implementing `KeyComponent::write_to_key` for another type _later_ doesn't involve thinking about null delimiters. As the code is written now, the null delimiters are something you don't have to think about when wrapping your type in `FjallKeyAble`, or implementing `KeyComponent::write_to_key`.

Or we can just restrict dynamically sized fields to only be the last key entry. I wanted to avoid it since it seemed unnecessarily restrictive, and I really really _really_ think the "overhead" from these null delimiters is so incredibly trivial.

svix-gabriel: Sure, my point is that parsing the whole key is so incredibly cheap and trivial already that I don't think it's a worthwhile concern. At least for Keys that don't contain a `String`.

But anyhow, nothing about the current implementation precludes what you're asking for. We can have the `#[derive(FjallKeyAble)]` macro also generate functions like `extract_timestamp`, `extract_id`, etc (for structs with an `timestamp` or `id` field). The string prefix and variable sized key fields don't prevent that.

crates/msgs/src/tables.rs
svix-gabriel: I really dislike the RowType pattern. We're saving what, 3-6 bytes of data per key at most? And we have to go through the effort of defining separate type and wiring it up with the key. I don't think that tradeoff is worth it.

But yes, we can force the prefix to be a single byte sized enum if we absolutely must. I think it makes the code more confusing/complex, for an incredibly trivial, if even measurable, performance benefit.

crates/diom-derive/src/fjall_key.rs
svix-gabriel: I addressed this here https://github.com/svix/diom-private/pull/899#discussion_r3060592537, so I'm gonna resolve this thread so we keep the discussion on that thread

crates/diom-derive/src/fjall_key.rs
svix-gabriel: I think that tradeoff is worth it.

I've dealt with macros that didn't use the fully qualified path names, and it can be very annoying when you run into those name collisions. I'd much rather optimize for using the macro rather than reading the macro's implementation.

crates/fjall-utils/src/fjall_key_able.rs
svix-gabriel: Yeah, fair. We can drop the signed ints

crates/msgs/src/tables.rs
tasn: It adds up! 3-6 bytes can definitely make a difference in the hot path. We are working on zero-copy everywhere else for this reason, these bytes add up.

crates/diom-derive/src/fjall_key.rs
tasn: Nothing is cheap in the hot path, and many of our keys contain strings.

Yeah, I know nothing precludes it from adding it in the future, but you're not just doing an internal implementation here, you're defining the patterns of how people build modules internally. I'm saying that we go as far as not having a way to do key -> struct. We don't want to encourage this behavior, and I can't think of anywhere where this would be useful (and if it ever is, we can add it).

crates/diom-derive/src/fjall_key.rs
```diff
@@ -0,0 +1,329 @@
+use proc_macro2::TokenStream;
```
tasn, line 1: Does the duplicate prefix detection work if I split the definition to different modules? Or do they all have to be in the same file/module?

svix-gabriel: I actually had to undo it. It worked at link time, but would break in CI. Beyond that I couldn't find a way to get the automatic prefix validation to work.

So now it's just a manual macro ☹️

tasn: Manual macro as in keeping the existing RowType?

crates/idempotency/src/storage.rs
```diff
@@ -20,6 +21,13 @@ impl From<IdempotencyState> for ByteString {
     }
 }
 
+#[derive(FjallKeyAble)]
```
tasn: What's this? We don't have storage in idempotency. Or is it just a test and the PR is not ready?

svix-gabriel: Whoops, it was mistakenly committed. I was trying to test cross crate definitions with the same prefix byte

crates/diom-derive/src/fjall_key.rs
svix-gabriel: Sorta. It's this https://github.com/svix/diom-private/pull/899/changes#diff-d80920f2294f35878943ad4c6a438c476881d5e6b9b3273c1ea58d849cf00033R12, basically.

So we are removing the RowType once we convert all rows, but still have something kinda manual to do.

tasn: I'm not a fan of rowtype, and was excited when I thought you could potentially fix it, but this feels worse than rowtype as it's just implementing the compiler check ourselves in an easy to miss manner (whoops, forgot to add it to that macro invocation). RowType forces you to define it in one place, and ensures the type is correct.

crates/msgs/src/tables.rs
```diff
@@ -265,6 +264,17 @@ impl TableRow for QueueConfigRow {
     const ROW_TYPE: u8 = RowType::QueueConfig as u8;
 }
 
+#[derive(FjallKeyAble)]
+#[table_key(prefix = 3)]
+pub(crate) struct MsgKey {
+    #[key(0)]
+    pub(crate) topic_id: TopicId,
```
tasn, line 269: Per https://github.com/svix/diom-private/pull/899/changes#r3060114505 let's make it:

```suggestion
    pub(crate) topic_id: &TopicId,
```

crates/diom-derive/src/fjall_key.rs
svix-gabriel: I personally don't really see the difference between

```
fjall_utils::assert_unique_fjall_prefixes!(StructOne, StructTwo...);
```

vs

```
#[repr(u8)]
enum RowType {
    One = 0,
    Two = 1,
    ....
}
```

In terms of "defining things ine one place".

Either way I couldn't figure out how to get the automatic prefix checking. I'd rather not block this PR on that functionality if all else is good

crates/msgs/src/entities.rs
```diff
@@ -24,6 +24,31 @@ pub const TOPIC_PARTITION_DELIMITER: &str = "~";
 #[serde(transparent)]
 pub struct Partition(u16);
 
+impl fjall_utils::KeyComponent for Partition {
```
tasn: Why do we need this? Is it because we wrap around u16? Is it something we can make easier (so it's a one-liner) instead of this? As we use a lot of newtypes everywhere.

crates/fjall-utils/src/fjall_key_able.rs
```diff
@@ -0,0 +1,414 @@
...
+    fn range_inclusive(start: Self, end: Self) -> RangeInclusive<crate::UserKey>;
+    fn range_end_inclusive(end: Self) -> RangeToInclusive<crate::UserKey>;
+}
+
+/// Asserts at compile time that all listed `FjallKeyAble` types have distinct
+/// prefix values. Place this in any module that defines multiple key types
+/// sharing a keyspace.
+///
+/// ```ignore
+/// assert_unique_fjall_prefixes!(MsgKey, CursorKey, ConfigKey);
```
svix-jbrown: this seems like an easy thing to screw up; I wonder if there's something we could do to make it more local (e.g., pass the keyspace name as an argument to the macro, and then automatically generate one of these for each keyspace name)?

crates/fjall-utils/src/fjall_key_able.rs
```diff
@@ -0,0 +1,414 @@
...
+    #[derive(FjallKeyAble)]
+    #[table_key(prefix = 2)]
+    struct ExampleCompositeKey {
+        #[key(0)]
+        id: u32,
+        #[key(1)]
+        group: String,
+    }
+
+    crate::assert_unique_fjall_prefixes!(ExampleSingleKey, ExampleCompositeKey);
```
svix-jbrown: there should probably be a negative test for this too

crates/diom-derive/src/key_component.rs
```diff
@@ -0,0 +1,61 @@
...
+
+            fn write_to_key(&self, buf: &mut [u8]) -> usize {
+                ::fjall_utils::KeyComponent::write_to_key(&self.0, buf)
+            }
+
+            fn read_from_key(
+                buf: &[u8],
+            ) -> ::std::result::Result<(Self, usize), ::std::borrow::Cow<'static, str>> {
+                let (val, len) = <#inner_ty as ::fjall_utils::KeyComponent>::read_from_key(buf)?;
+                ::std::result::Result::Ok((Self(val), len))
```
svix-jbrown, line 51:
```suggestion
                <#inner_ty as ::fjall_utils::KeyComponent>::read_from_key(buf)
```

minor code golf

crates/diom-derive/src/fjall_key.rs
```diff
@@ -0,0 +1,314 @@
...
+    let field_refs: Vec<TokenStream> = fields.iter().map(&field_expr).collect();
+
+    quote! {
+        let total_len = 1 #(+ ::fjall_utils::KeyComponent::key_len(&#field_refs))*;
+
+        // Use a stack buffer for small keys to avoid heap allocation.
+        // Most keys are well under 64 bytes (prefix + a few fixed-size fields).
+        let mut stack_buf = [0u8; 64];
+        let mut heap_buf;
+        let buf: &mut [u8] = if total_len <= 64 {
```
svix-jbrown:
```suggestion
        let buf: &mut [u8] = if total_len <= stack_buf.len() {
```

rustc should be able to constant-propagate this and remove the duplication a bit

crates/fjall-utils/src/fjall_key_able.rs
```diff
@@ -0,0 +1,414 @@
...
+    /// The prefix byte that identifies this key type in the binary layout.
+    const PREFIX: u8;
+
+    fn fjall_key(&self) -> crate::UserKey;
+
+    fn from_fjall_key(key: crate::UserKey) -> Result<Self, Cow<'static, str>>;
+
+    fn range(start: Self, end: Self) -> Range<crate::UserKey>;
+    fn range_inclusive(start: Self, end: Self) -> RangeInclusive<crate::UserKey>;
+    fn range_end_inclusive(end: Self) -> RangeToInclusive<crate::UserKey>;
```
svix-jbrown: any reason not to collapse these down to

```
fn range<RB: std::ops::RangeBounds<Self>>(bounds: RB) -> Range<crate::UserKey>
```

Then people could do somewhat intuitive things like

```
let range1 = keyspace.range(Key { value: 1 }..);
let range2 = keyspace.range(Key { value: 1}..=Key{value: 2});
```

Seems somewhat more idiomatic

svix-gabriel: No reason, that's much cleaner.

fixed

crates/fjall-utils/src/fjall_key_able.rs
svix-gabriel: Yeah I forgot to delete this, but Tom separately convinced me to stick w/the RowType enum earlier

crates/msgs/src/entities.rs
svix-gabriel: derive for KeyComponent is now in

crates/diom-derive/src/fjall_key.rs
svix-gabriel: Changed to use RowType after synchronous discussion

tasn (approved):

PR #900, opened 2026-04-08
cluster: support autoscaling the log fsync interval

svix-jbrown:
I'd worked on this way back around #314; figured I'd finish it out.

This automatically adjusts the log fsync interval based on observed times. It also removes the (imo useless) fsync-per-commit code and updates some comments.


tasn (dismissed):

svix-jbrown:
rebased after landing #913

tasn (approved):

PR #901, opened 2026-04-08
remove trace logging of request payloads

svix-jbrown:
we statically disable `tracing::trace!` in release builds, but this will make debug builds do less work

### DIFF
--- a/src/core/cluster/state_machine.rs
+++ b/src/core/cluster/state_machine.rs
@@ -473,7 +473,7 @@ impl Store {
                     Response::Blank
                 }
                 EntryPayload::Normal(req) => {
-                    tracing::trace!(log_id=?item.log_id, request=?req, "applying user request");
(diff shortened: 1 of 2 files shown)


src/core/cluster/state_machine.rs
```diff
@@ -473,7 +473,7 @@ impl Store {
                     Response::Blank
                 }
                 EntryPayload::Normal(req) => {
-                    tracing::trace!(log_id=?item.log_id, request=?req, "applying user request");
+                    tracing::trace!(log_id=?item.log_id, request=%req, "applying user request");
```
svix-jplatte, line 476: This PR doesn't seem to do what the PR description says.

svix-jbrown: the Display impl for req doesn't actually log the payload, whereas the debug impl ends up logging several kilobytes of `Vec<u8>` payload

tasn (approved):

PR #906, opened 2026-04-09
Replace Vec<u8> in models and API with custom ByteString type

svix-jplatte:
… that is cheaper to clone. Also lets us control the JSON Schema for these, though actually doing that is left for another PR.

### DIFF
--- a/crates/cache/src/lib.rs
+++ b/crates/cache/src/lib.rs
@@ -16,9 +16,6 @@ use coyote_namespace::{Namespace, entities::CacheConfig};
 use coyote_operations::{BackgroundError, BackgroundResult};
 use fjall_utils::Databases;
 use jiff::Timestamp;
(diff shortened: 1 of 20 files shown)


crates/cache/src/lib.rs
```diff
@@ -41,13 +38,6 @@ impl State {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Validate, JsonSchema)]
-pub struct CacheModel {
```
tasn, line 45: Why removed?

tasn (approved):

crates/cache/src/lib.rs
svix-jplatte: Unused. Cache uses `KvModel`.

PR #912, opened 2026-04-09
Simplify OpenAPI span attribute stuff

svix-jplatte:
I noticed this when looking at profiling data.
Should be a minimal perf improvement, but also just cleaner.

### DIFF
--- a/src/core/cluster/applier.rs
+++ b/src/core/cluster/applier.rs
@@ -6,7 +6,7 @@ use super::{
     state_machine::{Store, Stores},
 };
 use crate::AppState;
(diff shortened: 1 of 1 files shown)


src/core/cluster/handle.rs
```diff
@@ -147,6 +149,12 @@ impl RequestWithContext {
     }
 }
```
svix-jbrown, line 144: shouldn't we just use this existing method instead of making a new one?

svix-jplatte: Oh, lol

tasn (approved):

PR #914, opened 2026-04-09
Raft logs: change log cache to veqdeque from btreemap

tasn:
This should be faster as it only allocates once ahead of time, is a just a buffer, and it also allows us to truncate in one pass instead of two.

With that being said, it's hard to see specific results in our benchmarks.

### DIFF
--- a/src/core/cluster/logs.rs
+++ b/src/core/cluster/logs.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::BTreeMap,
+    collections::VecDeque,
     fmt::Debug,
     ops::{Bound, RangeBounds},
     sync::Arc,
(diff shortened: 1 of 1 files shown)


src/core/cluster/logs.rs
```diff
@@ -33,55 +33,65 @@ use coyote_error::Result;
...
+            self.base_index = self.inner.front().map(|e| e.log_id.index);
         }
-        self.inner.insert(entry.log_id.index, entry);
+        if self.base_index.is_none() {
+            self.base_index = Some(entry.log_id.index);
+        }
+        self.inner.push_back(entry);
     }
 
     fn purge(&mut self, log_index: u64) {
```
tasn, line 66: I kept this and truncate simple, I could have potentially just checked the item and then do math to truncate a range in one go (much more efficient). I just wanted to make sure that we are certain we can never have holes and numbers are always consecutive?

src/core/cluster/logs.rs
```diff
@@ -33,55 +33,65 @@ use coyote_error::Result;
...
         }
     }
 
     fn get(&self, log_index: &u64) -> Option<&LogEntry> {
-        self.inner.get(log_index)
+        let base = self.base_index?;
+        let offset = log_index.checked_sub(base)? as usize;
+        self.inner
+            .get(offset)
+            .filter(|e| e.log_id.index == *log_index)
```
tasn: This is a bit of a sanity check, but tbh, it should be removed. Related to https://github.com/svix/diom-private/pull/914/changes#r3060218986

src/core/cluster/logs.rs
```diff
@@ -33,55 +33,65 @@ use coyote_error::Result;
...
-            .collect::<Vec<_>>();
-        for key in keys {
-            self.inner.remove(&key);
+        while let Some(front) = self.inner.front() {
+            if front.log_id.index <= log_index {
+                self.inner.pop_front();
+            } else {
+                break;
+            }
         }
```
svix-jbrown: this can be chained
```suggestion
        while let Some(front) = self.inner.front() && front.log_id.index <= log_index {
          self.inner.pop_front();
        }
}
```

src/core/cluster/logs.rs
svix-jbrown: I would probably turn it into a `debug_assert!` instead; silently returning None just masks any errors

svix-jbrown (commented):
I would expect the savings here to be on the order of single-digit nanoseconds, so not particularly relevant, but whateavs

tasn:
The reason for this PR is not performance as much as reducing allocator pressure (which leads to performance).

svix-jbrown (approved):

PR #922, opened 2026-04-10
switch interserver protocol to use postcard

svix-jbrown:
FWIW, I don't think the proposal in svix/rfc#54 will work very well for interserver messages (since they contain the union of all modules' operation messages). Actually doing this correctly will be rather hard (because, e.g., the "version" of an AppendEntries message is going to be the union of all of the versions of the underlying requests). Right now, this will just make the cluster break extremely un-gracefully if we ever change any operation definition. It's also not meaningfully faster, since the actual network time dominates the msgpack time. Anywho, here it is as requested.

### DIFF
--- a/crates/diom-proto/src/internal_client.rs
+++ b/crates/diom-proto/src/internal_client.rs
@@ -6,7 +6,7 @@ use serde::{Serialize, de::DeserializeOwned};
 use tap::Tap;
 use tokio::sync::{mpsc, oneshot};
 
-const APPLICATION_MSGPACK: HeaderValue = HeaderValue::from_static("application/msgpack");
+use super::msgpack_client::APPLICATION_MSGPACK;
 
 /// API client for interacting with the internal "loopback" API server.
 #[derive(Clone)]
--- a/crates/diom-proto/src/postcard_.rs
+++ b/crates/diom-proto/src/postcard_.rs
@@ -1,0 +1,201 @@
+use axum::{
+    extract::{FromRequest, OptionalFromRequest, Request, rejection::BytesRejection},
+    response::{IntoResponse, Response},
+};
+use bytes::{Bytes, BytesMut};
+use http::{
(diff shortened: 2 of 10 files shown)


crates/diom-proto/src/internal_client.rs
tasn, line 1: I know we discussed on the call and said that since we serialize the operations to log, and we pass logs here, the concerns raised by this PR are already the case, no? (we already have the danger around operations).
Maybe worth updating the PR description with this information?

crates/diom-proto/src/postcard_.rs
tasn, line 1: Maybe a better name than `postcard_`? (as in, the trailing _)

src/core/cluster/network.rs
```diff
@@ -239,8 +327,11 @@ impl RaftNetworkV2<TypeConfig> for NetworkClient {
         rpc: openraft::raft::VoteRequest<TypeConfig>,
         option: RPCOption,
     ) -> Result<openraft::raft::VoteResponse<TypeConfig>, RPCError> {
-        self.send_request_with_timeout("/repl/raft/vote", rpc, option.soft_ttl())
-            .await
+        let req = self
+            .build_request("/repl/raft/vote", rpc)
+            .with_timeout(option.soft_ttl())
+            .with_format(Format::Msgpack);
```
tasn, line 333: Why is this one using msgpack and others postcard?

svix-jbrown: all the things that are low-volume use msgpack (it's the default; this one just calls it out explicitly because it's also overriding the TTL)

svix-jplatte:
We've run the experiment and learned what we wanted to learn, right? Can be closed?

svix-jbrown:
I'm going to punt on this for now because I don't think it's worth it yet (and it's a breaking change)

PR #923, opened 2026-04-10
Update TS SDK to check new duration and timestamp types

svix-pluiz:
I was going to do all SDKs in a single PR but this seems worse to review..

Part of https://github.com/svix/diom-private/issues/809


z-clients/javascript/src/models/idempotencyCompleteIn.ts
```diff
@@ -7,7 +7,7 @@ export interface IdempotencyCompleteIn {
     /** Optional metadata to store alongside the response */
     context?: { [key: string]: string } | null;
     /** How long to keep the idempotency response for. */
-    ttlMs: number;
```
svix-lucho, line 10: I think this should be a Duration, not a Date, no?

svix-pluiz: @svix-lucho The problem with `Duration` in typescript is that's the Temporal.Duration API was added literally in the [last month](https://bloomberg.github.io/js-blog/post/temporal/), so we would need to use either a polyfill to the API or a date library

svix-lucho: But what does `Date` mean in this context? it has a completely different meaning compared to `Duration`

svix-pluiz: It's the TTL set as a Date, like I updated the test here https://github.com/svix/diom-private/pull/923/changes#diff-56bd4b6e4ecf3c98d9cde5b835d7ea87eec43df7b03b8f9fdabff2d591206d83R54

svix-pluiz: I was checking the 'official' polyfill for Temporal API and it works with Node 14 and newer versions
https://www.npmjs.com/package/@js-temporal/polyfill

z-clients/javascript/tests/integration.test.mjs
```diff
@@ -64,6 +67,6 @@ describe("SDK Integration Tests", () => {
 
     // Verify deleted
     const getResp2 = await client.cache.get(key, {});
-    assert.strictEqual(getResp2.value, null);
+    assert.strictEqual(getResp2.value, undefined);
```
svix-jplatte: Why does this return undefined now? That seems unrelated to the purpose of this PR.

svix-pluiz: Different parts of the SDK were returning null or undefined, I noticed that and changed to return undefined by default

svix-jplatte: Hm, okay. It seems more idiomatic to me for this to return `null` though, no?

svix-pluiz: Hmm, I checked some DB libraries and found 3 using null and 2 using undefined,
Prisma, TypeORM and MikroORM are using `T | null`, Drizzle and Kysely are using `T | undefined`

svix-pluiz: for major API SDKs, 4 of them returning null and 2 returning undefined

svix-pluiz: I'll revert this change, just a moment

svix-jplatte (approved):

PR #924, opened 2026-04-10
Increase time tolerance for test_kv_set_and_get test

svix-pluiz:


tests/it/kv.rs
```diff
@@ -111,7 +111,7 @@ async fn test_kv_set_and_get() -> TestResult {
         expires_at
             .as_millisecond()
             .abs_diff(expected.as_millisecond())
-            < 50 // tolerance
+            < 100 // tolerance
```
svix-james, line 114: Is it flaky, or does it just always fail now? I'm just surprised that w/ all the changes recently this test would have gotten flakier.

svix-pluiz: It's flaky, check this thread here https://svixhq.slack.com/archives/C0A72988962/p1775828665189449

svix-jplatte (approved):

PR #925, opened 2026-04-10
fjall serialization: serialize postcard directly to byteview

tasn:
Byteview is used internally by fjall, so this makes it what fjall expects instead of the extra allocations and copies.

Before this change:
1. Alloc a Vec<u8>
2. Serialize it Vec<u8> (increases the size as needed).
3. Alloc a byteview.
4. Copy the Vec to the byteview.

After this change:
1. Measure the size of the object (fast).
2. Alloc a byteview of the right size.
2. Serialize to the byteview.

Benchmarks (in the PR) show ~2.5x improvement across all payload sizes; which adds up to multiple milliseconds of execution per second at thousands of ops per second and also decreases allocator pressure.

### DIFF
--- a/crates/fjall-utils/src/lib.rs
+++ b/crates/fjall-utils/src/lib.rs
@@ -23,6 +23,18 @@ pub enum V0Wrapper<T> {
     V0(T),
 }
 
+/// Serialize `value` directly into a [`byteview::ByteView`],
+/// avoiding an intermediate `Vec<u8>` allocation.
+pub(crate) fn postcard_to_byteview(
+    value: &impl serde::Serialize,
+) -> Result<byteview::ByteView, postcard::Error> {
+    use postcard::ser_flavors;
+    let size = postcard::serialize_with_flavor(value, ser_flavors::Size::default())?;
+    let mut builder = byteview::ByteView::builder(size);
+    postcard::serialize_with_flavor(value, ser_flavors::Slice::new(&mut builder))?;
+    Ok(builder.freeze())
+}
+
 /// Useful for verifying all table prefixes for a given keyspace are unique,
 /// at compile time.
 pub const fn are_all_unique(strings: &[&str]) -> bool {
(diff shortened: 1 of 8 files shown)


crates/fjall-utils/src/lib.rs
```diff
@@ -23,6 +23,18 @@ pub enum V0Wrapper<T> {
...
 
+/// Serialize `value` directly into a [`byteview::ByteView`],
+/// avoiding an intermediate `Vec<u8>` allocation.
+pub(crate) fn postcard_to_byteview(
+    value: &impl serde::Serialize,
+) -> Result<byteview::ByteView, postcard::Error> {
+    use postcard::ser_flavors;
+    let size = postcard::serialize_with_flavor(value, ser_flavors::Size::default())?;
+    let mut builder = byteview::ByteView::builder(size);
+    postcard::serialize_with_flavor(value, ser_flavors::Slice::new(&mut builder))?;
```
svix-jbrown:
```suggestion
    postcard::to_slice(&mut builder)?;
```

can be simplified a bit

crates/fjall-utils/src/lib.rs
```diff
@@ -23,6 +23,18 @@ pub enum V0Wrapper<T> {
...
 }
 
+/// Serialize `value` directly into a [`byteview::ByteView`],
+/// avoiding an intermediate `Vec<u8>` allocation.
+pub(crate) fn postcard_to_byteview(
+    value: &impl serde::Serialize,
+) -> Result<byteview::ByteView, postcard::Error> {
+    use postcard::ser_flavors;
+    let size = postcard::serialize_with_flavor(value, ser_flavors::Size::default())?;
+    let mut builder = byteview::ByteView::builder(size);
```
svix-jbrown, line 33: wow, this made me look at the internals of byteview and holy memory aliasing, batman. Also, I'm 90% sure that that crate has exploitable memory corruption on 32-bit platforms.

svix-jbrown (commented):
seems okay; one cleanup note

crates/fjall-utils/src/lib.rs
tasn: We already use it extensively because of fjall. Though if they made this mistake, I wonder what's up with fjall in general. 😬

svix-jbrown (approved):

PR #934, opened 2026-04-10
Cache: switch to compaction filters for expiry.

tasn:
Cache doesn't need to have strict expiry like kv, which we want to make sure expires exactly when it's supposed to expire as it's stored on persisted disk.

Cache can instead just rely on compaction to do the expiry, similarly to how it's done in #858.

We also made it so CacheGetOut always returns a timestamp (as expiry is no longer optional) - it just returns `now` when the value is not found.


svix-jbrown:
The issue with relying on compaction for expiry is that long-lived items may never be expired, because compaction of old data may never occur. I think it's confusing for our users if the dataset size on disk stays the same until they write enough to trigger a last-level compaction (which I think with the defaults could require a 10x increase in data size if the old data is all already in the last tier).

Other LSM-based systems (including both clickhouse and cassandra) solve this problem by making TTLs a first-class property of the underlying LSM and forcing compactions on chunks which exceed some configurable threshold of expired records, but fjall doesn't have any way to force compaction of specific segments in the LSM. You can force a major compaction of the entire Keyspace, but I don't think it would be good for performance to periodically do that.

Doing expiration through compaction also means that when we bootstrap a new node, we'll copy expired rows (since the snapshot process intentionally doesn't parse the rows or have semantic knowledge)...

tasn:
Summary: you think it's a bad idea and we shouldn't do it?

crates/cache/src/controller.rs
```diff
@@ -0,0 +1,111 @@
...
+use diom_id::NamespaceId;
+use fjall::{Database, Keyspace};
+use fjall_utils::{SerializableKeyspaceCreateOptions, TableRow, WriteBatchExt};
+use jiff::Timestamp;
+
+use crate::{CacheModel, tables::CacheRow};
+
+#[derive(Clone)]
+pub struct CacheController {
+    db: InstrumentedMutex<Database>,
```
svix-jbrown, line 14: why is this still behind a mutex if it's no longer being accessed concurrently?

src/v1/endpoints/cache.rs
```diff
@@ -186,7 +188,7 @@ async fn cache_get(
     let ret = match model {
         Some(m) => CacheGetOut::from_model(m),
         None => CacheGetOut {
-            expiry: None,
+            expiry: repl.time.now(),
```
svix-jbrown, line 191: this feels more confusing to me than just retaining the option...

svix-jbrown (commented):
in addition to the general feedback of why compaction isn't ideal for TTL'd deletes, some specific code feedback:

tasn:
Closing, probably not a good idea.

PR #937, opened 2026-04-10
API: use the repl stores instead of initing new ones ad-hoc.

tasn:
We already have the stores in repl, so we can just use that from the API. Anyway, we don't want to init the State manually every time in the API as it involves creating a keyspace object and potentially initializing other things, which is just wasteful.

### DIFF
--- a/src/core/cluster/state_machine.rs
+++ b/src/core/cluster/state_machine.rs
@@ -903,6 +903,36 @@ impl StoreHandle {
             .clone()
     }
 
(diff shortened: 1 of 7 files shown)


svix-jbrown (approved):
I think this is okay; one thing to note is that because these are going through the rwlock on the state machine, it means that they'll be unavailable briefly while we're taking a snapshot of the database (just for however long it takes the `fjall::Database::snapshot` method to run, which is usually just a couple of milliseconds)

PR #940, opened 2026-04-10
remove some #[serde] attributes

svix-jbrown:
I have a good feeling about this one


z-clients/rust/src/models/queue_msg_out.rs
```diff
@@ -8,8 +8,7 @@ pub struct QueueMsgOut {
     #[serde(with = "serde_bytes")]
```
tasn, line 8: Probably shouldn't touch the libs/API?

tasn (approved):

PR #941, opened 2026-04-10
move background cleanup back to being a leader task again

svix-jbrown:
effectively undoes #492

I don't think this makes any difference, but whatever.

this was originally #903 but it got screwed up by the history rewrite

### DIFF
--- a/crates/diom-operations/src/lib.rs
+++ b/crates/diom-operations/src/lib.rs
@@ -46,15 +46,17 @@ impl std::error::Error for BackgroundError {
 pub type BackgroundResult<T> = Result<T, BackgroundError>;
 
 // Not part of the trait below so do_write_request only has to be codegen'ed once
-pub trait OperationWriterBase {
+pub trait OperationWriterBase: Send + Sync + Clone + 'static {
     type Request: Sized;
     type Response: Sized;
 
     /// Execute a write against the replicated state machine
     ///
     /// This should typically not be called by users
-    #[allow(async_fn_in_trait)]
-    async fn do_write_request(&self, request: Self::Request) -> BackgroundResult<Self::Response>;
+    fn do_write_request(
+        &self,
+        request: Self::Request,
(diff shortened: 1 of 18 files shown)


z-clients/cli/src/cmds/benchmark.rs
```diff
@@ -595,7 +595,14 @@ impl BenchShard for BenchKvSet {
...
         let t = Instant::now();
-        client.kv().set(key.clone(), value, KvSetIn::new()).await?;
+        client
+            .kv()
+            .set(
+                key.clone(),
+                value,
+                KvSetIn::new().with_ttl(Duration::from_millis(500)),
+            )
+            .await?;
```
svix-jbrown, line 614: this makes the benchmark set TTLs so we can actually exercise this code (and make the benchmark mildly more interesting)

crates/diom-operations/src/lib.rs
```diff
@@ -46,15 +46,17 @@ impl std::error::Error for BackgroundError {
...
 // Not part of the trait below so do_write_request only has to be codegen'ed once
-pub trait OperationWriterBase {
+pub trait OperationWriterBase: Send + Sync + Clone + 'static {
     type Request: Sized;
     type Response: Sized;
 
     /// Execute a write against the replicated state machine
     ///
     /// This should typically not be called by users
-    #[allow(async_fn_in_trait)]
```
svix-gabriel, line 56: Is there a reason other than appeasing the clippy gods for removing allow(async_fn_in_trait)?

From a readability standpoint I think this hurts more than helps

svix-gabriel (dismissed):

crates/diom-operations/src/lib.rs
svix-jbrown: Unfortunately, yes, it doesn't compile with `async fn in trait` because it fails to infer the `Send` bound on the returned future.

svix-gabriel (approved):

PR #942, opened 2026-04-13
python: Add type checking

svix-jplatte:
Make sure our types are correct. I wanted to go with `ty`, but it doesn't seem to support pydantic properly at the moment (complains about field initialization, expecting us to use aliases instead). Tried `pyrefly` second, and that works fine.


svix-jplatte:
Weird, it passed locally.

svix-jbrown (approved):

PR #945, opened 2026-04-13
rust: Revamp error types

svix-jplatte:
Part of https://github.com/svix/diom-private/issues/302.

### DIFF
--- a/z-clients/rust/src/error.rs
+++ b/z-clients/rust/src/error.rs
@@ -41,7 +41,7 @@ impl Error {
     }
 
     pub(crate) fn timeout(op_id: &'static str) -> Error {
-        Self::new(op_id, ErrorKind::Timeout)
+        Self::new(op_id, ErrorKind::Timeout(TimeoutError))
     }
 
     pub(crate) fn other(
@@ -74,7 +74,7 @@ impl Error {
 
     #[must_use]
     pub fn is_timeout(&self) -> bool {
-        matches!(self.0.kind, ErrorKind::Timeout)
+        matches!(self.0.kind, ErrorKind::Timeout(_))
     }
 
     #[must_use]
@@ -84,7 +84,7 @@ impl Error {
 
     pub(crate) fn is_retryable(&self) -> bool {
         match self.0.kind {
-            ErrorKind::Network(_) | ErrorKind::Server(_) | ErrorKind::Timeout => true,
+            ErrorKind::Network(_) | ErrorKind::Server(_) | ErrorKind::Timeout(_) => true,
             ErrorKind::Client(_) | ErrorKind::Other(_) => false,
         }
     }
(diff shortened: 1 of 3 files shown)


z-clients/rust/src/error.rs
```diff
@@ -8,42 +8,59 @@ use serde::Deserialize;
...
+
+    #[must_use]
+    pub fn is_other(&self) -> bool {
+        matches!(self, Self::Other(_))
+    }
+
+    pub(crate) fn is_retryable(&self) -> bool {
+        match self {
+            Self::Network(_) | Self::Server(_) | Self::Timeout => true,
+            Self::Client(_) | Self::Other(_) => false,
```
tasn: I don't think if this is true... Client errors can happen because of timing and could be retryable. Validation (which is missing in your code, we should have a type for that) is not retriable though.

svix-jplatte: Can you elaborate? What kind of retrying client errors are you thinking about?

Validation errors are part of the `ClientError` type.

z-clients/rust/src/error.rs
```diff
@@ -56,48 +73,182 @@ impl Error {
...
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Error::Generic(s) => s.fmt(f),
-            Error::Http(e) => format!("Http error (status={}) {:?}", e.status, e.payload).fmt(f),
-            Error::Validation(e) => format!("Validation error {:?}", e.payload).fmt(f),
+            // don't print inner errors that are returned from Error::source
+            Self::Network(_) => write!(f, "network error"),
+            Self::Client(e) => write!(f, "client error: {e}"),
+            Self::Server(e) => write!(f, "server error: {e}"),
+            Self::Timeout => write!(f, "timeout"),
```
svix-jbrown: it would be nice if this variant included the configured timeout (and *which* timeout it was; connect vs data send)

svix-jbrown (commented):
I wonder if it would make sense to wrap all of these in some kind of a parent error that includes context (like the URL being requested)? I dunno, I'm just thinking about how I'd use these to actually debug something and it's pretty tough

z-clients/rust/src/error.rs
tasn: I'm saying that client errors are potentially retryable if the request is non-linearized. Though actually, I guess we can consider this as non-retryable, so yeah, probably fine.

tasn:
> I wonder if it would make sense to wrap all of these in some kind of a parent error that includes context (like the URL being requested)? I dunno, I'm just thinking about how I'd use these to actually debug something and it's pretty tough

Maybe not url, but the op name

z-clients/rust/src/error.rs
svix-jplatte: Currently we appear to only support a single timeout. I've added a non-exhaustive `TimeoutError` type though that we can backwards-compatibly add new stuff to.

svix-jbrown (approved):

PR #947, opened 2026-04-13
Rate-limit: add a fast-path to avoid raft for exhausted limits

tasn:
If a rate-limit is already exhausted (no tokens available) there's no need to spam raft with this request. We can just exit quickly locally. For extra efficiency we don't even need to wait for linearized reads, we can just see what our current version says (means extra noise on raft, but faster to respond in non-late-limited case).

We may want to change that to linearized for slight latency impact on the happy path, but less raft load on the bad path. I decided to not do it for now.

### DIFF
--- a/crates/rate-limit/src/controller.rs
+++ b/crates/rate-limit/src/controller.rs
@@ -74,6 +74,33 @@ impl RateLimitController {
         .await?
     }
 
+    /// Checks whether a request will be denied if called by `limit`.
+    pub async fn check_limit<I: AsRef<str> + 'static + Send>(
+        &self,
+        now: Timestamp,
+        namespace_id: NamespaceId,
+        identifier: I,
+        wanted: u64,
+        config: TokenBucket,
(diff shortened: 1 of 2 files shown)


crates/rate-limit/src/controller.rs
```diff
@@ -74,6 +74,33 @@ impl RateLimitController {
...
+                Ok((
+                    false,
+                    capacity,
+                    Some(config.calculate_retry_after(capacity, wanted)),
+                ))
+            } else {
+                Ok((true, capacity, None))
+            }
+        })
+        .await?
```
svix-lucho: isn't this the same as calling `get_remaining` and checking if `remaining > wanted`?

tasn: No, because `get_remaining` doesn't accept a `wanted` field. Though I can merge them I guess.

tasn: Done.

svix-lucho (approved):

PR #949, opened 2026-04-13
Remove docs and svix-specific deploy configs

svix-james:
Also deploys the on-main Docker publish steps to private repos.

For now, Helm step has not been removed. Publishing it to a private
Helm repository will take a bit more work.


.github/workflows/on-push.yml
```diff
@@ -23,81 +23,3 @@ jobs:
...
-  helm-publish:
-    needs: ci
-    uses: ./.github/workflows/helm-publish.yml
-    secrets: inherit
-    permissions:
-      contents: read
-      packages: write
-    with:
-      version: 0.0.0-${{ github.sha }}
-      image_tag: sha-${{ github.sha }}
```
svix-jbrown, line 76: I would think these would want to stay here...?

svix-jbrown (approved):

svix-jbrown (changes requested):
oops, meant to click the other radio button

.github/workflows/on-push.yml
svix-james: Don't we want these to only occur with our release job going forward?

svix-james: Ok updated. Only removing Helm for now.

svix-jbrown (approved):

PR #950, opened 2026-04-13
Update SDKs package metadata 

svix-pluiz:


svix-jbrown (commented):
I think tom wants to change the URLs to diom.svix.com

svix-pluiz:
> I think tom wants to change the URLs to diom.svix.com

@svix-jbrown @tasn Do you think it makes sense to update the Author/Vendor name to point to diom.svix.com as well?

Example:


```
POM_ORGANIZATION_URL=https://diom.svix.com

VENDOR_NAME=Svix Inc. (https://diom.svix.com)
```

tasn:
Homepage/URLs should defo point to diom.svix.com. Vendor name we can be flexible, but can also put diom.svix.com, why not

tasn (approved):

PR #951, opened 2026-04-13
cfg: teach write-config subcommand to pass through docstrings

### LINKED ISSUE #659: check in (automatically generate?) an example TOML configuration file

svix-jbrown:
fixes #659

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,7 +116,7 @@ jobs:
           comment-id: ${{ steps.fc.outputs.comment-id }}
           edit-mode: replace
   check-codegen:
-    name: Check openapi.json and SDK freshness
+    name: Check openapi.json, SDK, and default config freshness
     runs-on: ubicloud-standard-4-ubuntu-2404
     steps:
       - uses: actions/checkout@v6
@@ -128,6 +128,8 @@ jobs:
           save-if: ${{ github.ref == 'refs/heads/main' }}
       - name: Regerate openapi.json and SDKs
         run: cargo codegen
+      - name: Regerate config.default.toml
+        run: cargo run -- write-config config.defaults.toml
       - name: Check for uncommitted changes
         run: |
           gs=$(git status -s)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,7 +128,7 @@ jobs:
           save-if: ${{ github.ref == 'refs/heads/main' }}
       - name: Regerate openapi.json and SDKs
         run: cargo codegen
-      - name: Regerate config.default.toml
+      - name: Regerate config.defaults.toml
         run: cargo run -- write-config config.defaults.toml
       - name: Check for uncommitted changes
         run: |
--- a/crates/diom-derive/src/dumpable_config.rs
+++ b/crates/diom-derive/src/dumpable_config.rs
@@ -1,5 +1,5 @@
 use quote::quote;
-use syn::{DeriveInput, Ident, LitStr, Token, parse_macro_input};
+use syn::{DeriveInput, Ident, LitStr, Token};
 
 use crate::utils::as_ty_option;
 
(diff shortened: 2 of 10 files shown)


crates/diom-derive/src/dumpable_config.rs
```diff
@@ -0,0 +1,168 @@
...
+    for parsed in fields {
+        let field = parsed.field;
+        let name = parsed.name;
+        let is_optional = parsed.is_optional;
+        let docs = parsed.docstring.iter().map(|d| {
+            if d.trim().is_empty() {
+                quote! { write!(writer, "#\n")?; }
+            } else {
+                quote! { write!(writer, "# {}\n", #d.trim())?;  }
+            }
```
svix-jplatte: You can assemable the string outside `quote!` and interpolate that to not repeat `quote! { write!(writer, `.

crates/diom-derive/src/dumpable_config.rs
```diff
@@ -0,0 +1,168 @@
...
+            } else {
+                quote! {
+                    let serialized = serde::Serialize::serialize(
+                        &self.#field,
+                        toml::ser::ValueSerializer::new(&mut buffer)
+                    )?;
+                    write!(writer, "{} = {}\n", #name, serialized)?;
+                    buffer.clear();
+                }
+            };
```
svix-jplatte, line 137: Significant duplication in quote tokens, can be factored into a separate quote call right?

crates/diom-derive/src/dumpable_config.rs
```diff
@@ -0,0 +1,168 @@
...
+            fn dump_fields<W: std::io::Write>(&self, writer: &mut W, prefix: String) -> ::anyhow::Result<()> {
+                let mut buffer = String::new();
+                #(#lists)*;
+                Ok(())
+            }
+        }
+    };
+
+    // Hand the output tokens back to the compiler.
+    proc_macro::TokenStream::from(expanded)
```
svix-jplatte: Comment really useful? Also the convention is to use proc_macro2 in submodules and do the `proc_macro::TokenStream` conversion in `lib.rs` (usually just a `.into()`). I can do a separate PR for all submodules after this if you don't mind.

src/cfg/mod.rs
```diff
@@ -98,11 +100,12 @@ impl Serialize for PeerAddr {
     }
 }
 
-#[derive(Clone, Debug, Default, Deserialize, EnvOverridable, Serialize)]
+#[derive(Clone, Debug, Default, Deserialize, EnvOverridable, Serialize, DumpableConfig)]
 pub struct DatabaseConfig {
     /// Directory in which this database is stored
     pub path: PathBuf,
-    /// Filename under the directory specified in `path`
+    /// Filename under the directory specified in `path`. If not specified, uses a default value
```
svix-jplatte: Probably better as two paragraphs?

.github/workflows/ci.yml
```diff
@@ -128,6 +128,8 @@ jobs:
           save-if: ${{ github.ref == 'refs/heads/main' }}
       - name: Regerate openapi.json and SDKs
         run: cargo codegen
+      - name: Regerate config.default.toml
```
svix-jplatte:
```suggestion
      - name: Regerate config.defaults.toml
```

crates/diom-derive/src/dumpable_config.rs
svix-jbrown: nah, just copied from the existing one. I'll make this one proc_macro2 for now but won't touch any of the others.

svix-jplatte (approved):
I guess the generated `write!` is a bit weird now, I missed the different number of args. Whatever.

PR #955, opened 2026-04-13
Operator: Support increasing storage size

svix-james:
The first commit is primarily just re-factoring to try and make 
the code more readable, with some better handling for status
reporting.

The second commit actually adds the resize logic:
* Detect updates to storage size and if applicable / supported by
  the underlying storage, directly resize the PVCs. Afterward,
  orphan/delete/re-create the stateful set so that it reflects the
  updated values.

  This currently does _not_ support restarting pods for those
  file systems that require a pod restart. The vast majority of
  storage classes support online resizing, though, so I think this
  is good enough for now.

### DIFF
--- a/infra/operator/src/resources/pvcs.rs
+++ b/infra/operator/src/resources/pvcs.rs
@@ -15,7 +15,8 @@ use crate::{
     error::{Error, Result},
 };
 
-pub(crate) async fn reconcile_pvcs(ctx: &ClusterCtx) -> Result<()> {
+/// Reconcile persistent volume claims.
+pub(crate) async fn reconcile(ctx: &ClusterCtx) -> Result<()> {
     let storage = &ctx.cluster.spec.diom.storage;
     let nodes = ctx.cluster.spec.diom.replicas;
 
@@ -66,7 +67,7 @@ async fn reconcile_volume(
         let name = pvc_name(volume_name, &ctx.name, ordinal);
 
         let Some(pvc) = pvc_api.get_opt(&name).await? else {
-            tracing::debug!(name, "PVC not found; skipping resize.");
+            tracing::warn!(name, "PVC not found; skipping reconcile.");
             continue;
         };
 
@@ -119,7 +120,7 @@ async fn patch_pvc_size(
     });
     tracing::info!(
         name,
-        desired = desired.to_string(),
+        %desired,
         "Patching PVC with new size"
     );
     pvc_api
@@ -158,9 +159,7 @@ async fn storage_class_allows_expansion(
 fn pvc_requested_size(pvc: &PersistentVolumeClaim) -> Result<ParsedQuantity> {
     pvc.spec
         .as_ref()
-        .and_then(|s| s.resources.as_ref())
-        .and_then(|r| r.requests.as_ref())
-        .and_then(|r| r.get("storage"))
+        .and_then(|s| s.resources.as_ref()?.requests.as_ref()?.get("storage"))
         .cloned()
         .ok_or_else(|| Error::Storage("Storage not found".to_owned()))?
         .try_into()
--- a/infra/operator/src/resources/pvcs.rs
+++ b/infra/operator/src/resources/pvcs.rs
@@ -1,0 +1,216 @@
+use std::collections::BTreeMap;
+
+use k8s_openapi::{
+    api::{core::v1::PersistentVolumeClaim, storage::v1::StorageClass},
+    apimachinery::pkg::api::resource::Quantity,
+};
+use kube::{
+    Api,
+    api::{ListParams, Patch, PatchParams},
+};
+use kube_quantity::{ParseQuantityError, ParsedQuantity};
+
+use crate::{
+    context::ClusterCtx,
+    error::{Error, Result},
+};
+
+pub(crate) async fn reconcile_pvcs(ctx: &ClusterCtx) -> Result<()> {
+    let storage = &ctx.cluster.spec.diom.storage;
+    let nodes = ctx.cluster.spec.diom.replicas;
+
+    reconcile_volume(
+        ctx,
+        "persistent",
+        &storage.persistent.size,
+        storage.persistent.storage_class.as_deref(),
+        nodes,
+    )
+    .await?;
+
+    if let Some(logs) = &storage.logs {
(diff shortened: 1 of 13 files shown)


svix-james:
## OSS Review

This adds an additional crate -- [kube_quantity](https://crates.io/crates/kube_quantity) -- that adds a proper parser for k8s quantities. It's a small crate w/ minimal dependencies.

infra/operator/src/resources/services.rs
```diff
@@ -2,38 +2,60 @@ use k8s_openapi::{
...
+            headless.metadata.name.as_deref().unwrap(),
+            &pp,
+            &Patch::Apply(&headless),
+        )
+        .await?;
+
+    let client_svc = build_client_facing(ctx)?;
+    svc_api
+        .patch(
+            client_svc.metadata.name.as_deref().unwrap(),
```
svix-jplatte: I guess this only moved, but the casual unwrap here doesn't seem great to me.

infra/operator/src/resources/statefulset.rs
```diff
@@ -107,18 +116,14 @@ fn build_env(
 
     // These must come before any vars that reference them via $(VAR) substitution.
     let mut env: Vec<EnvVar> = vec![
-        // Downward API: pod name and namespace, used to construct stable DNS addresses.
         env_from_field("POD_NAME", "metadata.name"),
         env_from_field("POD_NAMESPACE", "metadata.namespace"),
-        // Each pod advertises its stable StatefulSet DNS name so peers can reach it.
-        // Uses k8s env var substitution: $(VAR) references earlier vars in this list.
```
svix-jplatte, line 114: Why did you remove the comments here and below?

infra/operator/src/resources/pvcs.rs
```diff
@@ -0,0 +1,216 @@
...
+    api::{ListParams, Patch, PatchParams},
+};
+use kube_quantity::{ParseQuantityError, ParsedQuantity};
+
+use crate::{
+    context::ClusterCtx,
+    error::{Error, Result},
+};
+
+pub(crate) async fn reconcile_pvcs(ctx: &ClusterCtx) -> Result<()> {
```
svix-jplatte: I see above there's a `PersistentVolumeClaim` import so that's probably what `pvc` stands for, but wouldn't hurt to spell it out again in places like this:


```suggestion
/// Reconcile persistent volume claims.
pub(crate) async fn reconcile_pvcs(ctx: &ClusterCtx) -> Result<()> {
```

(or, ideally, something a little more descriptive)

Would make this code more readable for ppl not intimately familiar with K8s.

infra/operator/src/resources/pvcs.rs
```diff
@@ -0,0 +1,216 @@
...
+
+        let desired_size: ParsedQuantity = desired_size
+            .try_into()
+            .map_err(|e: ParseQuantityError| Error::Storage(e.to_string()))?;
+        let current_size = pvc_requested_size(&pvc)?;
+
+        if desired_size <= current_size {
+            if desired_size < current_size {
+                tracing::warn!(name, "Desired size smaller than current; skipping resize.");
+            }
```
svix-jplatte, line 82: Maybe PR title should be "Support storage upsizing" if we don't support downsizing currently.

infra/operator/src/resources/pvcs.rs
```diff
@@ -0,0 +1,216 @@
...
+    storage_class: Option<&str>,
+    nodes: i32,
+) -> Result<()> {
+    let pvc_api: Api<PersistentVolumeClaim> = Api::namespaced(ctx.client.clone(), &ctx.ns);
+
+    for ordinal in 0..nodes {
+        let name = pvc_name(volume_name, &ctx.name, ordinal);
+
+        let Some(pvc) = pvc_api.get_opt(&name).await? else {
+            tracing::debug!(name, "PVC not found; skipping resize.");
```
svix-jplatte: Should this really just be a debug message, not warning or even error? Without having a lot of context, this does sound bad.

infra/operator/src/resources/pvcs.rs
```diff
@@ -0,0 +1,216 @@
...
+    let patch = serde_json::json!({
+        "spec": {
+            "resources": {
+                "requests": requests
+            }
+        }
+    });
+    tracing::info!(
+        name,
+        desired = desired.to_string(),
```
svix-jplatte:
```suggestion
        %desired,
```

infra/operator/src/resources/pvcs.rs
```diff
@@ -0,0 +1,216 @@
...
+        .and_then(|sc| sc.allow_volume_expansion)
+        .unwrap_or(false))
+}
+
+fn pvc_requested_size(pvc: &PersistentVolumeClaim) -> Result<ParsedQuantity> {
+    pvc.spec
+        .as_ref()
+        .and_then(|s| s.resources.as_ref())
+        .and_then(|r| r.requests.as_ref())
+        .and_then(|r| r.get("storage"))
```
svix-jplatte:
```suggestion
        .and_then(|s| s.resources.as_ref()?.requests.as_ref()?.get("storage"))
```

svix-jplatte (dismissed):

infra/operator/src/resources/statefulset.rs
svix-james: A lot of this info is just pedantic/was added by Claude, so removed.

infra/operator/src/resources/pvcs.rs
svix-james: K8s doesn't support decreasing storage size at all. I will change the title.

infra/operator/src/resources/services.rs
svix-james: Yep, sorry, lost track of that.

infra/operator/src/resources/pvcs.rs
svix-james: I'll change it to WARN. The reason it's not currently WARN or ERROR is that they are likely to not exist yet when the chart is first provisioned so it's fairly common for this to be logged.

svix-jplatte (approved):

PR #957, opened 2026-04-14
create a new trait for things that get serialized with postcard

### LINKED ISSUE #926: ban skip_serialize_if / etc in modules
We should make sure that none of the operations structs contain `#[serde(skip_serialize_if)]` and `#[serde(tag)]` because they break postcard. Ideally we can use clippy to just ban these attributes, but I'm not sure if clippy can run on macro attributes or if we'll need to do something else. Frankly, we could probably just add a CI job that runs `grep "serde.*skip_serialize_if"` on the modules. :-)

svix-jbrown:
This turned out to be a lot more work than I thought.

Fixes #926.

## Basic implementation

We create a new trait (`PersistableValue`) representing any primitive that can be safely serialized to the database. We then provide a derive macro for deriving that on a struct which checks that all of the fields are persistable and that the struct doesn't have any unsafe serde attributes (skip_serializing_if, alias, flatten).

We then implement that on every type that (transitively) gets written to fjall, and make the high-level interfaces (TableRow, handle::Request) require it.

## Pending Questions

 - Do we want to ban `serde::with`? There's no real way to figure out the type that it returns from a proc-macro and someone could probably do something untoward in there; imagine something like

```rust
#[derive(Serialize, Deserialize, PersistableValue)]
struct Safe {
  #[serde(with = "make_me_unsafe")]
  innocuous_value: u32
}

struct Unsafe {
  #[serde(skip_serializing_if = "u32::zero")]
  inner
}

mod make_me_unsafe {
  pub(crate) fn deserialize<'de, D>(deserializer: D) -> Result<Safe, D::Error>
  where
    D: Deserializer<'de>,
  {
    let buf = Unsafe::deserialize(deserializer)?;
    Safe { innocuous_value: buf.inner }
  } 

  pub(crate) fn serialize<S>(safe: Safe, serializer: S) -> Result<S::Ok, S::Error>
where
    S: Serializer,
  {
    Unsafe { inner: safe.innocuous_value }.serialize(serializer)
  }
```

 - Is any of this actually worth the complexity? Would it be easier to just switch from serde to something else like rkyv since we need to annotate every struct anyway?

### DIFF
--- a/crates/auth-token/src/operations/create.rs
+++ b/crates/auth-token/src/operations/create.rs
@@ -7,7 +7,7 @@ use crate::{
     entities::TokenHashed,
     operations::{AuthTokenRaftState, AuthTokenRequest, CreateResponse},
 };
-use diom_core::types::Metadata;
+use diom_core::{PersistableValue, types::Metadata};
 use diom_error::Result;
 use diom_id::{AuthTokenId, NamespaceId, UuidV7RandomBytes};
 use diom_operations::OpContext;
@@ -16,7 +16,7 @@ pub struct CreateResponseData {
     pub model: AuthTokenModel,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, PersistableValue)]
 pub struct CreateAuthTokenOperation {
     namespace_id: NamespaceId,
     id_random_bytes: UuidV7RandomBytes,
--- a/crates/diom-core/src/persistable_value.rs
+++ b/crates/diom-core/src/persistable_value.rs
@@ -1,6 +1,6 @@
 use std::{
     collections::{BTreeMap, BTreeSet, HashMap, HashSet},
-    num::NonZeroU64,
+    num::{NonZeroU16, NonZeroU32, NonZeroU64},
 };
 
 /// Marker trait for whether something can be
@@ -28,6 +28,8 @@ impl PersistableValue for i32 {}
 impl PersistableValue for i64 {}
 impl PersistableValue for i128 {}
 // TODO: use std::num::NonZero<T> when stabilized
+impl PersistableValue for NonZeroU16 {}
+impl PersistableValue for NonZeroU32 {}
 impl PersistableValue for NonZeroU64 {}
 impl PersistableValue for String {}
 impl PersistableValue for &str {}
--- a/crates/diom-core/src/persistable_value.rs
+++ b/crates/diom-core/src/persistable_value.rs
@@ -1,0 +1,86 @@
+use std::{
+    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
+    num::NonZeroU64,
+};
+
+/// Marker trait for whether something can be
+/// safely represented in fjall.
+///
+/// # Safety
+///
+/// Implementers must guarantee the following:
+///
+/// 1. The serde implementation for this object doesn't do #[serde(skip_serializing_if)] or #[serde(flatten)]
+///  2. The serialization of this object is stable across version
+///
+/// Generally, you should prefer to use the `PersistableValue` derive to implement this safely.
+pub trait PersistableValue {}
+
+impl PersistableValue for bool {}
+impl PersistableValue for u8 {}
+impl PersistableValue for u16 {}
+impl PersistableValue for u32 {}
+impl PersistableValue for u64 {}
+impl PersistableValue for u128 {}
+impl PersistableValue for i8 {}
+impl PersistableValue for i16 {}
+impl PersistableValue for i32 {}
+impl PersistableValue for i64 {}
+impl PersistableValue for i128 {}
+// TODO: use std::num::NonZero<T> when stabilized
+impl PersistableValue for NonZeroU64 {}
+impl PersistableValue for String {}
+impl PersistableValue for &str {}
+impl PersistableValue for &[u8] {}
+impl PersistableValue for uuid::Uuid {}
+// TODO: do we want to serialize jiff timestamps raw?
+impl PersistableValue for jiff::Timestamp {}
+impl PersistableValue for crate::types::ByteString {}
+impl PersistableValue for crate::types::DurationMs {}
+impl PersistableValue for crate::types::UnixTimestampMs {}
+impl PersistableValue for crate::types::Metadata {}
+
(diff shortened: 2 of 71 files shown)


crates/diom-core/src/persistable_value.rs
```diff
@@ -0,0 +1,88 @@
...
+unsafe impl PersistableValue for i64 {}
+unsafe impl PersistableValue for i128 {}
+// TODO: use std::num::NonZero<T> when stabilized
+unsafe impl PersistableValue for NonZeroU64 {}
+unsafe impl PersistableValue for String {}
+unsafe impl PersistableValue for &str {}
+unsafe impl PersistableValue for &[u8] {}
+unsafe impl PersistableValue for uuid::Uuid {}
+// TODO: do we want to serialize jiff timestamps raw?
+unsafe impl PersistableValue for jiff::Timestamp {}
```
svix-jplatte: … or switch all serialized timestamps to `UnixTimestampMs`?

crates/diom-derive/src/persistable_value.rs
```diff
@@ -0,0 +1,119 @@
+use quote::quote;
+use syn::{Attribute, DataEnum, DataStruct, DeriveInput, Field, LitStr, Token, parse_macro_input};
+
+fn check_serde_container_attrs(attrs: &[Attribute]) -> Result<(), syn::Error> {
+    for attr in attrs {
+        if !attr.path().is_ident("serde") {
+            continue;
+        }
+        attr.parse_nested_meta(|meta| {
+            if meta.path.is_ident("flatten") {
```
svix-jplatte: `flatten` is a field attribute, not a container attribute. Though we should have this function, and check for `tag`, `untagged` and maybe `default` (didn't know this one existed until I found it in docs).

crates/diom-derive/src/persistable_value.rs
```diff
@@ -0,0 +1,119 @@
...
+        let ty = &field.ty;
+        inner.push(quote! { #ty });
+    }
+
+    let ident = &input.ident;
+
+    Ok(quote! {
+        #[allow(unsafe_code)]
+        unsafe impl #impl_generics diom_core::persistable_value::PersistableStruct for #ident #ty_generics #where_clause {
+            type INNER = (#(#inner),* );
```
svix-jplatte:
```suggestion
            type INNER = ( #(#inner,)* );
```

crates/diom-derive/src/persistable_value.rs
```diff
@@ -0,0 +1,119 @@
...
+            inner.push(quote! { #ty });
+        }
+    }
+
+    let ident = &input.ident;
+
+    Ok(quote! {
+        #[allow(unsafe_code)]
+        unsafe impl #impl_generics diom_core::persistable_value::PersistableStruct for #ident #ty_generics #where_clause {
+            type INNER = (#(#inner),* );
```
svix-jplatte:
```suggestion
            type INNER = ( #(#inner,)* );
```

crates/diom-core/src/persistable_value.rs
```diff
@@ -0,0 +1,90 @@
...
+impl_persistable_value_for_tuple! { 0 1 2 }
+impl_persistable_value_for_tuple! { 0 1 2 3 }
+impl_persistable_value_for_tuple! { 0 1 2 3 4 }
+impl_persistable_value_for_tuple! { 0 1 2 3 4 5 }
+impl_persistable_value_for_tuple! { 0 1 2 3 4 5 6 }
+impl_persistable_value_for_tuple! { 0 1 2 3 4 5 6 7}
+impl_persistable_value_for_tuple! { 0 1 2 3 4 5 6 7 8}
+impl_persistable_value_for_tuple! { 0 1 2 3 4 5 6 7 8 9}
+impl_persistable_value_for_tuple! { 0 1 2 3 4 5 6 7 8 9 10}
+impl_persistable_value_for_tuple! { 0 1 2 3 4 5 6 7 8 9 10 11}
```
svix-jplatte:
```suggestion
impl_persistable_value_for_tuple! { 0 1 2 3 4 5 6 7 }
impl_persistable_value_for_tuple! { 0 1 2 3 4 5 6 7 8 }
impl_persistable_value_for_tuple! { 0 1 2 3 4 5 6 7 8 9 }
impl_persistable_value_for_tuple! { 0 1 2 3 4 5 6 7 8 9 10 }
impl_persistable_value_for_tuple! { 0 1 2 3 4 5 6 7 8 9 10 11 }
```

crates/diom-derive/src/persistable_value.rs
svix-jplatte: I guess with how you've set up the trait, this doesn't need to be a tuple for the single-inner-type case. But maybe more consistent?

crates/diom-id/src/public.rs
```diff
@@ -48,6 +48,9 @@ impl<'de, M: PublicIdMarker> Deserialize<'de> for Public<Id<M>> {
     }
 }
 
+#[allow(unsafe_code)] // we promise to keep the serialization of Id stable
```
svix-jplatte: Maybe we should remove `deny(unsafe_code)` and enable that clippy lint that requires safety comments.

svix-jplatte (commented):
Seems like a good solution, at least for the short term.

crates/auth-token/src/operations/create.rs
```diff
@@ -7,7 +7,7 @@ use crate::{
     entities::TokenHashed,
```
tasn, line 7: Just a knee-jerk reaction: let's avoid `unsafe` if we can.

crates/diom-core/src/persistable_value.rs
svix-jbrown: yeah, I think we should do that one if only because it's more compact. I'll do it in a follow-on PR.

svix-jbrown:
I removed the `unsafe` and made the rest of the changes, as well as rebasing on top of main

src/core/cluster/logs.rs
```diff
@@ -129,6 +130,9 @@ enum RowType {
 #[serde(transparent)]
 struct Log(LogEntry);
 
+// this one right here officer, this is the bad one
```
svix-jplatte, line 133: what 😂

svix-jplatte (approved):

PR #958, opened 2026-04-14
Support roles and polices in the bootstrap process

### LINKED ISSUE #818: ACL: add support for policy and role in the bootstrap process
At the moment we add them manually, though tbh we can probably automatically generate this support or at least make it much easier to add.
The bootstrap script is intended to look like the CLI and we can just use the internal_client to make the calls. So it should be very little boilerplate to add items there.

Anyhow, we need both policies and role there to make them usable.

svix-pluiz:
closes https://github.com/svix/diom-private/issues/818

# Changes

1. Rename `raw-admin` to `admin` in the cli
2. Add `AuthPolicy` and `AuthRole` to the BootstrapCommand enum
3. Update BootstrapCommand `from_str` pattern match to include auth-policy and auth-role commands


z-clients/cli/src/main.rs
```diff
@@ -78,10 +78,10 @@ enum RootCommands {
     Msgs(MsgsArgs),
     RateLimit(RateLimitArgs),
     Health(HealthArgs),
-    /// Send raw administrative commands.
+    /// Send administrative commands.
     ///
     /// Prefer the cluster-admin family of CLI commands.
```
svix-jplatte, line 83: The "prefer the cluster-admin family of CLI commands" bit no longer makes sense now with roles and policies being part of the admin API. Not really sure what to do about that though.

tasn: It looks like `raw-admin` is some stuff that JB added for himself. We should probably move all the cluster stuff from here to `cluster-admin raw ...` and keep `admin` as just for the auth/policy/role etc.

svix-pluiz: Are we going to do this in this PR?

svix-jplatte: I think it makes more sense to have separate PRs, but then this one probably shouldn't touch the CLI at all.

src/bootstrap.rs
```diff
@@ -23,7 +27,9 @@ enum BootstrapCommand {
     Idempotency(IdempotencyCreateNamespaceIn),
     RateLimit(RateLimitCreateNamespaceIn),
     Msgs(MsgNamespaceCreateIn),
-    AuthToken(AuthTokenCreateNamespaceIn),
+    AdminAuthToken(AuthTokenCreateNamespaceIn),
+    AdminAuthPolicy(AdminAccessPolicyUpsertIn),
+    AuthRole(AdminRoleUpsertIn),
```
svix-jplatte: This is not what I meant, this is:


```suggestion
    AuthToken(AuthTokenCreateNamespaceIn),
    AdminAuthPolicy(AdminAccessPolicyUpsertIn),
    AdminAuthRole(AdminRoleUpsertIn),
```

Auth tokens exist both as a regular resource (where namespaces can be managed) and an admin resource.

src/bootstrap.rs
```diff
@@ -90,50 +121,54 @@ impl FromStr for BootstrapCommand {
...
                     format!("invalid JSON for rate-limit namespace: {json_str:?}")
                 })?,
             )),
-            "msgs" => Ok(BootstrapCommand::Msgs(
+            ("msgs", "namespace", "create") => Ok(BootstrapCommand::Msgs(
                 serde_json::from_str(json_str)
                     .with_context(|| format!("invalid JSON for msgs namespace: {json_str:?}"))?,
             )),
-            _ => bail!("unknown module {module:?}"),
+            ("raw-admin", "auth-policy", "upsert") => Ok(BootstrapCommand::AdminAuthPolicy(
```
svix-jplatte: I think you can drop the `raw-` here in this PR already.

svix-pluiz: okok, reverted it

svix-jplatte (approved):

PR #970, opened 2026-04-14
cli: Remove admin cluster commands

svix-jplatte:
We have cluster-admin as a higher-level version of the same thing.

### DIFF
--- a/codegen/templates/cli/api_resource.rs.jinja
+++ b/codegen/templates/cli/api_resource.rs.jinja
@@ -12,10 +12,13 @@ use clap::{Args, Subcommand};
 use crate::prelude::*;
 
 {% for _, sub in resource.subresources | items -%}
-use super::{{ sub.name | to_upper_camel_case }}Args;
+    {# exclude internal-ish API, exposed in a better manner under cluster-admin -#}
+    {% if sub.name != "admin.cluster" -%}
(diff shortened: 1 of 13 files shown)


svix-lucho (approved):

codegen/templates/cli/api_resource.rs.jinja
```diff
@@ -12,10 +12,13 @@ use clap::{Args, Subcommand};
 use crate::prelude::*;
 
 {% for _, sub in resource.subresources | items -%}
-use super::{{ sub.name | to_upper_camel_case }}Args;
+    {# exclude internal-ish API, exposed in a better manner under cluster-admin -#}
+    {% if sub.name != "admin.cluster" -%}
```
tasn, line 16: Why do this and not just rename the op of the API endpoints?

svix-jplatte: I was not under the impression that we wanted to change anything for the SDKs. What would you propose renaming it to?

tasn: Why not change it under the SDKs? Can be `ClusterAdmin` in the SDKs if we feel like it's a separate category (which it is). We don't want people auto-completing in the SDKs to face confusion any more than we want the CLI users to face confusion.

svix-jplatte:
Closing in favor of #1005.

PR #972, opened 2026-04-14
js: Add /internal entrypoint and export serializers from it

svix-lucho:
We need the model Serializers to use them in the JS CLI in diom-www. But I don't want to export them from `@diomhq/diom` because they are not necessary for normal SDK users. 

This adds a separate entrypoint (`/internal`) to the JS SDK that exports these Serializers (and can be used for other things if needed). 

With this, I can then import the Serializers like `import ModelInSerializer from "@diomhq/diom/internal"`, while regular SDK users won't be affected at all by this. 

The alternative is to publish a completely separate npm package with the serializers, but that seemed like a lot of work, and another thing we have to maintain. This looks simple enough.


codegen/templates/javascript/summary.ts.jinja
svix-lucho, line 1: This works but it's probably wrong - I don't understand what summary/api_summary mean

svix-jplatte: All of `summary`, `api_summary`, `component_type_summary` just mean openapi-codegen calls the template once, with the full `api`  object as template context. So this is fine, although not a usecase I originally thought of :)

z-clients/javascript/src/internal.ts
```diff
@@ -0,0 +1,122 @@
+// this file is @generated
```
svix-jplatte, line 1: Can you add this file to `.gitattributes`?

svix-lucho: Done

svix-jplatte (dismissed):
Nice 👍 
Fix for failing CI is in https://github.com/svix/diom-private/pull/986.

svix-jplatte:
Needs re-codegen

svix-jplatte (approved):

PR #983, opened 2026-04-15
CI: add a server security check CI job.

tasn:


tasn:
NVM, it's already there in the mega one.

PR #988, opened 2026-04-15
Support multi-line JSON objects in bootstrap

svix-pluiz:
From https://svixhq.slack.com/archives/C0A72988962/p1776215766376059

This PR makes parse_bootstrap support multi-line JSON configs*
Also add new tests mixing one-line and multi-line to check if the updated parser is working as expected.

*Known caveat: multiline string values cannot be parsed

### DIFF
--- a/src/bootstrap.rs
+++ b/src/bootstrap.rs
@@ -33,6 +33,63 @@ enum BootstrapCommand {
 }
 
 impl BootstrapCommand {
+    /// Shared logic for matching module/resource/action and deserializing JSON.
+    fn from_parts(
+        module: &str,
+        resource: &str,
+        action: &str,
+        json_str: &str,
+    ) -> anyhow::Result<Self> {
+        match (module, resource, action) {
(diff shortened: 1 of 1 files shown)


src/bootstrap.rs
```diff
@@ -542,4 +592,90 @@ mod tests {
...
+            \"id\": \"pol\",\n\
+            \"description\": \"full access\",\n\
+            \"rules\": [\n\
+                {\n\
+                    \"effect\": \"allow\",\n\
+                    \"resource\": \"*:*:*\",\n\
+                    \"actions\": [\"*\"]\n\
+                }\n\
+            ]\n\
+        }";
```
svix-jplatte: You can use a raw string literal to make this more readable:


```rust
        let input = r#"admin auth-policy configure {
            "id": "pol",
            ...
        }"#;
```

svix-jplatte (dismissed):

svix-jplatte (approved):

PR #989, opened 2026-04-15
Allow dumping and env-overriding JWT config

svix-jplatte:
Ref. https://svixhq.slack.com/archives/C0A72988962/p1776215510425579

### DIFF
--- a/src/cfg/mod.rs
+++ b/src/cfg/mod.rs
@@ -585,7 +585,7 @@ pub struct ConfigurationInner {
     #[env_overridable(skip)]
     #[dumpable_config(skip)] // no straightforward way to serialize a #[serde(flatten)]
     #[validate(nested)]
-    pub jwt: Option<JwtConfig>,
+    pub jwt: JwtConfig,
 }
 
 impl ConfigurationInner {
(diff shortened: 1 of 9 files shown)


src/cfg/mod.rs
```diff
@@ -649,12 +653,286 @@ pub enum JwtKey {
     Ps512 { public_key_pem: String },
 }
 
+impl JwtKey {
```
tasn, line 656: I could swear we have all of this code somewhere, don't we?

tasn (approved):

src/cfg/mod.rs
```diff
@@ -649,12 +653,286 @@ pub enum JwtKey {
...
+                    format!("{prefix}_PUBLIC_KEY_PEM")
+                },
+                docstring: Some(
+                    "Public key PEM for JWT algorithms RS256 / RS384 / RS512 / ES256 / ES384 / PS256 / PS384 / PS512",
+                ),
+            },
+        ]
+    }
+}
+
```
svix-jbrown, line 935: spending ¼ of the config file on fragile hand-parsing just this one option seems pretty gross and suggests to me that the way we're representing our config is wrong. can we try to instead brainstorm a different way to represent this particular configuration so that we don't need to hardcode it?

maybe it should just be two fields (`algorithm` and `parameter`) where we treat the parameter as either a public key PEM or a shared-secret depending on the algorithm?

We can still have `ConfigurationInner` have a method that returns a sum-typed `JwtKey` of course

svix-jplatte: Yeah I'm not super happy with this, but I expect any alternative along the lines you mentioned will be similarly annoying. I'd rather try to cut down on boilerplate without fundamentally changing the approach. #1024 does a small piece of that.

PR #998, opened 2026-04-15
Add background task for ACTUALLY deleting expired messages

svix-gabriel:
While we filter out expired messages immedietly in user-facing queries, we still want to clean up IRL disk space using a background task.

I _think_ I followed the right patterns for background tasks. But if not lmk.


svix-jbrown (commented):
I think this is good; we should probably add some metrics for these background jobs at some point.

tests would be nice too. :-)

svix-jbrown (approved):

PR #1016, opened 2026-04-15
Show >real< struct examples on --help CLI commands

svix-pluiz:


codegen/templates/cli/api_resource.rs.jinja
```diff
@@ -15,11 +15,69 @@ use crate::prelude::*;
...
+    {%- elif type.is_bool() -%}
+        true
+    {%- elif type.is_int_or_uint() -%}
+        123
+    {%- elif type.is_list() or type.is_set() -%}
+        [\"...\"]
+    {%- elif type.is_map() or type.is_json_object() -%}
+        {\"key\": \"...\"}
+    {%- elif type.is_schema_ref() -%}
+        {\"...\": \"...\"}
```
svix-jbrown: can we just recurse for these?

codegen/templates/cli/api_resource.rs.jinja
```diff
@@ -15,11 +15,69 @@ use crate::prelude::*;
...
+    {%- elif type.is_unix_timestamp_ms() -%}
+        1234567890123
+    {%- elif type.is_duration_ms() -%}
+        60000
+    {%- elif type.is_datetime() -%}
+        \"2030-01-01T00:00:00Z\"
+    {%- elif type.is_u64() -%}
+        123
+    {%- elif type.is_bytes() -%}
+        \"dGVzdA==\"
```
svix-jplatte: Note that we do not base64-encode bytes. Instead, we escape any bytes that are not (displayable) UTF-8. So the string would for example contain `\u0000` for a NUL byte.

svix-pluiz: > Note that we do not base64-encode bytes. Instead, we escape any bytes that are not (displayable) UTF-8. So the string would for example contain \u0000 for a NUL byte.

Hm... Do you have any other suggestion for the `is_bytes()` example?

codegen/templates/cli/api_resource.rs.jinja
svix-pluiz: To be honest I didn't check if recursion would work with Jinja, It ended up working (!) and I also updated the `is_schema_ref` branch here https://github.com/svix/diom/pull/1016/commits/61ff99e18f737400ec6553183c97d6ae64d79fe9

codegen/templates/cli/api_resource.rs.jinja
svix-jplatte: Honestly I think even `"..."` would be better as the base64 is actively misleading IMHO.

svix-jplatte (approved):

PR #1021, opened 2026-04-16
add a header with the log_id of the last change committed in a request

svix-jbrown:
this is an extension of the OCC feature for kv which adds a header recording the last log_id of a committed transaction in a request. this can be used to implement linearization for external clients if we're clever about it.

### DIFF
--- a/src/core/cluster/handle.rs
+++ b/src/core/cluster/handle.rs
@@ -21,9 +21,14 @@ use itertools::Itertools;
 use jiff::Timestamp;
 use maplit::btreeset;
 use openraft::{RaftNetworkFactory, ServerState};
+use parking_lot::Mutex;
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 use tokio::sync::mpsc::Sender;
+
+tokio::task_local! {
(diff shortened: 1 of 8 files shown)


src/core/cluster/handle.rs
```diff
@@ -21,10 +21,15 @@ use itertools::Itertools;
 use jiff::Timestamp;
 use maplit::btreeset;
 use openraft::{RaftNetworkFactory, ServerState};
+use parking_lot::Mutex;
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 use tokio::sync::mpsc::Sender;
 
+tokio::task_local! {
+    pub(super) static APPLIED_LOG_ID: Arc<Mutex<Option<LogId>>>;
```
svix-gabriel: I'm not super familiar with task_local!, but based on my reading of the docs, why do we need the Arc?

svix-jbrown: I can get rid of the Arc; it just means switching the `.get()` for a `.with()`.

svix-gabriel (approved):

PR #1024, opened 2026-04-16
Simplify DumpableConfig implementations

svix-jplatte:
Paying down some tech debt introduced in #989.

### DIFF
--- a/src/cfg/dumpable_config.rs
+++ b/src/cfg/dumpable_config.rs
@@ -11,21 +11,21 @@ pub(crate) trait DumpableConfig {
     }
 }
 
-pub(crate) fn dump_field<T: Serialize + ?Sized>(
+pub(crate) fn dump_field<T: Serialize + ?Sized, W: Write>(
     name: &str,
     value: &T,
-    writer: &mut impl Write,
+    writer: &mut W,
 ) -> Result<(), anyhow::Error> {
     let mut buffer = String::new();
-    let serialized = value.serialize(toml::ser::ValueSerializer::new(&mut buffer))?;
-    writeln!(writer, "{name} = {serialized}")?;
+    value.serialize(toml::ser::ValueSerializer::new(&mut buffer))?;
+    writeln!(writer, "{name} = {buffer}")?;
     Ok(())
 }
 
-pub(crate) fn dump_optional_field<T: Serialize + ?Sized>(
+pub(crate) fn dump_optional_field<T: Serialize + ?Sized, W: Write>(
     name: &str,
     value: Option<&T>,
-    writer: &mut impl Write,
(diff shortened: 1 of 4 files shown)


src/cfg/dumpable_config.rs
```diff
@@ -1,8 +1,37 @@
...
+    fn dump_map<W: Write>(&self, writer: &mut W, prefix: String) -> anyhow::Result<()> {
         writeln!(writer, "[{prefix}]")?;
         self.dump_fields(writer, prefix)
     }
 }
+
+pub(crate) fn dump_field<T: Serialize + ?Sized>(
+    name: &str,
+    value: &T,
+    writer: &mut impl Write,
```
svix-jbrown: nit, I still don't like named type variables and `impl Trait` in the same function call
```suggestion
pub(crate) fn dump_field<T: Serialize + ?Sized, W: Write>(
    name: &str,
    value: &T,
    writer: &mut W,
```

src/cfg/dumpable_config.rs
```diff
@@ -1,8 +1,37 @@
...
+    let mut buffer = String::new();
+    let serialized = value.serialize(toml::ser::ValueSerializer::new(&mut buffer))?;
+    writeln!(writer, "{name} = {serialized}")?;
+    Ok(())
+}
+
+pub(crate) fn dump_optional_field<T: Serialize + ?Sized>(
+    name: &str,
+    value: Option<&T>,
+    writer: &mut impl Write,
```
svix-jbrown: same here
```suggestion
pub(crate) fn dump_optional_field<T: Serialize + ?Sized, W: Write>(
    name: &str,
    value: Option<&T>,
    writer: &mut W,
```

src/cfg/dumpable_config.rs
```diff
@@ -1,8 +1,37 @@
...
 }
+
+pub(crate) fn dump_field<T: Serialize + ?Sized>(
+    name: &str,
+    value: &T,
+    writer: &mut impl Write,
+) -> Result<(), anyhow::Error> {
+    let mut buffer = String::new();
+    let serialized = value.serialize(toml::ser::ValueSerializer::new(&mut buffer))?;
+    writeln!(writer, "{name} = {serialized}")?;
```
svix-jbrown: um, shouldn't you be writing the `buffer`, not the return value?

svix-jplatte: Well, it clearly works :P

Appears that it's the same value due to <https://docs.rs/toml/latest/toml/ser/struct.ValueSerializer.html#associatedtype.Ok>. But I'll change it, simpler that way.

svix-jbrown (approved):

PR #1030, opened 2026-04-16
Remove unused ErrorType::Validation

svix-jplatte:
We only return 422's via `MsgPackOrJsonRejection`.


svix-jplatte:
Closing in favor of #1114.

PR #1054, opened 2026-04-17
Unquote `PRERELEASE_FLAG` from release workflow host step

svix-pluiz:
One of the shellcheck warnings that I fixed here https://github.com/svix/diom/pull/1012/commits/432514fe74be73f98b5b7b50251e743a730e9013 ended up breaking the command to create the release

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -282,7 +282,8 @@ jobs:
           # If we're editing a release in place, we need to upload things ahead of time
           gh release upload "${{ needs.plan.outputs.tag }}" artifacts/*
 
-          gh release edit "${{ needs.plan.outputs.tag }}" --target "$RELEASE_COMMIT" "$PRERELEASE_FLAG" --draft=false
+          # shellcheck disable=SC2086
+          gh release edit "${{ needs.plan.outputs.tag }}" --target "$RELEASE_COMMIT" $PRERELEASE_FLAG --draft=false
 
   publish-homebrew-formula:
     # figure out if we can set this w/o `allow-dirty = ["ci"]` in dist-workspace.toml


tasn (approved):

.github/workflows/release.yml
```diff
@@ -282,7 +282,8 @@ jobs:
           # If we're editing a release in place, we need to upload things ahead of time
           gh release upload "${{ needs.plan.outputs.tag }}" artifacts/*
 
-          gh release edit "${{ needs.plan.outputs.tag }}" --target "$RELEASE_COMMIT" "$PRERELEASE_FLAG" --draft=false
+          # shellcheck disable=SC2086
+          gh release edit "${{ needs.plan.outputs.tag }}" --target "$RELEASE_COMMIT" $PRERELEASE_FLAG --draft=false
```
svix-jbrown, line 286: ugh flags in env vars is a terrible pattern; another wray to do it that is a little less janky would be the following:

```suggestion
PRERELEASE_FLAG: "${{ fromJson(steps.host.outputs.manifest).announcement_is_prerelease && '1' || '0' }}"
RELEASE_COMMIT: "${{ github.sha }}"
run: |
         # If we're editing a release in place, we need to upload things ahead of time
         gh release upload "${{ needs.plan.outputs.tag }}" artifacts/*

         ARGS=("--target" "$RELEASE_COMMIT" "--draft=false")
         if [[ "$PRERELEASE_FLAG" = "1" ]]; then
            ARGS+=("--prerelease")
         fi
         gh release edit "${{ needs.plan.outputs.tag }}" "${ARGS[@]}"
```

svix-jplatte: I think we really should not be editing this file at all, instead letting upstream know when things are jenky. It is auto-generated by `dist`, I'm somewhat surprised that it doesn't complain when it's hand-edited.

PR #1072, opened 2026-04-17
Prepare version 0.2.1.

tasn:


svix-lucho (dismissed):

svix-lucho (approved):

z-clients/javascript/package.json
svix-pluiz, line 1: @tasn we need to run the command again to update package-lock.json

PR #1082, opened 2026-04-17
Fix package-lock.json version // Adjust Java version RegExp

svix-pluiz:


tools/bump_version.py
```diff
@@ -136,14 +136,14 @@ def cmd_bump(new_version: str) -> int:
...
 
-    if semver.Version.is_valid(old_version) and semver.Version.parse(
-        new_version
-    ) <= semver.Version.parse(old_version):
-        print(
-            f"ERROR: new version '{new_version}' is not greater than current '{old_version}'",
-            file=sys.stderr,
-        )
-        return 1
+    # if semver.Version.is_valid(old_version) and semver.Version.parse(
```
tasn, line 139: ?

svix-pluiz: I force-pushed to fix this here https://github.com/svix/diom/commit/def28cf8d8d3d7e72f1004d405c23f2c851797cb

I forgot to uncomment this when I was running the command passing the current version to check what would change

tasn (approved):

PR #1089, opened 2026-04-18
rust: expose .kind() on errors

svix-jbrown:
doesn't seem that helpful to have a kind if you can't get it out again, right?

should we make `ErrorKind` non-exhaustive?


svix-jplatte (dismissed):
Yeah, I forgot this, thanks. Should be non-exhaustive imo, and we might want to use `doc(hidden)` on the Other variant too.

svix-jplatte (approved):

PR #1095, opened 2026-04-20
Enable str_to_string clippy lint in lower crates

svix-jplatte:
[Why restrict this?](https://rust-lang.github.io/rust-clippy/master/?search=str_to_s)

> The `to_string` method is also used on other types to convert them to a string. When called on a `&str` it turns the `&str` into the owned variant `String`, which can be more specifically expressed with `.to_owned()`.

I'll do the higher-up crates of our dependency graph in more PRs and ultimately turn this into a workspace-wide lint.

### DIFF
--- a/crates/diom-core/src/fifo_cache.rs
+++ b/crates/diom-core/src/fifo_cache.rs
@@ -59,7 +59,7 @@ mod tests {
     #[test]
     fn get_returns_inserted_value() {
         let mut cache = FifoCache::new(10);
-        cache.put("a".to_string(), 1);
+        cache.put("a".to_owned(), 1);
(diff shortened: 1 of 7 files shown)


svix-gabriel (approved):
Not that I have a strong preference either way, but why is `to_owned` preferred over `to_string` in this context?

svix-jplatte:
Did you read the PR description?

If you did, another way of looking at it: `to_owned` better expresses _why_ you the method call is needed at all. The input to it already is a "string" in the common meaning of the word. The reason for the method call is to get an owned `String`, not to "stringify" what's already a string.

PR #1097, opened 2026-04-20
tools: Don't update Rust dependencies to fix up Cargo.lock

svix-jplatte:
Same idea as #1078, but using a command with minimal side effects (ideally none, but that seems to not be true under certain circumstances).

I tested it and it doesn't update dependencies unnecessarily here as it stands right now.


svix-pluiz (approved):
We need to update svix-webhooks as well

PR #1099, opened 2026-04-20
ci: Avoid recompilation of crates/* in check-codegen

svix-jplatte:


svix-jplatte:
> thread 'main' (4370) panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tracing-subscriber-0.3.23/src/util.rs:94:14:
> failed to set global default subscriber: SetLoggerError(())

wat

edit: I suspect life-before-main shenanigans from usage of `ctor`. We should stop using it.

PR #1100, opened 2026-04-20
Bootstrap: support setting mulitple config scripts at once.

tasn:
We previously only supported on script or inline, now we support both inline and multiple scripts at once. This is intended to make it easier to bootstrap multiple different unrelated components at once, as well as bootstrapping potentially sensitive information.

### DIFF
--- a/src/cfg/mod.rs
+++ b/src/cfg/mod.rs
@@ -542,7 +542,7 @@ pub struct ConfigurationInner {
     #[dumpable_config(nest)]
     pub cluster: ClusterConfiguration,
 
-    /// Paths to YAML bootstrap files, applied in order.
+    /// Paths to bootstrap scripts, applied in order.
     #[serde(default)]
     pub bootstrap_cfg_paths: Vec<String>,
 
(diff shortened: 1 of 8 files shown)


src/cfg/mod.rs
```diff
@@ -542,11 +542,17 @@ pub struct ConfigurationInner {
...
+    pub bootstrap_cfg_paths: Vec<String>,
+
+    /// Deprecated: use `bootstrap_cfg_paths` instead.
+    #[serde(default)]
+    #[env_overridable(skip)]
+    #[dumpable_config(skip)]
     pub bootstrap_cfg_path: Option<String>,
 
-    /// YAML bootstrap data. Takes precedence over `bootstrap_cfg_path`
+    /// Inline YAML bootstrap data, applied before any `bootstrap_cfg_paths`.
```
svix-james: this is not YAML anymore, right?

tasn: Blah, I'll fix it. Didn't even notice.

tests/it/bootstrap.rs
```diff
@@ -83,3 +83,45 @@ async fn test_bootstrap_env_var_based() -> TestResult {
...
+    assert_bootstrap_namespaces(&test_server.client).await
+}
+
+#[tokio::test]
+async fn test_bootstrap_multiple_sources() -> TestResult {
+    let inline = include_str!("static/bootstrap.test").to_string();
+    let test_server = TestServerBuilder::with_default_config()
+        .tap_cfg(|cfg| {
+            cfg.bootstrap_cfg = Some(inline);
+            cfg.bootstrap_cfg_paths = vec!["tests/it/static/bootstrap2.test".to_string()];
```
svix-james: Can we instead test here that `paths` supports multiple files?

tasn: Done.

svix-james (approved):

PR #1102, opened 2026-04-20
Operator: Update otel settings

svix-james:
This changed last week in the main application config. Update the CRD
to have similar structure. Also, fix a bug in the protocol parsing
code.

Note that I've snuck in a version for the operator, which will be needed
prior to releasing.

### DIFF
--- a/infra/operator/Cargo.toml
+++ b/infra/operator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "diom-operator"
-version.workspace = true
+version = "0.1.0"
 license.workspace = true
 rust-version.workspace = true
 repository.workspace = true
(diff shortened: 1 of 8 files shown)


infra/operator/tests/it/e2e.rs
```diff
@@ -121,13 +121,6 @@ async fn test_storage_resize() -> TestResult {
         .replace(env.name(), &PostParams::default(), &cluster)
         .await?;
 
-    run_with_retries(async || {
```
svix-james, line 124: Unneeded assertion that's causing CI flakiness.

svix-jbrown (approved):

infra/operator/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "diom-operator"
-version.workspace = true
+version = "0.1.0"
```
svix-jbrown, line 3: not sure why this one is in here?

svix-james: b/c the operator will have its own version independent of Diom. Does that not make sense?

PR #1103, opened 2026-04-20
gha: try pulling ghcr->dockerhub by digest instead of tag

svix-jbrown:
I don't actually know what the `digest` output contains from a multiarch build, but hopefully it's something immutable we can pull by. Let's experiment!

### DIFF
--- a/.github/workflows/docker-build-release.yml
+++ b/.github/workflows/docker-build-release.yml
@@ -104,7 +104,7 @@ jobs:
       - name: "Copy images"
         run: |
           for dest in ${{ steps.meta.outputs.tags }}; do
-            source="${{ inputs.staging_image_repo }}:${{ needs.build.outputs.digest }}"
+            source="${{ inputs.staging_image_repo }}@${{ needs.build.outputs.digest }}"
             echo "Copying $source to $dest"
             regctl image copy "$source" "$dest"
           done
--- a/.github/workflows/docker-build-release.yml
+++ b/.github/workflows/docker-build-release.yml
@@ -57,11 +57,12 @@ jobs:
       cache-scope: ${{ inputs.cache-scope }}
       cache-mode: max
       meta-flavor: |
-        latest=true
+        latest=false
+      # write this to GHCR with the sha-012345 and version docker tags;
+      # these are unused because we pull by output-digest and are just for debugging
       meta-tags: |
-        type=semver,pattern={{version}},value=${{inputs.version}}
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-build-release.yml
```diff
@@ -102,9 +103,8 @@ jobs:
             type=semver,pattern={{major}},value=${{inputs.version}}
       - name: "Copy images"
         run: |
-          for source in ${{ steps.meta.outputs.tags }}; do
-            # shellcheck disable=SC2001
-            dest="$(echo "$source" | sed -e 's,${{ inputs.staging_image_repo }},${{ inputs.image_repo }},')"
+          for dest in ${{ steps.meta.outputs.tags }}; do
```
tasn, line 106: We need to tag it in dockerhub though, no?

svix-jbrown: so, `steps.meta.outputs.tags` contains the fully-qualified dockerhub tags (it's a bunch of strings like `docker.io/svix/diom-server:1.2.3` and `docker.io/svix/diom-server:latest`

then `regctl` is copying from ghcr (by digest) and pushing to those tags

.github/workflows/docker-build-release.yml
```diff
@@ -102,9 +103,8 @@ jobs:
...
+            source="${{ inputs.staging_image_repo }}:${{ needs.build.outputs.digest }}"
```
tasn:
```suggestion
            source="${{ inputs.staging_image_repo }}@sha256:${{ needs.build.outputs.digest }}"
```
No?

.github/workflows/docker-build-release.yml
tasn: OK, great!

.github/workflows/docker-build-release.yml
svix-jbrown: the regular build-push-action's digest output already has the `sha256:` on the front of it, so I'm **hoping** this one does too

tasn: But isn't the syntax different for sha256 pinning?
It's `image@sha256:123` so because you put the `:` already, at best it will be `image:sha256:123`.

So at the very least it should be:
```suggestion
            source="${{ inputs.staging_image_repo }}@${{ needs.build.outputs.digest }}"
```
No?

tasn (approved):

Issue #682, opened 2026-03-27
Store ACLs (resolved by role) more optimally for permission checking


svix-jplatte:
Right now (as of https://github.com/svix/coyote-private/pull/668) `Permissions` contains a `Arc<[AccessRule]>`. This is not the most optimal format for permissions checking. We should

- Have deny and allow rules separate, and check deny first (then once those have been checked, we can short-circuit on the first matching allow rule)
- Instead of a flat list (or two), we should probably split by module

PR #1108, opened 2026-04-21
Use a more optimal in-memory representation for access rule lists

svix-jplatte:
Part of #682 (doesn't split by module yet, so doesn't close that issue).


crates/diom-authorization/tests/it/verification.rs
```diff
@@ -1,9 +1,11 @@
-use diom_authorization::{AccessRule, Context, RequestedOperation, verify_operation};
+use diom_authorization::{
+    AccessRuleList, Context, RequestedOperation, api::AccessRule, verify_operation,
+};
 use diom_id::Module;
 use serde_json::json;
 
-fn example_rules() -> Vec<AccessRule> {
-    serde_json::from_value(json!([
+fn example_rules() -> AccessRuleList {
```
svix-james, line 7: I'll leave it up to you whether you think it's worthwhile, but maybe we should add a simple test that asserts that the constructed AccessRuleList has the expected items in its `Vec`s?

svix-james (approved):

crates/diom-authorization/tests/it/verification.rs
svix-jplatte: I'll make a follow-up PR for it.

PR #1113, opened 2026-04-21
cfg: move almost everything to NonZeroDurationMs

svix-jbrown:
I saw #1107 and got excited about type-level validation.


src/core/cluster/discovery.rs
```diff
@@ -173,7 +173,7 @@ impl Discovery {
 
         tracing::debug!("starting new cluster discovery");
 
-        let deadline = Instant::now() + self.cfg.cluster.discovery_timeout;
+        let deadline = Instant::now() + self.cfg.cluster.discovery_timeout.as_duration();
```
svix-jplatte, line 176: Don't want to write an `Add` implementation instead? x)

svix-jplatte (approved):

PR #1116, opened 2026-04-21
Operator: Better configuration test

svix-james:
Test directly that we can load the Diom configuration from Operator-supplied env vars. This will potentially capture a lot of the failures that would cause `DiomCluster`s to fail to deploy without needing to run Operator e2e tests on all PRs.


svix-james:
## OSS Review

Adds `temp-env` crate as a dev dependency. Small crate with minimal dependencies, only used for testing.

svix-jplatte (approved):

PR #1117, opened 2026-04-21
CI: Only run operator e2e when operator changes

svix-james:
And on pushes to main.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -187,7 +187,7 @@ jobs:
           fi
   operator-e2e:
     name: Operator e2e tests
-    if: inputs.run_operator_e2e
+    if: ${{ inputs.run_operator_e2e }}
     runs-on: ubicloud-standard-4-ubuntu-2404
     continue-on-error: true
     timeout-minutes: 20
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -179,6 +187,7 @@ jobs:
           fi
   operator-e2e:
     name: Operator e2e tests
+    if: inputs.run_operator_e2e
```
svix-jbrown:
```suggestion
    if: ${{ inputs.run_operator_e2e }}
```

I know it's not strictly necessary, but I just had a bug in another repo where a variable contained a forbidden character so got interpreted as a literal by `if` unless it was wrapped in `${{ }}` and it took me forever to figure out; best to just avoid the ambiguity IMO

.github/workflows/on-pr.yml
```diff
@@ -48,13 +49,18 @@ jobs:
               - '.github/workflows/docker-test-build.yml'
               - 'infra/operator/Dockerfile'
               - 'infra/operator/Cargo.toml'
+            operator:
+              - 'infra/operator/**'
```
svix-jbrown, line 53: I would probably also include `.github/workflows/ci.yml` and `.github/workflows/on-pr.yml` so that it gets run if we change the CI config

svix-jbrown (approved):

PR #1118, opened 2026-04-21
server: add a global request timeout

svix-jbrown:
helps provide backpressure if things are going really south


svix-jplatte:
This PR will conflict with my next one for error types. If CI wasn't failing, I'd probably just approve and merge and deal with it myself. Since it isn't, maybe it makes sense to wait for my PR and rebase this one after? It will change things a fair bit, but shouldn't be too annoying.

svix-jplatte:
The error refactor work I mentioned all landed. If you could update this to have `fn request_timeout` delegate to `server_error`, just like `fn not_ready` and `fn shutting_down`, that would make this fit right in.

svix-jbrown:
> The error refactor work I mentioned all landed. If you could update this to have `fn request_timeout` delegate to `server_error`, just like `fn not_ready` and `fn shutting_down`, that would make this fit right in.

Sorry — I realized that because we support long-polling, I need to make this more complicated to exempt the long-polling endpoints. I'll mark as draft until I come up with an ergonomic way to do this

PR #1123, opened 2026-04-22
Clean up our custom derive macros

svix-jplatte:
See individual commits.

### DIFF
--- a/crates/diom-derive/src/env_overridable.rs
+++ b/crates/diom-derive/src/env_overridable.rs
@@ -1,5 +1,6 @@
+use proc_macro2::TokenStream;
 use quote::quote;
-use syn::{DeriveInput, Ident, LitStr, Token, parenthesized, parse_macro_input};
+use syn::{DeriveInput, Ident, LitStr, Token, parenthesized};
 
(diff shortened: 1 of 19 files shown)


crates/diom-derive/src/env_overridable.rs
```diff
@@ -213,6 +211,7 @@ pub(crate) fn derive_env_overridable(input: proc_macro::TokenStream) -> proc_mac
     let name = input.ident;
 
     let expanded = quote! {
+        #[automatically_derived]
```
svix-gabriel, line 214: What does `#[automatically_derived]` do?

svix-gabriel (approved):

crates/diom-derive/src/env_overridable.rs
svix-jplatte: According to the [docs](https://doc.rust-lang.org/reference/attributes/derive.html#r-attributes.derive.automatically_derived), nothing really. I know it suppresses some lints though, to speed up compilation of the sometimes significant amounts of code produced by some proc-macros :smile:

Issue #1128, opened 2026-04-23
Clean up error reporting in the server


svix-jplatte:
For https://github.com/svix/diom/issues/302 to make sense, we also need to principled about the errors we report from the server.

PR #1129, opened 2026-04-23
Clean up error reporting for invalid partitions

svix-jplatte:
Part of #1128.

### DIFF
--- a/crates/msgs/src/entities.rs
+++ b/crates/msgs/src/entities.rs
@@ -37,6 +37,9 @@ pub const TOPIC_PARTITION_DELIMITER: &str = "~";
 pub struct Partition(u16);
 
 impl Partition {
+    pub const ZERO: Self = Self(0);
+    pub const ONE: Self = Self(1);
+
     pub fn new(index: u16) -> Result<Self, Error> {
         if index < MAX_PARTITION_COUNT {
             Ok(Self(index))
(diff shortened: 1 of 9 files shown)


crates/msgs/src/entities.rs
```diff
@@ -37,13 +43,14 @@ pub const TOPIC_PARTITION_DELIMITER: &str = "~";
 pub struct Partition(u16);
 
 impl Partition {
-    pub fn new(index: u16) -> Result<Self, Error> {
+    pub const ZERO: Self = Self(0);
+    pub const ONE: Self = Self(1);
+
+    pub(crate) fn new(index: u16) -> Option<Self> {
```
svix-james, line 49: I'm not super familiar with this code. But having `new` return `Option<Self>` seems really weird to me.

svix-jplatte: It's not too special for a type with trivial invariants. `NonZeroU16::new` also returns `Option<Self>`.

svix-james: Ok but why if we're mapping it to an error everywhere we call it anyway?

svix-jplatte: Because it depends on the call sites whether this is a bad-request or internal-server-error. I refactored away all the callsites that used this where it would have been a bad request to use `FromStr` instead, but it would be super easy for us to add new ones.

svix-james (approved):

PR #1131, opened 2026-04-23
Release v0.2.3

svix-jbrown:
I'd love to get this out today (specifically to get #1126 out)


svix-jplatte:
There's a decent chance we have broken public API since the release. Let me look over the commits since 0.2.2..

svix-jplatte:
Okay so https://github.com/svix/diom/commit/47f78ece6f0d0400a99f92d7859ca8d56b8090ce is a breaking change to some extent, it will lead to worse error reporting in SDKs, though it's specifically in case that the request doesn't pass deserialization / validation so maybe not the worst to just include it? We haven't really documented error handling yet anyways, and no tests had to be updated for that change. I'd say depends on how easy it is to release from a separate v0.2.x branch rather than main.

svix-jplatte (approved):

svix-jbrown:
> Okay so [47f78ec](https://github.com/svix/diom/commit/47f78ece6f0d0400a99f92d7859ca8d56b8090ce) is a breaking change to some extent, it will lead to worse error reporting in SDKs, though it's specifically in case that the request doesn't pass deserialization / validation so maybe not the worst to just include it? We haven't really documented error handling yet anyways, and no tests had to be updated for that change. I'd say depends on how easy it is to release from a separate v0.2.x branch rather than main.

I definitely don't think we should try to make version branches for an application without any users.

PR #1134, opened 2026-04-24
cfg: add --env-file flag for loading secrets from an envfile

tasn:
Adds a --env-file <PATH> CLI argument that loads environment variables from the given file before config is applied, giving users a dedicated place to keep secrets outside of the TOML config and instead of passing as env variables.

It kind of meaningless in how it works, as it actually sets the environment variables so it's the same as just setting the env vars directly, though we'll change that to load directly to config soon.

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -1,6 +1,7 @@
 #![warn(clippy::all)]
 #![forbid(unsafe_code)]
 
+use anyhow::Context as _;
 use clap::{Parser, Subcommand};
 use comfy_table::{Cell, Table};
 use diom_backend::{
@@ -28,6 +29,10 @@ struct Args {
     /// Path to a TOML configuration file
     #[clap(short = 'C', long)]
     config_path: Option<PathBuf>,
+
+    /// Path to a file of environment variable overrides to load in addition to .env
+    #[clap(long)]
+    env_file: Option<PathBuf>,
 
     #[clap(subcommand)]
     command: Option<Commands>,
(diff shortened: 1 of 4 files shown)


server/src/main.rs
```diff
@@ -29,6 +30,10 @@ struct Args {
     #[clap(short = 'C', long)]
     config_path: Option<PathBuf>,
 
+    /// Path to a file of environment variable overrides (like .env)
```
svix-jbrown: You should probably document that we always load `.env` if present, and load it after `--env-file` (or remove line 141)

server/src/main.rs
```diff
@@ -127,10 +132,14 @@ fn dump_variables(path: Option<PathBuf>) -> anyhow::Result<()> {
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
-    _ = dotenv();
-
     let args = Args::parse();
 
+    if let Some(env_file) = &args.env_file {
+        dotenvy::from_path(env_file).context("loading env file")?;
```
svix-jbrown, line 138: are we worried about the fact that dotenvy is technically unsafe (if someone does something like put `TZ` or another var read by C code inside an env file)?

tasn: I don't think this change makes it any worse, but added a note to Jonas: https://svixhq.slack.com/archives/C0A72988962/p1776991355515899

server/src/main.rs
tasn: Clarified, thanks! I don't think loading order should be documented in this argument, but defo should clarify this doesn't replace .env.

svix-jbrown (approved):

PR #1136, opened 2026-04-24
Operator: Fix otel var naming

svix-james:
Missing camelCase. Added test to try and catch this sort of thing. Also, made some changes to how the helm chart names the operator pod to avoid confusion.

### DIFF
--- a/infra/helm-diom/charts/crds/crds/diomclusters.json
+++ b/infra/helm-diom/charts/crds/crds/diomclusters.json
@@ -1035,12 +1035,12 @@
                         "nullable": true,
                         "type": "string"
                       },
-                      "metrics_address": {
+                      "metricsAddress": {
                         "description": "The OpenTelemetry address to send metrics to if given.\n\nIf not specified, the server will attempt to fall back\nto `opentelemetry_address`.",
                         "nullable": true,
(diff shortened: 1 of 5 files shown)


svix-jplatte (approved):
I'm guessing we went with `camelCase` for operator things due to existing k8s / helm / whatever conventions? Maybe we should set up a vaccum CI job for `infra/helm-diom/charts/crds/crds/diomclusters.json`?

svix-james:
> I'm guessing we went with camelCase for operator things due to existing k8s / helm / whatever conventions? 

Correct.

> Maybe we should set up a vaccum CI job for infra/helm-diom/charts/crds/crds/diomclusters.json?

Sorry I don't know what you mean here.

svix-jplatte:
> > Maybe we should set up a vaccum CI job for infra/helm-diom/charts/crds/crds/diomclusters.json?
> 
> Sorry I don't know what you mean here.

This file appears to be / contain an OpenAPI spec which changed in this PR. We could have vaccum, the OpenAPI linter we also use for our main spec, check this file in CI whenever it changes, to catch issues like `snake_case` field names. Just like we do for the main spec.

PR #1139, opened 2026-04-27
Improve not-found errors

svix-jplatte:
Part of #1128.

### DIFF
--- a/src/v1/endpoints/auth_token.rs
+++ b/src/v1/endpoints/auth_token.rs
@@ -136,7 +136,7 @@ async fn auth_token_create(
     let namespace: AuthTokenNamespace = state
         .namespace_state
         .fetch_namespace(data.namespace.as_ref())?
-        .ok_or_not_found()?;
+        .ok_or_not_found("namespace")?;
 
     let token = TokenPlaintext::generate(&data.prefix, data.suffix.as_deref())?;
     let operation = CreateAuthTokenOperation::new(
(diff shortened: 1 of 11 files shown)


src/v1/endpoints/auth_token.rs
```diff
@@ -136,7 +136,7 @@ async fn auth_token_create(
     let namespace: AuthTokenNamespace = state
         .namespace_state
         .fetch_namespace(data.namespace.as_ref())?
-        .ok_or_not_found()?;
+        .ok_or_not_found("namespace")?;
```
svix-james, line 139: Can we possibly factor this out into `ok_or_namespace_not_found` so that the same literal doesn't have to repeatedly written?

svix-james (dismissed):

src/v1/endpoints/auth_token.rs
svix-jplatte: I was actually thinking whether we should have that, _and_ have it return a different error code. I'll bring it up on Slack.

svix-pluiz (approved):
The proposed solution is fine for now, but we could either use constants or specific functions, as suggested by James Lucas, to handle errors.

Something like this:
```rs
pub const ACCESS_POLICY_LITERAL = "access policy"
```
[...]
```rs
.ok_or_not_found(ACCESS_POLICY_LITERAL)?;
```

PR #1149, opened 2026-04-30
go: Revamp error types

svix-jplatte:
… and two drive-by improvements.

Part of https://github.com/svix/diom/issues/302.


z-clients/rust/src/error.rs
```diff
@@ -135,7 +135,7 @@ impl fmt::Display for Error {
             ErrorKind::ServerError(e) => write!(f, "server error {e}"),
             // don't print inner errors that are returned from Error::source
             ErrorKind::Connection(_) => write!(f, "connection error"),
-            ErrorKind::Other(_) => write!(f, "other"),
+            ErrorKind::Other(_) => write!(f, "internal error"),
         }
```
svix-pluiz, line 139: intentional?

svix-jplatte: Yes, see commit message. I looked back at what I had done in Rust after doing the Go bits, and decided the Go one was better.

svix-pluiz (approved):

PR #1150, opened 2026-04-30
cluster: add otel metrics for interserver requests

svix-jbrown:
that might help us debug a bit

### DIFF
--- a/src/core/metrics.rs
+++ b/src/core/metrics.rs
@@ -9,7 +9,7 @@ use opentelemetry::{
 use super::cluster::NodeId;
 
 #[derive(Debug, Clone, Copy, strum::IntoStaticStr)]
-#[strum(serialize_all = "snake_case")]
+#[strum(serialize_all = "kebab-case")]
 pub enum DbType {
     Persistent,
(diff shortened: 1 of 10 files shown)


src/core/metrics.rs
```diff
@@ -588,3 +582,77 @@ impl openraft::metrics::MetricsRecorder for OpenraftMetrics {
...
+pub struct ClusterNetworkMetrics {
+    requests: Counter<u64>,
+    unaddressed_requests: Counter<u64>,
+    request_latency: Histogram<u64>,
+
+    source_node_id: KeyValue,
+}
+
+#[derive(Clone, strum::IntoStaticStr)]
+#[strum(serialize_all = "snake_case")]
```
svix-jplatte: For API, we use kebab-case, maybe we should replicate that for config and metric things? No strong opinion.

svix-jplatte (dismissed):

src/core/metrics.rs
svix-jbrown: ok

svix-jplatte (approved):

PR #1152, opened 2026-04-30
cluster: fix bootstrap running too early

svix-jbrown:
This was an interesting bug that could happen and cause a crash on startup.

Imagine a three node cluster with nodes A, B, and C.

Node A is the leader.
Node B shuts down
Node C becomes the leader (for whatever reason)
Node B starts up

At this point, `is_up` would return on Node B quickly because we still think it's a Follower of Node A. As soon as we start receiving traffic, we'll be proven wrong here. Instead, let's wait until our theoretical leader actually writes some logs to us before we call ourselves "up".


svix-james (approved):
Uhhproved

PR #1153, opened 2026-04-30
cluster: allow forwarded messages to be forwarded twice

svix-jbrown:
During a leadership change, forward-to-leader requests may fail because they'll go to a stale leader and we don't allow re-forwarding. This adds a new `hop_ttl` field to the forward request which allows us to forward multiple times without looping forever.

### DIFF
--- a/src/core/cluster/handle.rs
+++ b/src/core/cluster/handle.rs
@@ -28,6 +28,7 @@ use tokio::sync::mpsc::Sender;
 
 tokio::task_local! {
     pub(super) static APPLIED_LOG_ID: Mutex<Option<LogId>>;
+    pub(super) static FORWARD_HOP_COUNT: Mutex<Option<usize>>;
 }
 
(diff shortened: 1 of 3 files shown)


src/core/cluster/handle.rs
```diff
@@ -28,6 +28,7 @@ use tokio::sync::mpsc::Sender;
 
 tokio::task_local! {
     pub(super) static APPLIED_LOG_ID: Mutex<Option<LogId>>;
+    pub(super) static FORWARD_HOP_COUNT: Mutex<Option<usize>>;
```
svix-jplatte: This doesn't appear to be used.

svix-jbrown: oh yeah the first version of this captured it as a header, then I remembered I can just put it in the body. Will remove.

svix-jplatte (approved):

PR #1166, opened 2026-05-01
feat: add a system for exposing otel metrics on-demand

svix-jbrown:
while trying to debug stuff, it was driving me a bit crazy to wait for prometheus to get new series delivered to it.

This adds an endpoint which dumps the most recent metrics.

Depends on svix/openapi-codegen#159; the PR is currently using an un-landed hash of openapi-codegen.

I put this under admin instead of health because it includes some privileged things, like the names of topics and consumer groups.

This gets us like 90% of the way to https://github.com/svix/diom/issues/70#issuecomment-3842838450


server/src/otel.rs
```diff
@@ -152,29 +155,45 @@ pub(crate) fn setup_metrics(cfg: &ConfigurationInner) {
...
+    let most_recent_reader = PeriodicReader::builder(most_recent_store.clone(), runtime::Tokio)
+        .with_interval(cfg.opentelemetry.metrics_period.into())
+        .build();
 
-        opentelemetry::global::set_meter_provider(provider);
-    };
+    let mut provider_builder = SdkMeterProvider::builder().with_reader(most_recent_reader);
+
+    if let Some(reader) = reader {
+        provider_builder = provider_builder.with_reader(reader)
```
svix-jplatte:
```suggestion
        provider_builder = provider_builder.with_reader(reader);
```

server/src/otel.rs
```diff
@@ -122,8 +125,8 @@ pub(crate) fn setup_tracing(
     (dispatch.into(), otel_tracer_provider)
 }
 
-pub(crate) fn setup_metrics(cfg: &ConfigurationInner) {
-    if let Some(addr) = cfg
+pub(crate) fn setup_metrics(cfg: &ConfigurationInner) -> MostRecentMetricStore {
+    let reader = if let Some(addr) = cfg
```
svix-jplatte: Can use `Option::map`.

src/v1/endpoints/admin/metrics.rs
```diff
@@ -0,0 +1,93 @@
...
+            })
+            .collect();
+        Self { metrics }
+    }
+}
+
+#[aide_annotate(op_id = "v1.admin.metrics.get")]
+/// Dump the current metrics (which would otherwise be sent to the OTLP metrics receiver)
+async fn get_metrics(State(state): State<AppState>) -> Result<MsgPackOrJson<GetMetricsOut>> {
+    let metrics = if let Some(metrics) = state.metrics.as_ref() {
```
svix-jplatte:
```suggestion
    let metrics = if let Some(metrics) = &state.metrics {
```

src/metrics.rs
```diff
@@ -0,0 +1,372 @@
...
+        let id = InternHandle(self.values.len());
+        self.values.push(s.to_owned());
+        self.map.insert(s.to_owned(), id);
+        id
+    }
+
+    fn redeem(&self, handle: InternHandle) -> Option<&str> {
+        self.values.get(handle.0).map(|s| s.as_str())
+    }
+}
```
svix-jplatte, line 47: Some docs for this would be nice.

src/metrics.rs
```diff
@@ -0,0 +1,372 @@
...
+    sync::Arc,
+};
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+struct InternHandle(usize);
+
+#[derive(Default)]
+struct StringStore {
+    values: Vec<String>,
+    map: HashMap<String, InternHandle>,
```
svix-jplatte: I'm sure we already depend on `indexmap` one way or the other, if we used that we wouldn't have to duplicate the keys (just store an `IndexSet` as the inner representation and use len comparison + usize indexing for access).

src/metrics.rs
```diff
@@ -0,0 +1,372 @@
...
+        let key_values = rv
+            .iter()
+            .filter_map(|(k, v)| {
+                if let Some(key) = ss.redeem(*k)
+                    && let Some(value) = ss.redeem(*v)
+                {
+                    Some((key, value))
+                } else {
+                    None
+                }
```
svix-jplatte: It's a bit shorter, but feel free to ignore the suggestion if you like the current version better

```suggestion
                ss.redeem(*k).and_then(|key| {
                    let value = ss.redeem(*v)?;
                    Some((key, value))
                })
```

src/metrics.rs
```diff
@@ -0,0 +1,372 @@
...
+}
+
+impl Metric {
+    /// The ID of a series is the name, combined with the unique set of k/v pairs in the resource
+    /// list
+    fn stable_id(&self) -> u64 {
+        let mut hasher = std::hash::DefaultHasher::new();
+        hasher.write_usize(self.name.0);
+        hasher.write_usize(self.resources.len());
+        for (k, v) in self.resources.iter() {
```
svix-jplatte:
```suggestion
        for (k, v) in &self.resources {
```

src/metrics.rs
```diff
@@ -0,0 +1,372 @@
...
+        point_attributes: I,
+        metric_type: MetricType,
+    ) where
+        T: Into<MetricDatum> + Copy + 'static,
+        I: Iterator<Item = &'b opentelemetry::KeyValue>,
+    {
+        let mut resources = self.scope_attributes.to_owned();
+        resources.extend(point_attributes.map(|attribute| {
+            let key = self.strings.intern(attribute.key.as_str());
+            let value = self.strings.intern(&attribute.value.as_str());
```
svix-jplatte: Surely `as_str()` returns a `&str` and we don't need to take another reference here?

```suggestion
            let value = self.strings.intern(attribute.value.as_str());
```

src/metrics.rs
```diff
@@ -0,0 +1,372 @@
...
+        let id = metric.stable_id();
+        self.metrics.insert(id, metric);
+    }
+
+    fn visit_gauge<T>(&mut self, gauge: &opentelemetry_sdk::metrics::data::Gauge<T>)
+    where
+        T: Into<MetricDatum> + Copy + 'static,
+    {
+        let Ok(time) = gauge.time().try_into() else {
+            return;
```
svix-jplatte, line 227: Maybe log sth. when the `try_into()` fails?

src/metrics.rs
```diff
@@ -0,0 +1,372 @@
...
+            );
+        }
+    }
+
+    fn visit_counter<T>(&mut self, sum: &opentelemetry_sdk::metrics::data::Sum<T>)
+    where
+        T: Into<MetricDatum> + Copy + 'static,
+    {
+        let Ok(time) = sum.time().try_into() else {
+            return;
```
svix-jplatte, line 245: (same here)

src/metrics.rs
```diff
@@ -0,0 +1,372 @@
...
+        Self::new()
+    }
+}
+
+impl MostRecentMetricStore {
+    pub fn new() -> Self {
+        Self {
+            inner: Arc::new(Mutex::new(MostRecentMetricStoreInner::default())),
+        }
+    }
```
svix-jplatte:
```suggestion
#[derive(Clone, Default)]
pub struct MostRecentMetricStore {
    inner: Arc<Mutex<MostRecentMetricStoreInner>>,
}

impl MostRecentMetricStore {
    pub fn new() -> Self {
        Self::default()
    }
```

src/metrics.rs
```diff
@@ -0,0 +1,372 @@
...
+struct InternHandle(usize);
+
+#[derive(Default)]
+struct StringStore {
+    values: Vec<String>,
+    map: HashMap<String, InternHandle>,
+}
+
+impl StringStore {
+    fn intern(&mut self, s: &str) -> InternHandle {
```
svix-jplatte: Taking `impl Into<String>` would allow passing the `Cow<'_, str>`s from `opentelemetry::common::Value::as_str` directly and potentially avoid an allocation.

src/metrics.rs
svix-jbrown: the issue is thas `Value::as_str()` is poorly named and gives you back a `Cow<'_, str>` and we need a `&str`, so we actually do need the extra `&` to coerce it. I'll just do your suggestion below to parameterize the `intern` method

src/metrics.rs
svix-jbrown: oh yeah I deleted all the other fields didn't I

src/metrics.rs
svix-jbrown: ok

svix-jplatte (approved):

PR #1168, opened 2026-05-04
python: Revamp error types

svix-jplatte:
Part of https://github.com/svix/diom/issues/302.


z-clients/python/diom/internal/errors.py
```diff
@@ -0,0 +1,59 @@
...
+def raise_diom_error(body: ErrorBody, op_id) -> t.NoReturn:
+    match body.type:
+        case "invalid-input":
+            raise InvalidInput(op_id, body.code, body.detail, body.location)
+        case "operation-error":
+            raise OperationError(op_id, body.code, body.detail, body.location)
+        case "server-error":
+            raise ServerError(op_id, body.code, body.detail, body.location)
+        case _:
+            raise OtherError(op_id) from ValueError(f"invalid error type `{body.type}`")
```
svix-jplatte, line 59: Is this a good idea? Should I instead attach the error message to `OtherError` directly? Its superclass `Exception` has an `add_note` method that I guess could be used here.

z-clients/python/diom/internal/errors.py
```diff
@@ -0,0 +1,59 @@
...
+def raise_diom_error(body: ErrorBody, op_id) -> t.NoReturn:
+    match body.type:
+        case "invalid-input":
+            raise InvalidInputError(op_id, body.code, body.detail, body.location)
+        case "operation-error":
+            raise OperationError(op_id, body.code, body.detail, body.location)
+        case "server-error":
+            raise ServerError(op_id, body.code, body.detail, body.location)
+        case _:
+            raise OtherError(op_id) from ValueError(f"invalid error type `{body.type}`")
```
svix-gabriel, line 59: Why not say `unexpected error type`, in the event we add more error variants to https://github.com/svix/diom/blob/4e0db5fa44891c8a6fd1970e125e96e3244b52eb/crates/diom-proto/src/error.rs#L8 later but clients are slow to update?

z-clients/python/diom/internal/errors.py
svix-gabriel: I don't have a strong opinion either way, maybe just because python isn't my native tongue, but both what you did and the alternate suggested sound reasonable?

svix-gabriel (approved):

z-clients/python/diom/internal/errors.py
svix-jplatte: The idea is that we _never_ add more error types (only error codes). See also WIP docs at https://github.com/svix/diom-docs/pull/6.

svix-gabriel: ok

z-clients/python/diom/internal/errors.py
svix-jbrown: `add_note` only landed in 3.11 and I don't think I've ever seen anything use it, so I wouldn't even think to look in `.notes` as an application developer. I think what you did here is good.

svix-jbrown (approved):

PR #1172, opened 2026-05-04
cfg: allow disabling storing otel context in raft logs

svix-jbrown:
for small requests, the big opentelemetry string/string hashmap can be more than half of the log size. It's really only useful when doing single-node performance investigation anyway.

### DIFF
--- a/src/core/cluster/handle.rs
+++ b/src/core/cluster/handle.rs
@@ -374,11 +374,11 @@ impl RaftState {
     {
         let inner: Request = op.into().into();
         let now = self.time.update_now();
-        let ctx = if self.cfg.cluster.forward_opentelemetry_context {
-            Some(opentelemetry::Context::current().into())
-        } else {
-            None
-        };
+        let ctx = self
+            .cfg
+            .cluster
+            .forward_opentelemetry_context
+            .then(|| opentelemetry::Context::current().into());
         let request = RequestWithContext::new(inner, now.into(), ctx);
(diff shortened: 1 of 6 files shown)


src/core/cluster/handle.rs
```diff
@@ -374,11 +374,12 @@ impl RaftState {
...
-        let request = RequestWithContext::new(
-            inner,
-            now.into(),
-            Some(opentelemetry::Context::current().into()),
-        );
+        let ctx = if self.cfg.cluster.forward_opentelemetry_context {
+            Some(opentelemetry::Context::current().into())
+        } else {
+            None
+        };
```
svix-jplatte: Could use `bool::then` 🤓

src/core/cluster/handle.rs
```diff
@@ -823,11 +824,12 @@ impl diom_operations::OperationWriterBase for RaftState {
...
-        let request = RequestWithContext::new(
-            request,
-            now.into(),
-            Some(opentelemetry::Context::current().into()),
-        );
+        let ctx = if self.cfg.cluster.forward_opentelemetry_context {
+            Some(opentelemetry::Context::current().into())
+        } else {
+            None
+        };
```
svix-jplatte: Same.

svix-jplatte (dismissed):

src/core/cluster/handle.rs
svix-jbrown: ok

svix-jplatte (approved):

Issue #302, opened 2026-03-02
SDKs should have good error types for the kinds of error each route is allowed?


tasn:
Or at least differentiate between types of errors?

svix-jplatte:
For languages with exceptions, make sure to have different exception types for different kinds of errors (see `ErrorKind`s server-side). For Rust and similar, will be an enum.

svix-jplatte:
Done!

PR #1179, opened 2026-05-05
java: Revamp error types

svix-jplatte:
Last part of #302.


z-clients/java/src/main/java/com/svix/diom/DiomException.java
```diff
@@ -0,0 +1,40 @@
...
+        final String location
+    ) {
+        StringBuilder b = new StringBuilder("code=");
+        b.append(code);
+        if (location != null) {
+            b.append(" location=");
+            b.append(location);
+        }
+        b.append(" detail=\"");
+        b.append(detail.replace("\"", "\\\""));
```
svix-pluiz, line 36: almost related: I'm not sure if JS SDK is adding quotes here https://github.com/svix/diom/pull/1174/changes#diff-84e6c58ea078e6f47509c5b50e71168995755e26567b11cacc1eaeb47aa9f7c6R62-R63

svix-pluiz, line 36: `String` in Java can be null, we should check if it's defined and not an empty string before appending to the response message

svix-jplatte: I don't think we need to. The exception constructors are package-private and if the server ever returned `null` for any of the error fields (other than `location`) that would be a serious bug - I don't think it matters that much if that results in the exception constructor throwing a `NullPointerException`.

svix-pluiz (approved):

PR #1183, opened 2026-05-05
Fix error code for serde value errors from rmp-serde

svix-jplatte:
Something I noticed when writing the error test for Java (https://github.com/svix/diom/pull/1179).
Need to rebase that PR if this merges first, or the other way around.


svix-gabriel (approved):

svix-jplatte:
Forgot it earlier, but open source review for `assert_matches2`: It's a tiny macro-only crate written by yours truly. I think we can depend on it :smile:

PR #1188, opened 2026-05-05
add support for tokio-console

svix-jbrown:
it's behind a feature flag


svix-jplatte (dismissed):
GitHub is not letting me send an inline comment. I was going to say

> Nit: I'd inline the `.into()` in the tuple constructor below.

about the `let dispatch = dispatch.into();` line.

svix-jplatte (approved):

PR #1189, opened 2026-05-05
cluster: rewrite leader-only background job trigger

svix-jbrown:
openraft 0.10 has this handy [`on_cluster_leader_change`](https://docs.rs/openraft/0.10.0-alpha.19/openraft/raft/struct.Raft.html#method.on_cluster_leader_change) thing that lets us avoid waking up on every single raft metrics update.

### DIFF
--- a/src/core/cluster/background.rs
+++ b/src/core/cluster/background.rs
@@ -1,10 +1,7 @@
-use std::{
-    sync::{Arc, Mutex},
-    time::Duration,
-};
+use std::time::Duration;
 
 use super::{
-    LogId, NodeId,
+    LogId,
     handle::{BackgroundCommand, RaftState},
     operations::{RecordLogTimestampOperation, TickOperation},
     raft::TypeConfig,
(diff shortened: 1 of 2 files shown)


src/core/cluster/background.rs
```diff
@@ -129,143 +133,148 @@ impl BackgroundJobRunner {
...
+                    Some(BackgroundJobLeaderMessage::StopBeingLeader) => {
+                        runner.stop_all().await?;
+                    }
+                    None => {
+                        tracing::warn!("leader detection died");
+                        break;
+                    }
+                }
+            },
+            res = runner.jobs.join_next(), if !runner.jobs.is_empty() => {
```
svix-jplatte, line 245: This new `if` means we don't run this select branch for the very last job to exit, I guess? Why is that?

svix-jbrown: `join_next` immediately returns `Ready(None)` if the jobs set is empty, which turns this select loop into an infinite loop and borks the server. :-/

svix-jplatte: I checked docs, so the `if` is actually a _pre_-condition. `join_next` is only called if this the set isn't empty and after it returns, the condition isn't checked again. Man, I hate this macro, it's so unintuitive.

svix-jplatte (approved):

src/core/cluster/background.rs
svix-jbrown: ask me how long it took me to figure out that the `,` in the macro before the if is load-bearing and not a documentation artifact

PR #1192, opened 2026-05-06
add an explicit end-to-end test for replication

svix-jbrown:


crates/test-utils/src/json.rs
```diff
@@ -12,6 +13,13 @@ impl JsonFastAndLoose for serde_json::Value {
         self.as_u64().unwrap()
     }
 
+    #[track_caller]
+    fn assert_u8(&self) -> u8 {
+        let value = self.as_u64().unwrap();
+        assert!(value < 256);
+        value as _
```
svix-jplatte:
```suggestion
        self.assert_u64().try_into().unwrap()
```

crates/test-utils/src/server.rs
```diff
@@ -341,3 +342,102 @@ pub fn default_server_config(workdir: &Path) -> ConfigurationInner {
...
 }
+
+pub struct ClusterTestContext {
+    pub handles: BTreeMap<NodeId, TestContext>,
+    pub admin_token: String,
+}
+
+impl ClusterTestContext {
+    /// Return a reference to a client pointed at what it currently
+    /// the leader
```
svix-jplatte: This is not a correct sentence

crates/test-utils/src/server.rs
```diff
@@ -341,3 +342,102 @@ pub fn default_server_config(workdir: &Path) -> ConfigurationInner {
...
+
+    run_with_retries(async || {
+        let cluster_status = client
+            .get("v1.cluster-admin.status")
+            .await?
+            .ensure(StatusCode::OK)?
+            .json();
+        let Some(nodes) = cluster_status["nodes"].as_array() else {
+            anyhow::bail!("invalid cluster_status output");
+        };
```
svix-jplatte: Could use assert_array. If this is not an array, that's clearly a bug and not something to retry on, right?

tests/it/cluster_admin.rs
```diff
@@ -52,78 +49,28 @@ async fn test_cluster_status() -> TestResult {
...
-    let resp = client
+    let resp = leader_client
         .post("v1.cluster-admin.remove-node")
-        .json(json!({"node_id": second_node_id}))
+        .json(json!({"node_id": follower_id}))
         .await?
         .expect(StatusCode::OK)
         .json();
-    assert_eq!(resp["node_id"].assert_str(), second_node_id.to_string());
+    assert_eq!(resp["node_id"].assert_str(), follower_id.to_string());
```
svix-jplatte:
```suggestion
    assert_eq!(resp["node_id"], follower_id.to_string());
```

Value can be compared with string directly for the same effect.

crates/test-utils/src/json.rs
svix-jbrown: decided to do it a bit of a different way

svix-jplatte (approved):

PR #1197, opened 2026-05-07
cluster: do not block as much for snapshots

svix-jbrown:
The previous way I was doing snapshots ended up holding a read lock on the RwLock for the duration of the snapshot; this is dumb, because `apply_` requires a write lock, so we block apply for way too long.

The locks are now much more explicit in the code and the handle's rwlock shouldn't ever be held during the slow part of snapshot-building now.

### DIFF
--- a/src/core/cluster/state_machine.rs
+++ b/src/core/cluster/state_machine.rs
@@ -212,6 +212,8 @@ static LAST_SNAPSHOT: FjallFixedKey<LastSnapshot> = FjallFixedKey::new("last_sna
 static LAST_MEMBERSHIP: FjallFixedKey<StoredMembership> = FjallFixedKey::new("last_membership");
 static CLUSTER_UUID: FjallFixedKey<ClusterId> = FjallFixedKey::new("cluster_uuid");
 
+type SnapshotTarget = (StorageType, Database, fjall::Snapshot, Vec<String>);
+
 impl Store {
     pub async fn new(
         persistent_db: Database,
@@ -639,10 +641,6 @@ impl Store {
         })
     }
 
-    fn prep_snapshot_builder_(&mut self) {
-        self.snapshot_idx += 1;
-    }
-
     #[tracing::instrument(skip_all)]
     async fn get_current_snapshot_(&mut self) -> anyhow::Result<Option<Snapshot>> {
         // clone to avoid holding a lock over an await point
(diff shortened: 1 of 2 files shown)


src/core/cluster/state_machine.rs
```diff
@@ -871,28 +805,109 @@ impl StoredSnapshot {
...
+                    dbs.ephemeral.snapshot(),
+                    list_keyspaces(&dbs.ephemeral),
+                ),
+            ]
+        })
+        .await
+        .context("failed to generate snapshot targets")?;
+
+        // release the lock while we actually serialize the keyspace
+        drop(guard);
```
svix-jplatte: Would it be annoying to extract everything from acquiring the guard up until here to a separate method that drops the guard automatically when it ends? I'd find that a bit cleaner.

src/core/cluster/state_machine.rs
```diff
@@ -871,28 +805,109 @@ impl StoredSnapshot {
...
+        let snapshot = StoredSnapshot::new(&meta, &self.snapshot_directory, targets)
             .await
-            .map_err(io_err)?;
-        // but set the last_snapshot_ field with a write lock to serialize
+            .context("failed to build snapshot")?;
+
+        let path = snapshot.path.clone();
+
+        let size =
+            spawn_blocking_in_current_span(move || std::fs::metadata(&path).map(|m| m.len()))
```
svix-jplatte: Why not use `tokio::fs::metadata`?

src/core/cluster/state_machine.rs
```diff
@@ -871,28 +805,109 @@ impl StoredSnapshot {
...
+                .context("getting size of snapshot")?;
+
+        let runtime = start.elapsed();
+        let path = snapshot.path.clone();
+
+        tracing::debug!(
+            ?path,
+            last_log_id = ?meta.last_log_id(),
+            size,
+            runtime = ?runtime,
```
svix-jplatte:
```suggestion
            ?runtime,
```

src/core/cluster/state_machine.rs
svix-jbrown: lmk what you think

src/core/cluster/state_machine.rs
```diff
@@ -875,28 +857,62 @@ impl StoredSnapshot {
...
+            .context("failed to build snapshot")?;
+
+        let path = snapshot.path.clone();
+
+        let size = tokio::fs::metadata(&path)
             .await
-            .map_err(io_err)?;
-        // but set the last_snapshot_ field with a write lock to serialize
+            .map(|m| m.len())
+            .context("getting size of snapshot")?;
```
svix-jplatte:
```suggestion
        let size = tokio::fs::metadata(&path)
.await
            .context("getting size of snapshot")?
            .len();
```

svix-jplatte (dismissed):

src/core/cluster/state_machine.rs
svix-jplatte: Wow GH makes a mess out of this suggestion. The idea is just that I'm pretty sure you can move the `.len()` call out of the `.map` and chain it off of the `.context()?` instead.

svix-jbrown: ok

svix-jplatte (approved):

PR #1198, opened 2026-05-07
cluster: rewrite is_up function

svix-jbrown:
the polling was unreliable. I think this will work better.

### DIFF
--- a/src/bootstrap.rs
+++ b/src/bootstrap.rs
@@ -326,9 +326,7 @@ async fn wait_for_up(config: &AppConfig, raft_state: &RaftState) -> anyhow::Resu
         shutdown.run_until_cancelled(raft_state.wait_for_up()).await
     }
     .ok_or_else(|| anyhow::anyhow!("node shut down before bootstrapping finished"))?;
-    if !value {
-        anyhow::bail!("node failed to come up");
-    }
+    anyhow::ensure!(value, "node has failed to come up");
     Ok(())
 }
 
--- a/src/bootstrap.rs
+++ b/src/bootstrap.rs
@@ -315,31 +315,21 @@ fn load_commands(
 }
 
 async fn wait_for_up(config: &AppConfig, raft_state: &RaftState) -> anyhow::Result<()> {
-    let mut deadline: std::pin::Pin<Box<dyn Future<Output = ()> + Send>> =
-        if let Some(time) = config.bootstrap_max_wait_time {
-            Box::pin(tokio::time::sleep(time.into()))
-        } else {
-            Box::pin(futures_util::future::pending())
-        };
     let shutdown = crate::shutting_down_token();
-    let mut interval = tokio::time::interval(std::time::Duration::from_millis(100));
-
-    loop {
-        tokio::select! {
-            _ = shutdown.cancelled() => {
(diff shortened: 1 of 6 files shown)


src/core/cluster/raft.rs
```diff
@@ -65,6 +65,93 @@ pub(crate) async fn initialize_cluster(
...
+    handle: Option<openraft::raft::WatchChangeHandle<TypeConfig>>,
+}
+
+impl RaftStateWatcherInner {
+    fn recompute_ready(&mut self) {
+        let new_ready = self.leader.is_some() && (self.is_single_node || self.has_applied_log);
+        if new_ready != self.ready {
+            self.ready = new_ready;
+            self.tx.send_replace(new_ready);
+        }
```
svix-jplatte: This looks like the API of my observables library (https://docs.rs/eyeball/latest/eyeball/struct.Observable.html) would be a better fit (it has `set_if_not_eq`). Though maybe better to avoid the extra dependency :)

Tokio also has a `send_if_modified` that looks applicable here, but probably not worth codegolfing this.

src/bootstrap.rs
```diff
@@ -315,31 +315,21 @@ fn load_commands(
...
+            .await
+            .transpose()
+            .context("waiting for node to be up for bootstrapping")?
+    } else {
+        shutdown.run_until_cancelled(raft_state.wait_for_up()).await
+    }
+    .ok_or_else(|| anyhow::anyhow!("node shut down before bootstrapping finished"))?;
+    if !value {
+        anyhow::bail!("node failed to come up");
     }
```
svix-jplatte:
```suggestion
    anyhow::ensure!(value, "node failed to come up");
```

svix-jplatte (dismissed):

src/core/cluster/raft.rs
svix-jbrown: `eyeball` looks cool; does the subscriber have an atomic "wait-for-value" option, though?

It seems like any cycle that looks like 

```
if subscriber.get() {
   return true;
}
loop {
    let Some(value) = subscriber.next().await { else return false; };
    if value {
        return true
    }
}
```

would be inherently racy because the value could turn to `true` between the first check and the first wait

src/core/cluster/raft.rs
```diff
@@ -66,42 +66,35 @@ pub(crate) async fn initialize_cluster(
 }
 
 struct RaftStateWatcherInner {
-    ready: bool,
     has_applied_log: bool,
     is_single_node: bool,
     leader: Option<(NodeId, u64)>,
-    tx: tokio::sync::watch::Sender<bool>,
+    is_ready: eyeball::SharedObservable<bool>,
```
svix-jplatte: Since you're only updating it through `&mut` and never clone it (right?), you can use the "basic" `Observable` type instead of `SharedObservable`. Though it will be `Observable::set_if_not_eq(&mut self.is_ready, new_ready);` then, as receiver-style methods are officially verboten on `Deref` types.

svix-jplatte (dismissed):

src/core/cluster/raft.rs
svix-jbrown: ok

svix-jplatte (approved):

PR #1201, opened 2026-05-07
cluster: more snapshot work

svix-jbrown:
The main thing here is that the network client for the install_snapshot RPC needs to actually parse out the error, because the snapshot installer is the only thing where the caller actually _reads_ the error body sent back by the RPC server. Specifically, if a node is restarted in the middle of loading a snapshot, it uses this to signal to the other peer node to restart snapshot transmission. If this message isn't parsed correctly, the other node ends up in an infinite loop trying to send the same end-of-file segment over and over again until you restart them all.

Our implementation (`rpc_response` in `app.rs`) only emits 500's when there's an underlying response; we do emit some other status codes (e.g., if we receive a request to a node that's shutting down or booting).

Other fixes:

 - Some more logging
 - Remove a duplicated call to `delete_unused_snapshots`

I'm going to make this work, dangit

### DIFF
--- a/src/core/cluster/network.rs
+++ b/src/core/cluster/network.rs
@@ -11,7 +11,7 @@ use crate::{
 use super::{LogId, Node, NodeId, proto, raft::TypeConfig};
 use anyhow::Context;
 use diom_proto::prelude::*;
-use http::{HeaderMap, HeaderValue, header};
+use http::{HeaderMap, HeaderValue, StatusCode, header};
 use openraft::{
     RaftNetworkFactory, RaftNetworkV2,
     error::{NetworkError, Unreachable},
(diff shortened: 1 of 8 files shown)


src/core/cluster/network.rs
```diff
@@ -23,6 +23,31 @@ use tap::Pipe;
...
+}
+
+impl std::fmt::Display for BadStatusError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "BadStatusError({})", self.0)
+    }
+}
+
+impl std::error::Error for BadStatusError {
+    fn description(&self) -> &str {
```
svix-james: This is deprecated, right?

src/core/cluster/network.rs
```diff
@@ -168,33 +193,51 @@ impl NetworkClient {
...
+                let err_response = response.msgpack().await.map_err(|err| {
+                    tracing::warn!(?err, ?status, "error deserializing response body");
+                    self.metrics.record_request(
+                        self.target,
+                        ClusterRequestStatus::DeserializationError,
+                        start.elapsed(),
+                    );
+                    RPCError::Network(NetworkError::new(&err))
+                })?;
+                let error = openraft::error::RemoteError::new(self.target, err_response);
```
svix-james, line 217: Are we missing a metric here?

src/core/cluster/network.rs
svix-jbrown: I guess so; I always forget how the core traits have changed in the last couple of years

svix-james (approved):

PR #1207, opened 2026-05-15
Msgs: Add lease cancellation operation

svix-james:
Add `v1.msgs.stream.cancel-lease` so that a lease can be aborted prematurely if needed.


crates/msgs/src/operations/stream/cancel_lease.rs
```diff
@@ -0,0 +1,89 @@
...
+                .metrics
+                .record_stream_lease_cancelled(&self.topic.topic, &self.consumer_group);
+            Ok(StreamCancelLeaseResponseData {})
+        })
+        .await?
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct StreamCancelLeaseResponseData {}
```
svix-jbrown, line 79: should this include a field to indicate whether or not something was actually cancelled? maybe even the offset(s) of the cancelled lease?

svix-james: I based this on the other stream endpoints, which in most cases don't return anything on successful operations. Plus, this should error if there is not an available lease, so if it succeeds it will have cancelled something.

svix-jbrown: okey

svix-jbrown (approved):

PR #1212, opened 2026-05-19
Some API reference improvements

svix-jplatte:
Side improvements from https://github.com/svix/monorepo-private/issues/12997.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -55,27 +55,34 @@ fn test_setup() {
 
 mod docs {
     use aide::{axum::ApiRouter, openapi::OpenApi};
-    use axum::{
-        response::{Html, IntoResponse, Redirect},
-        routing::get,
-    };
+    use axum::{response::Redirect, routing::get};
 
-    // TODO: switch to generated docs instead of hardcoded JSON once generated
-    // is comparable/better than hardcoded one.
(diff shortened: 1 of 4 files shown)


src/static/docs.html
```diff
@@ -115,16 +122,18 @@
   </header>
   <div id="app"></div>
   <!-- Load the Script -->
-  <script src="https://cdn.jsdelivr.net/npm/@scalar/api-reference"></script>
+  <script src="https://cdn.jsdelivr.net/npm/@scalar/api-reference@1.57.0"></script>
```
svix-jbrown: can we use [SRI](https://www.jsdelivr.com/using-sri-with-dynamic-files) for this now that we're pinning a verison?

src/lib.rs
```diff
@@ -55,27 +55,34 @@ fn test_setup() {
...
-    use axum::{
-        response::{Html, IntoResponse, Redirect},
-        routing::get,
-    };
+    use axum::{response::Redirect, routing::get};
 
-    // TODO: switch to generated docs instead of hardcoded JSON once generated
-    // is comparable/better than hardcoded one.
     pub(crate) fn router(_docs: OpenApi) -> ApiRouter {
+        cfg_select! {
```
svix-jbrown, line 61: 😍

svix-jbrown (approved):

PR #1222, opened 2026-05-19
make msgs use the same otel meter as everything else

svix-jbrown:
this doesn't have any observable changes since we create the shared MeterProvider from the globally-registered SdkMeterProvider, but it's still good practice


svix-james (dismissed):

svix-james:
... but Clippy.

svix-jbrown:
oh, sorry, I thought I made this a draft. It depends on the opentelemetry dependency upgrade.

svix-jplatte (approved):

PR #1225, opened 2026-05-20
Add JavaScript code samples

svix-jplatte:
… and some drive-by fixes.

Part of https://github.com/svix/monorepo-private/issues/12997.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -77,6 +77,11 @@
         ],
         "x-codeSamples": [
           {
+            "lang": "TypeScript",
+            "label": "JavaScript",
+            "source": "const response = await diom.clusterAdmin.status();"
+          },
+          {
             "lang": "Rust",
             "label": "Rust",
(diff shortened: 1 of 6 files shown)


openapi.json
```diff
@@ -160,6 +165,11 @@
           }
         ],
         "x-codeSamples": [
+          {
+            "lang": "TypeScript",
+            "label": "JavaScript",
+            "source": "const response = await diom.clusterAdmin.initialize(\n    {\n    },\n);"
```
svix-lucho: `diom.clusterAdmin.initialize({}, );`

is this valid? the `,` and then nothing?

svix-jplatte: It's probably valid, but obviously ugly. Just missed it, I'll fix it.

svix-lucho: also, if I'm being nitpicky the newlines for empty objects and empty params are a bit ugly, but we can leave that for another time

svix-lucho (approved):

PR #1230, opened 2026-05-22
Bugfix: track max offset separately, to avoid resetting offsets after msgs are deleted.

### LINKED ISSUE #1204: Msgs module: Don't lose the offset when hard-deleting messages from a topic
If a partition is empty after deleting messages, we should preserve the highest offset in some capacity so that new messages don't restart from offset 0.

svix-gabriel:
Closes https://github.com/svix/diom/issues/1204

Currently, if a topic has all of its messages hard-deleted, we inadvertently reset the offset to zero.

This fixes that bug by separately tracking the offset, or "high water mark", so it's not deleted. We then get the highest offset by referencing the HWM.

There's a test to validate the behavior.

I also added a pretty rudimentary cache for the high water mark.
* The cache self clears in the same background task where we delete expired messages, to avoid unbounded growth. We could've gone with something fancier, but I opted for simplicity here.
* I opted for dashmap, because it was already an indirect dependency from fjall.


crates/msgs/src/storage.rs
```diff
@@ -11,16 +11,23 @@ use serde::{Deserialize, Serialize};
...
     Idempotency = 5,
 }
 
+/// Prefixes for rows stored in the `msg_table` keyspace.
+#[repr(u8)]
+enum MsgPrefix {
+    Msg = 2,
+    HighWaterMark = 0,
+}
+
```
svix-jbrown, line 30: what's the advantage of having two prefixes instead of just one unified one per module? Worried we'll exceed 255 types?

svix-gabriel: Oh, I guess I should've called this out in the PR description.

So, the Msg rows are currently stored in a different keyspace than the other "metadata" rows, so it wasn't really necessary to begin with that the MsgRow prefix be distinct from the other "metadata" models.

It was much simpler (and I think intuitive) for the high water mark to go in the same keyspace as the msg rows, which caused me to split out the RowType enum to actually match the different keyspaces.

svix-jbrown (approved):

PR #1231, opened 2026-05-22
tests: remove startup timeout

svix-jbrown:
Sometimes we hit this. I don't know why; when I've profiled it there's nothing interesting going on, but it just takes us 6 seconds to start up. It doesn't seem to lead to the test running slower.

We already have a max test timeout in nextest; let's remove this extra timeout for now.

### DIFF
--- a/crates/test-utils/src/server.rs
+++ b/crates/test-utils/src/server.rs
@@ -142,11 +142,9 @@ impl TestServerBuilder {
         };
         let client = TestClient::new(base_uri, &token);
 
-        const MAX_INITIALIZATION_TIME: Duration = Duration::from_secs(5);
-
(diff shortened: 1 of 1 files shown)


svix-jplatte:
> it just takes us 6 seconds to start up

but this timeout should result in the test failing due to `expect` panicking?

svix-jbrown:
> > it just takes us 6 seconds to start up
> 
> but this timeout should result in the test failing due to `expect` panicking?

yep, which we see happen, but I had a branch before this where I removed the timeout and logged the startup time and none of them were, like, 19 seconds.

svix-jplatte (approved):

svix-jplatte:
I don't fully understand but w/e. If this causes things to actually just spin forever, we'll notice :)

PR #1234, opened 2026-05-26
Add garbage collection for stale leases

### LINKED ISSUE #1205: Msgs module: Garbage Collection for Leases
There's already some "just in time" garbage collection that happens, whereby overlapping leases are merged.

We should make sure when messages are deleted (due to retention configurations), leases that cover those message ranges are also deleted

svix-gabriel:
This closes https://github.com/svix/diom/issues/1205.

Essentially, for both Queue and Stream, we weren't cleaning up Leases that referenced message offsets that had been "hard deleted" by the background task. Over time, this could lead to stale leases taking up space unnecessarily.

We build off the extant background task that deletes expired messages, so when messages are deleted, we look for leases on that partition and delete them.

### DIFF
--- a/crates/msgs/src/storage.rs
+++ b/crates/msgs/src/storage.rs
@@ -470,6 +470,116 @@ pub(crate) fn delete_expired_partition(
     Ok(deleted)
 }
 
+/// Returns the offset of the earliest message in a partition, or `None` if the partition is empty.
+#[tracing::instrument(skip_all, level = "debug")]
+pub(crate) fn earliest_offset(
+    msg_table: &impl fjall_utils::ReadableKeyspace,
+    topic_id: TopicId,
+    partition: Partition,
+) -> Result<Option<Offset>> {
+    let prefix = MsgKey::prefix_partition(&topic_id, &partition);
+    match msg_table.prefix(prefix).next() {
+        Some(kv) => {
+            let key = kv.key()?;
+            let offset = MsgKey::extract_offset(&key)
+                .map_err(|e| diom_error::Error::internal(e.to_string()))?;
+            Ok(Some(offset))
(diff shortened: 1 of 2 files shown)


crates/msgs/src/storage.rs
```diff
@@ -470,6 +470,136 @@ pub(crate) fn delete_expired_partition(
...
+
+        if lease.offset < cutoff {
+            let consumer_group = StreamLeaseKey::extract_consumer_group(&key)
+                .map_err(|e| diom_error::Error::internal(e.to_string()))?;
+
+            let mut batch = db.batch();
+            batch.remove(metadata_tables, key);
+
+            // Also remove all QueueLeaseRows for this stale consumer group.
+            let ql_prefix = QueueLeaseKey::prefix_topic_id(&topic_id);
```
svix-james: I don't totally understand what's going on here.

svix-james: To be more specific, it looks like we're deleting _everything_ (ie for all partitions?) for the _queue_ consumer group when we clean up stream leases? Is that intended?

svix-gabriel: Yeah, it was wrong. We we're deleting way too many leases here. Should be fixed now.

svix-james: Can we add assertions that leases on unrelated partitions aren't affected?

svix-gabriel: sure (I updated the test)

svix-james (approved):

crates/msgs/src/storage.rs
svix-james: :sunglasses:

PR #1250, opened 2026-06-10
sec: ignore "unmaintained crate" rustsec warning about build-time dep

svix-jbrown:
[RUSTSEC-2026-0173](https://rustsec.org/advisories/RUSTSEC-2026-0173.html) is a report that [proc-macro-error2](https://crates.io/crates/proc-macro-error2) is not actively maintained. Ok.


svix-jplatte (approved):
Maybe we should get rid of `validator` entirely. We're only using it for config validation at this point.

PR #1257, opened 2026-06-21
Add feature that uses Svix AutoConfig, and appends to a topic

svix-gabriel:
The goal of this feature is to allow for Diom users to easily consume webhooks from Svix and append them to a Stream. This is done by integrating with [Autoconfig](https://docs.svix.com/webhooks-autoconfig)

The way this feature works is
* You create a "Poller" for a topic using `v1.msgs.svix-poller.create`. The Poller has everything needed to, well, use autoconfig to ingest webhooks sent svix.
* In a background worker, the API calls are made to svix to, well, Poll.
* Those messages are then appended to the topic internally in Diom.

The base implementation is fairly straightforward, although automated testing was tricky since we don't have an easy way to stub the Svix API. So I had to do a lot of manual testing.

In the background worker, the flow is basically:
1. Do a fetch from Svix (Autoconfig::receive)
2. Publish msgs to the topic
3. Do a commit on Svix (Autoconfig::commit)

The tricky/annoying part is that, if (2) takes too long, it risks causing Autoconfig::commit to fail, since the underlying lease on the Svix side would expire. So had to use a very long lease duration to minimize the risk. Ideally, (2) would also be done in a timeout, to completely nullify the risk, but I'm not sure how we'd safely revert (2) if step (3) failed. If someone has ideas, I'm all ears.


crates/msgs/src/operations/svix_poller/create.rs
```diff
@@ -0,0 +1,83 @@
...
+        let topic_row = TopicRow::fetch_or_create(
+            &state.metadata_tables,
+            &mut batch,
+            self.namespace_id,
+            &self.topic,
+            now,
+            self.topic_id_random_bytes,
+        )?;
+
+        let key = SvixPollerKey::build_key(&self.namespace_id, &topic_row.id, &self.poller_id);
```
svix-jbrown, line 54: what if there's already a poller for this ID and topic? is it safe to overwrite, or do we need to reset its offsets?

and are IDs intended to only be unique within a given topic?

src/v1/endpoints/msgs.rs
```diff
@@ -843,6 +844,96 @@ async fn topic_configure(
...
 
+// ---------------------------------------------------------------------------
+// svix_poller
+
+#[derive(Clone, Debug, Deserialize, JsonSchema)]
+pub struct SvixPollerCreateIn {
+    #[serde(default)]
+    pub namespace: Option<NamespaceName>,
+    pub topic: TopicName,
+    pub poller_id: String,
```
svix-jbrown, line 862: what is a poller id? is it meant to be an opaque value that the user can put anything into? might be worth adding a docstring

src/v1/endpoints/msgs.rs
```diff
@@ -843,6 +844,96 @@ async fn topic_configure(
...
+    pub token: String,
```
svix-jbrown, line 864: ditto, would be good to clarify that this is the token from *svix*

crates/msgs/src/operations/svix_poller/delete.rs
```diff
@@ -0,0 +1,59 @@
...
+            TopicKey::build_key(&self.namespace_id, &self.topic),
+        )? {
+            let key = SvixPollerKey::build_key(&self.namespace_id, &topic_row.id, &self.poller_id);
+            state.metadata_tables.remove_row::<SvixPollerRow, _>(key)?;
+        }
+
+        Ok(SvixPollerDeleteResponseData {
+            topic: self.topic,
+            poller_id: self.poller_id,
+        })
```
svix-jbrown, line 36: should this return whether or not something was removed?

src/v1/endpoints/msgs.rs
```diff
@@ -913,4 +1004,14 @@ pub fn router() -> ApiRouter<AppState> {
...
+        .api_route_with(
+            svix_poller_create_path,
+            post_with(svix_poller_create, svix_poller_create_operation),
+            &tag,
+        )
+        .api_route_with(
+            svix_poller_delete_path,
+            post_with(svix_poller_delete, svix_poller_delete_operation),
+            &tag,
+        )
```
svix-jbrown, line 1111: probably needs a poller_list too before it's shippable; right now, there's no way to see what exists or what its state is...?

crates/msgs/src/svix_poller.rs
```diff
@@ -0,0 +1,437 @@
...
+    pub fn new(state: State, poll_interval: Duration, handle: F) -> Self {
+        Self {
+            state,
+            poll_interval,
+            handle,
+            _client: PhantomData,
+        }
+    }
+
+    async fn poll_cycle(&self) -> BackgroundResult<()> {
```
svix-jbrown, line 118: my reading of this is that the algorithm is to poll every configured poller once, serially, from this function?

that seems like it could be slow, particularly if the svix endpoint is on the other side of the internet and has hundreds of milliseconds of latency. Would it be better to move the HTTP calls to run concurrently?

crates/msgs/src/svix_poller.rs
```diff
@@ -0,0 +1,437 @@
...
+                            TopicIn::TopicName(topic_name),
+                            msgs,
+                            None,
+                        )
+                        .map_err(diom_operations::BackgroundError::Other)?;
+
+                        self.handle.write_request(op).await?;
+
+                        if let Err(e) = client.commit(&consumer_id, offset).await {
+                            tracing::warn!(
```
svix-jbrown: do we need to reset any offsets somewhere if this happens?

crates/msgs/src/operations/svix_poller/create.rs
svix-gabriel: Yeah, the intent is just that poller_id is the unique id the user applies. So if the poller already exists, we'd just overwrite/upsert it.

crates/msgs/src/operations/svix_poller/delete.rs
svix-gabriel: I didn't think of it, but yeah, makes sense to follow the same `success: true/false` pattern we have elsewhere. Updated.

crates/msgs/src/svix_poller.rs
svix-gabriel: On the svix side we don't need to reset anything. If `client.commit` does fail, eventually the "lock" on the messages received from `poll_from_config` will expire, and so the poll cycle will just transparently retry. Since we can't practically rollback the `PublishOperation`, we do risk publishing the same messages twice in this situation, but there's not a lot we can do about that.

crates/msgs/src/svix_poller.rs
svix-gabriel: Yeah, that's how it was setup.

I was aiming for simplicity initially, but it wasn't too hard to update it to poll for different topics concurrently, so I've updated it appropriately.

crates/diom-core/src/tokio_nursery.rs
```diff
@@ -0,0 +1,186 @@
...
+    tasks: JoinSet<()>,
+}
+
+impl TaskNursery {
+    /// Log at `warn` when a spawn waits longer than this for a permit.
+    const WARN_SPAWN_WAIT: Duration = Duration::from_millis(10);
+
+    /// Create a nursery that runs at most `limit` tasks concurrently.
+    pub fn new(limit: usize) -> Self {
+        debug_assert!(limit > 0, "TaskNursery limit must be greater than zero");
```
svix-jbrown: since the config is already a NonZeroUsize, why not just take a NonZeroUsize here and make the debug assert an impossible condition?

crates/msgs/src/svix_poller.rs
```diff
@@ -0,0 +1,513 @@
...
+/// Does NOT commit — the caller must commit after successfully persisting the
+/// messages to avoid data loss on publish failure.
+#[instrument(skip_all, fields(namespace=?key.namespace_id, topic = ?key.topic_id, poller_id = %key.poller_id, count, success = false))]
+pub(crate) async fn poll_from_config(
+    client: &impl SvixAutoConfigClient,
+    key: &SvixPollerKey,
+    max_messages: usize,
+    lease_duration: Duration,
+) -> Result<(Vec<MsgIn>, Option<i32>), SvixClientError> {
+    tracing::debug!("Polling from AutoConfig");
```
svix-jbrown: might want to drop this to a trace
```suggestion
    tracing::trace!("Polling from AutoConfig");
```

crates/diom-core/src/tokio_nursery.rs
```diff
@@ -0,0 +1,186 @@
...
+    pub async fn spawn<F>(&mut self, future: F)
+    where
+        F: Future<Output = ()> + Send + 'static,
+    {
+        let permit = {
+            let start_wait = Instant::now();
+
+            let permit = self
+                .semaphore
+                .clone()
```
svix-jbrown: optional nit: minor preference for `Arc::clone(&self.semaphore)` over `self.semaphore.clone()`; I thought we'd actually enabled the lint for this but I guess that wasn't in diom...

crates/diom-core/src/tokio_nursery.rs
```diff
@@ -0,0 +1,186 @@
...
+            if wait > Self::WARN_SPAWN_WAIT {
+                tracing::warn!(?wait, "slow task spawn, waiting for nursery permit");
+            }
+
+            permit
+        };
+
+        let task = async move {
+            let _permit = permit;
+            future.await;
```
svix-jbrown:
```suggestion
            future.await;
            drop(permit)
```

maybe clearer about expected lifetime? feel free to disregard if you disagree

crates/diom-core/src/tokio_nursery.rs
```diff
@@ -0,0 +1,186 @@
...
+
+        self.tasks.spawn(task.in_current_span());
+    }
+
+    /// Wait for all outstanding tasks to finish.
+    pub async fn join_all(mut self) {
+        // use join_next() over join_all() to avoid propagating panics
+        while let Some(result) = self.tasks.join_next().await {
+            if let Err(e) = result {
+                tracing::error!(error = %e, "nursery task did not complete cleanly");
```
svix-jbrown: the only possible error here is a panic, right?

svix-jbrown (commented):
couple of little notes and questions, but nothing major

crates/diom-core/src/tokio_nursery.rs
svix-gabriel: As far as I know, yeah

svix-jbrown (approved):

PR #1263, opened 2026-06-26
deps: bump openraft to alpha24

svix-jbrown:
no major changes; passes the test suite and performs equivalently in local testing

PR #1265, opened 2026-06-26
deps: bump openraft to alpha25

svix-jbrown:
obviously right after I bump openraft to alpha24 in #1263, they release an alpha25 with more bugfixes


svix-jplatte (approved):
Number go up.

PR #1272, opened 2026-07-01
rust: bump to 1.96.1

svix-jbrown:


.github/workflows/benchmark.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v7
       - uses: dtolnay/rust-toolchain@fa04a1451ff1842e2626ccb99004d0195b455a88 # master
         with:
-          toolchain: '1.96'
+          toolchain: '1.96.1'
```
svix-jplatte, line 17: This is the same thing. I avoided bumping elsewhere because I don't want the next bump to include the patch version for no reason, but maybe it's fine 🤷

svix-jplatte: Same for other workflows.

svix-jplatte (approved):

.github/workflows/benchmark.yml
svix-jbrown: I just like precision

PR #1278, opened 2026-07-08
Improve Diom Operator Status reporting

svix-james:
In line with API recommendations outlined in K8s architecture
documents, this changeset adds `Condition`s to the Diom cluster's 
reported status. These will include a `Ready` condition, which should 
always be present, along with possible failure conditions of 
`Reconciling` or `Stalled`. The practical effect of these changes
are that monitoring successful/failed deployments should be much
easier now since the status conditions can be directly watched
during the deployment process.

The `observed_generation` of the cluster resource has also been added 
and is incremented whenever the cluster resource is updated.

`Phase` has been removed entirely from status reporting, as it's now
redundant. (Also note that the K8s API Conventions document states that 
"the pattern of using `phase` is deprecated.")

References:
* https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
* https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/README.md

### DIFF
--- a/infra/operator/src/crd.rs
+++ b/infra/operator/src/crd.rs
@@ -6,7 +6,7 @@ use k8s_openapi::{
         Affinity, EmptyDirVolumeSource, EnvVar, EnvVarSource, HTTPGetAction, Probe,
         ResourceRequirements, SecretKeySelector, Toleration, TopologySpreadConstraint,
     },
-    apimachinery::pkg::api::resource::Quantity,
+    apimachinery::pkg::{api::resource::Quantity, apis::meta::v1::Condition},
 };
 use kube::CustomResource;
 use schemars::JsonSchema;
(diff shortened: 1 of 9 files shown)


infra/operator/src/reconciler.rs
```diff
@@ -40,94 +197,422 @@ impl Reconciler {
...
+    Action::requeue(ctx.requeue_interval)
+}
+
+fn find_condition<'a>(conditions: &'a [Condition], type_: &str) -> Option<&'a Condition> {
+    conditions.iter().find(|c| c.type_ == type_)
+}
+
+fn condition_unchanged(previous: Option<&Condition>, new: &Condition) -> bool {
+    previous.is_some_and(|c| {
+        c.status == new.status && c.reason == new.reason && c.message == new.message
```
svix-gabriel, line 351: Is there a reason we don't just compare `new == c` directly? Are there other fields on `Condition` we're deliberately ignoring?

infra/operator/src/crd.rs
```diff
@@ -381,18 +381,13 @@ impl ProbeSpec {
 #[serde(rename_all = "camelCase")]
 pub struct DiomClusterStatus {
     #[serde(default)]
-    pub phase: Phase,
+    pub ready_replicas: i32,
```
svix-gabriel, line 384: Do we need to worry about "breaking" API changes in this layer? (I guess I can't really tell if this is actually a breaking change?)

svix-james: Yeah, this is definitely "breaking" as the API goes. I will note that in the changelog. It will be a major version increment for sure

infra/operator/src/reconciler.rs
svix-james: I think this is safer. There are other fields there that should match, but it doesn't particularly matter if they do, and since I'm not concerned about them, I think we should limit it for now to the relevant ones.

svix-gabriel (approved):

PR #1283, opened 2026-07-11
server: allow hitting v1.admin.metrics.get before bootstrap finishes

svix-jbrown:
the point of the debugging metrics endpoint is to be able to introspect the running server...

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -59,7 +59,8 @@ pub(crate) async fn fail_until_bootstrapped(
     next: middleware::Next,
 ) -> axum::response::Response {
     let is_admin_route = path.as_str().starts_with("/api/v1.admin.cluster.")
-        || path.as_str() == "/api/v1.health.ping";
+        || path.as_str() == "/api/v1.health.ping"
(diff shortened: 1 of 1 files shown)


src/utils.rs
```diff
@@ -59,7 +59,8 @@ pub(crate) async fn fail_until_bootstrapped(
     next: middleware::Next,
 ) -> axum::response::Response {
     let is_admin_route = path.as_str().starts_with("/api/v1.admin.cluster.")
-        || path.as_str() == "/api/v1.health.ping";
+        || path.as_str() == "/api/v1.health.ping"
+        || path.as_str() == "/api/v1.admin.metrics.get";
```
svix-jplatte, line 63: Should this also be a starts_with, or are we not expecting to ever add more routes under v1.admin.metrics?

svix-jplatte (approved):

src/utils.rs
svix-jbrown: I filed it under YAGNI for now

PR #1292, opened 2026-07-16
embed metadata in cli and server binaries; add SBOM/SPDX inside docker images

svix-jbrown:
part of svix/monorepo-private#13574


svix-jplatte (commented):
Needs Cargo.toml reformatting to make cargo-sort happy.

svix-jplatte (approved):

PR #1299, opened 2026-07-17
Feature: HTTP sink to forward requests from a topic automatically

svix-gabriel:
This closes https://github.com/svix/monorepo-private/issues/13591 (from https://github.com/svix/monorepo-private/issues/13223).

The implementation and feature set is based on the comments in that ticket, so I recommend reading the linked ticket and comments first to understand the motivation and end feature more.


svix-jbrown (commented):
Is there some documentation somewhere on why this would exist?

Architecturally, I would never build a system where the resource-constrained central resource (the database) had to do a bunch of extra work, nor one where the database had permissions to make outbound HTTP requests. All of this seems like it can scale horizontally and would make more sense in some separately-deployed service. Before I start looking at a 6000 line diff... what's the use-case?

crates/msgs/src/entities.rs
```diff
@@ -464,6 +465,127 @@ pub enum SeekPosition {
...
+        SinkConfigTaggedRef::json_schema(generator)
+    }
+}
+
+/// Configuration for an HTTP sink. The `url`, `headers`, and `body` are templates rendered
+/// per-message (see [`diom_core::template_str`]).
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, JsonSchema, PersistableValue)]
+pub struct HttpSinkConfig {
+    /// Destination URL.
+    pub url: Template,
```
svix-james, line 568: SSRF?

crates/msgs/src/operations/sink/delete.rs
```diff
@@ -0,0 +1,75 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    entities::TopicName,
+    operations::{MsgsRaftState, MsgsRequest, SinkDeleteResponse},
+    storage::{SinkKey, SinkRow, TopicKey, TopicRow},
+};
+
+#[derive(Debug, Clone, Serialize, Deserialize, PersistableValue)]
+pub struct SinkDeleteOperation {
```
svix-james, line 14: If we support this, do we also need to clean up stuff in the metadata tables?

svix-james:
> All of this seems like it can scale horizontally and would make more sense in some separately-deployed service. 

I do agree here, though. It seems like we could have "plugins" that do this and run alongside the main service without them directly being a part of it.

tasn:
Github sucks, I wish I could reply in thread there. Anyhow, @svix-jbrown:

> Is there some documentation somewhere on why this would exist?
> 
> Architecturally, I would never build a system where the resource-constrained central resource (the database) had to do a bunch of extra work, nor one where the database had permissions to make outbound HTTP requests. All of this seems like it can scale horizontally and would make more sense in some separately-deployed service. Before I start looking at a 6000 line diff... what's the use-case?

By being an omni-tool we do more than what a traditional tool would do (btw, Diom is not a db...) with the goal of making things easier to setup and use. It would definitely make sense in a separately deployed service once you scale out, but the small deployment version would benefit from this a lot (less stuff to deploy, monitor, etc).

Some thoughts:
1. That extra connector service can be a special Diom node even when scaled out. So you'd have a horizontally scaled Diom cluster, and then this one node that is just the worker node for the connectors (that is otherwise isolated).
2. It's a very easy and powerful way to let Diom interact with other existing systems (messaging or otherwise) and is essentially a different way of consuming events from it. This is especially useful for smaller/simpler deployments, but not only.
3. Realistically most people aren't going to run this with the CPU pegged at 100% (it's for simpler deployments), so the CPU constrains is less of an issue. Anyway though, this should be fairly light weight.
4. As an anecdote: Together AI really liked the built-in Kafka connector I built for them (that was just part of Diom) as part of the demo/testing.
5. The next step/feature Gabriel will be adding is Diom -> Svix automatic publishing. So they can just send to their local Diom and it will automatically publish to Svix, making it easy for us to have a beachhead into their internal event-driven system.

I'm probably forgetting some points, but this is not "on by default" in the sense, that if you don't use it, nothing happens. So it's just giving people the ability to do it if they really want to.

crates/msgs/src/operations/sink/delete.rs
svix-gabriel: I didn't originally think about it, but it's a good point.

Updated so we cleanup some of the cursors on deletion.

crates/msgs/src/entities.rs
svix-gabriel: Uhhh @tasn do you have thoughts/concerns here?

My assumption behind this feature was that whoever's configuring the topic/sink is the same person writing msgs to the topic, but I guess that's not necessarily true.

The original idea behind making the URL templatable is that it lets you plug in things like path parameters from the msg.

svix-gabriel (commented):
Tom and others asked for this, so here's a write up behind the feature as well https://github.com/svix/monorepo-private/issues/13591#issuecomment-5064167436 (worth reading the whole comment chain)

crates/msgs/src/entities.rs
tasn: I don't actually know. I feel like we can start without SSRF, because of the reasons you said, but there's also the concern of some DNS rebinding that may be a problem even if you only use a trusted external destinations.

Not sure.

tests/it/msgs/sink.rs
```diff
@@ -0,0 +1,594 @@
...
+
+    assert_eq!(response["topic"], "orders");
+    assert_eq!(response["consumer_group"], "sink_del");
+    assert_eq!(response["success"], true);
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_sink_delete_is_idempotent() -> TestResult {
```
svix-james: I don't really understand what this test is supposed to do, at least in terms of idempotency

svix-gabriel: Yeah it's stupid. The original intent was just to do 2 delete calls and make sure the 2nd one doesn't error (even if the sink doesn't exist)... and just forgot the 2nd delete call. It's not a super useful thing to test anyway so I just axed it.

tests/it/msgs/sink.rs
```diff
@@ -0,0 +1,594 @@
...
+    run_with_retries(async || {
+        let n = mock.received_requests().await.unwrap_or_default().len();
+        anyhow::ensure!(n >= 1, "expected first delivery, got {n}");
+        Ok(())
+    })
+    .await?;
+    tokio::time::sleep(Duration::from_secs(2)).await;
+    let before = mock.received_requests().await.unwrap_or_default().len();
+
+    // Delete the sink (which should clear its cursor) and re-create it with the same consumer group.
```
svix-james, line 545: Just curious -- what happens if we delete it while the requests are still in progress.

svix-gabriel: 🤦 I completely missed this comment

I think the worst case is
1. There's some residual dispatch
2. Possibly some orphaned lease rows that are created during that residual dispatch

We already have garbage collection for stale / expired msg data, I'll double check if it would clear this stuff up.

The way to avoid this would be to just check if the sink exists every "tick" of the dispatch loop. I guess we can do that, IDK if the extra disk reads are worth it though.

svix-james (approved):

PR #1300, opened 2026-07-17
CI: Add tags on Helm/Operator release

svix-james:
I guess this isn't strictly necessary, but I think it will make things easier to track, so let's add tags for now via the CI job.


svix-jbrown (approved):
love me some tags

Issue #494, opened 2026-03-17
Add tokio metrics


svix-jplatte:
We can probably copy-paste what we have in webhooks-private.

PR #1305, opened 2026-07-22
Add Tokio runtime metrics

KaustubhOG:
Adds Tokio runtime metrics per #494, following the webhooks-private approach.

New **TokioMetrics** in **core::metrics** reports the runtime's health as **diom.tokio.*** gauges (alive tasks, queue depths, blocking threads, and the poll-time histogram), collected by a background task in the same shape as the
existing **DbMetrics** loop.

The poll-time histogram is only populated when the runtime is built with **enable_metrics_poll_time_histogram**, which **#[tokio::main]** cannot set, so **main** now builds the runtime explicitly. Reporting is guarded by
**poll_time_histogram_enabled**, so runtimes without it (the test harness) are unaffected.

## Testing

fmt, clippy, and the integration suite pass. Booted a single node locally (dummy admin token) and confirmed the diom.tokio.* series are emitted via the admin metrics endpoint:

    diom.tokio.alive_tasks    = 28
    diom.tokio.poll_histogram = 161   # worker_id=11
    ...                               # per-worker buckets across workers 0-15

Screenshot from 2026-07-23 04-21-20
Screenshot from 2026-07-23 04-20-56

### DIFF
--- a/src/core/metrics.rs
+++ b/src/core/metrics.rs
@@ -1,11 +1,12 @@
 use std::time::Duration;
 
+use diom_core::shutdown::shutting_down_token;
 use http::StatusCode;
 use opentelemetry::{
     KeyValue, Value,
     metrics::{Counter, Gauge, Histogram, Meter},
 };
-use tokio::runtime::RuntimeMetrics;
+use tokio::runtime::{Handle, RuntimeMetrics};
 
 use super::cluster::NodeId;
 
@@ -692,10 +693,28 @@ impl ClusterNetworkMetrics {
     }
 }
 
+trait GaugeExt {
+    fn record_if_nonzero(&self, value: u64, attributes: &[KeyValue]);
+}
+
+impl GaugeExt for Gauge<u64> {
+    fn record_if_nonzero(&self, value: u64, attributes: &[KeyValue]) {
+        if value > 0 {
+            self.record(value, attributes);
+        }
+    }
+}
+
 /// Number of poll-time histogram buckets reported per worker, from the slowest end.
 const NUM_REPORTED_POLL_HISTOGRAM_BUCKETS: usize = 10;
 
-const QUEUE_DEPTH_REPORT_THRESHOLD: u64 = 2;
+/// Runtime saturation values at or below this are not reported.
+///
+/// Queue depths and the number of active blocking threads sit low almost all of
+/// the time, and only become interesting when they spike or stay elevated.
+/// Skipping the uninteresting values keeps the volume of reported metrics, and
+/// the cost of storing them, down.
+const SATURATION_REPORT_THRESHOLD: u64 = 2;
 
 /// Previous poll-time histogram bucket values, kept so each collection can report
 /// the per-interval difference of the runtime's monotonic bucket counters.
@@ -762,13 +781,11 @@ impl TokioMetrics {
     pub fn record(&self, metrics: &RuntimeMetrics, ctx: &mut RuntimeMetricsContext) {
         let node = std::slice::from_ref(&self.node_id_kv);
 
-        let alive_tasks = metrics.num_alive_tasks() as u64;
-        if alive_tasks != 0 {
-            self.alive_tasks.record(alive_tasks, node);
-        }
+        self.alive_tasks
+            .record_if_nonzero(metrics.num_alive_tasks() as u64, node);
 
         let global_queue_depth = metrics.global_queue_depth() as u64;
-        if global_queue_depth > QUEUE_DEPTH_REPORT_THRESHOLD {
+        if global_queue_depth > SATURATION_REPORT_THRESHOLD {
             self.global_queue_depth.record(global_queue_depth, node);
         }
 
@@ -778,13 +795,13 @@ impl TokioMetrics {
             .num_blocking_threads()
             .saturating_sub(metrics.num_idle_blocking_threads())
             as u64;
-        if active_blocking_threads > QUEUE_DEPTH_REPORT_THRESHOLD {
+        if active_blocking_threads > SATURATION_REPORT_THRESHOLD {
             self.active_blocking_threads
                 .record(active_blocking_threads, node);
         }
 
         let blocking_queue_depth = metrics.blocking_queue_depth() as u64;
-        if blocking_queue_depth > QUEUE_DEPTH_REPORT_THRESHOLD {
+        if blocking_queue_depth > SATURATION_REPORT_THRESHOLD {
             self.blocking_queue_depth.record(blocking_queue_depth, node);
         }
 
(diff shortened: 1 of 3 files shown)


svix-jplatte:
CI failure may just be a case of a workflow that doesn't work when run from a fork. I'll try to do a proper review soon.

src/core/metrics.rs
```diff
@@ -690,3 +691,159 @@ impl ClusterNetworkMetrics {
...
+                .build(),
+            node_id_kv: node_id.into(),
+        }
+    }
+
+    pub fn record(&self, metrics: &RuntimeMetrics, ctx: &mut RuntimeMetricsContext) {
+        let node = std::slice::from_ref(&self.node_id_kv);
+
+        let alive_tasks = metrics.num_alive_tasks() as u64;
+        if alive_tasks != 0 {
```
svix-jplatte: I see now that the snippet I shared included uses of a helper that was not part of the snippet. I think it would make sense to use it here as well.

This is the helper:

```rust
trait GaugeExt {
    fn record_if_nonzero(&self, value: u64, attributes: &[KeyValue]);
}

impl GaugeExt for Gauge<u64> {
    fn record_if_nonzero(&self, value: u64, attributes: &[KeyValue]) {
        if value > 0 {
            self.record(value, attributes);
        }
    }
}
```

src/core/metrics.rs
```diff
@@ -690,3 +691,159 @@ impl ClusterNetworkMetrics {
...
+            self.global_queue_depth.record(global_queue_depth, node);
+        }
+
+        // saturating: the two counters are independent atomics, so a reader can
+        // briefly observe more idle threads than total during pool teardown.
+        let active_blocking_threads = metrics
+            .num_blocking_threads()
+            .saturating_sub(metrics.num_idle_blocking_threads())
+            as u64;
+        if active_blocking_threads > QUEUE_DEPTH_REPORT_THRESHOLD {
```
svix-jplatte: Number of active blocking threads is not related to any queue so the named constant doesn't really make sense here. Oh and for context, this is also a cost-efficiency thing we did: most of the time queue depth + number of blocking threads is low, and everything is fine. It's only when there's a spike or constant elevation of one of these when these metrics become interesting. Might be worth putting a small comment somewhere given you've already documented a couple things that weren't in the original snippet :)

src/core/metrics.rs
```diff
@@ -690,3 +691,159 @@ impl ClusterNetworkMetrics {
...
+            total_buckets.saturating_sub(NUM_REPORTED_POLL_HISTOGRAM_BUCKETS);
+
+        for worker in 0..metrics.num_workers() {
+            let prev_values = &mut ctx.previous_poll_histogram_bucket_values[worker];
+
+            for (bucket, prev) in prev_values.iter_mut().enumerate() {
+                let underlying_bucket = first_reported_bucket + bucket;
+                if underlying_bucket >= total_buckets {
+                    break;
+                }
```
svix-jplatte: This shouldn't be needed, I'm pretty sure. Why did you add it?

src/core/metrics.rs
```diff
@@ -690,3 +691,159 @@ impl ClusterNetworkMetrics {
...
+        }
+
+        let total_buckets = metrics.poll_time_histogram_num_buckets();
+        let first_reported_bucket =
+            total_buckets.saturating_sub(NUM_REPORTED_POLL_HISTOGRAM_BUCKETS);
+
+        for worker in 0..metrics.num_workers() {
+            let prev_values = &mut ctx.previous_poll_histogram_bucket_values[worker];
+
+            for (bucket, prev) in prev_values.iter_mut().enumerate() {
```
svix-jplatte, line 838: Thanks for making me realize the clippy lint was actually appropriate here 😄 
I think there was a better reason to silence it in an older version of the code.

src/serve.rs
```diff
@@ -215,6 +219,23 @@ pub async fn run_with_listeners(
     tracing::info!("shutdown complete");
 }
 
+/// Spawn a background task that periodically reports Tokio runtime metrics.
+fn spawn_tokio_metrics(metrics: TokioMetrics) {
```
svix-jplatte: I think it would be little nicer if this was part of `metrics.rs` and took care of instantiating `TokioMetrics`, i.e.


```suggestion
fn spawn_tokio_metrics(meter: &Meter, node_id: NodeId) {
    let metrics = TokioMetrics::new(meter, node_id);
```

svix-jplatte (commented):
Thanks, looks good overall! Got a couple small notes.

svix-jplatte:
Hey @KaustubhOG, gentle ping. Do you intend to come back to this and address the review feedback?

KaustubhOG:
> Hey @KaustubhOG, gentle ping. Do you intend to come back to this and address the review feedback?

Hey, got busy with work. I’ll review the feedback soon. Thanks for the ping, and also for merging my OmniQueue PRs!

svix-jplatte:
Great. No rush, just wanted to make sure it didn't fall off your radar.

src/core/metrics.rs
KaustubhOG: Added GaugeExt with record_if_nonzero and used it for alive_tasks and the poll histogram. The other places check against SATURATION_REPORT_THRESHOLD, not zero, so I left those as they are.

src/core/metrics.rs
KaustubhOG: Renamed it to SATURATION_REPORT_THRESHOLD, since it is used for queue depths and for the blocking thread count. I also added a comment on the constant with the reason you gave: these numbers stay low most of the time and only matter when they spike or stay high, so skipping the low ones keeps the metric volume and cost down.

src/core/metrics.rs
KaustubhOG: I added it just to be safe, but you are right, it is not needed. My worry was a histogram with fewer buckets than NUM_REPORTED_POLL_HISTOGRAM_BUCKETS. In that case saturating_sub gives 0, and the loop would go past the last bucket and panic in poll_time_histogram_bucket_range.
But that cannot happen here. The runtime uses the default poll time histogram, which has exactly 10 buckets, the same as NUM_REPORTED_POLL_HISTOGRAM_BUCKETS. So the loop only reads buckets 0 to 9 and stays in range. I removed the check.
This only holds while we use the default config. If someone later sets a custom metrics_poll_time_histogram_configuration with fewer buckets, this loop will need another look.

src/core/metrics.rs
KaustubhOG: Got it, no change needed here. I kept the iterator version.

src/serve.rs
KaustubhOG: Moved spawn_tokio_metrics into metrics.rs with that signature. It now creates the TokioMetrics itself, and serve.rs just calls spawn_tokio_metrics(&app_state.meter, node_id)

svix-jplatte (approved):
Sorry for the delay, GitHub is showing that the most recent push to this branch was before my comment asking if you were getting back to this which seems wrong? Either way I didn't realize the review feedback had been incorporated.

svix-jplatte:
Ah dang it, I can't merge because of that CI issue..

KaustubhOG:
Thank you @svix-jplatte  for merging #1305 and for fixing the CI workflow in #1340! I’ll rebase #1302 onto the latest main and resolve the remaining conflict so it’s ready for review.

PR #1313, opened 2026-08-05
Implement FIFO semantics for consuming msgs from topics

svix-gabriel:
This closes https://github.com/svix/monorepo-private/issues/13765

IDK if "FIFO" is really the right name, so feel free to bikeshed about it. But Tom's been asking for it for a while.

The idea behind "FIFO" semantics is that strict ordering is preserved for msgs with the same `key`, but beyond that there are no restrictions on ordering (though general first-in-first-out is still done as a best effort).

`stream` semantics does get at this behavior, except `stream` semantics enforce strict ordering on an entire partition. So different keys that map to the same partition will be strictly ordered, which FIFO does NOT guarantee.

The examples are probably best understood by reading the tests first. Beyond that, I mostly relied on extant `queue` functions / logic.


crates/msgs/src/storage.rs
```diff
@@ -279,6 +279,7 @@ pub(crate) struct MsgRow {
     pub headers: HashMap<String, String>,
     pub timestamp: UnixTimestampMs,
     pub scheduled_at: Option<UnixTimestampMs>,
+    pub key: Option<String>,
 }
```
svix-jbrown, line 283: remember that we're using `postcard`, which is not a self-describing format, so adding new optional fields will probably break any existing databases.

We talked about (but never implemented) something to infer a version tag on fields and automatically turn them into enums. You're probably going to need to do that work to make this change.

svix-jbrown (commented):
Before we try to review this, can you write down the algorithm you are expecting this to implement, and what you think its performance characteristics are? 

To me, it looks like this is going to be something like O(n^2 log n) because it's going to be traversing QueueLeaseRow multiple times for every possible message, and that this algorithm is somewhat unlikely to ever finish for dense data (that is to say, to fetch 100 records, it will attempt to scan past up to 1600 records, but if the insert pattern looks like `1601 * a` followed by one b, the b is unreachable because there's no skip index. Maybe I'm misunderstanding, though.

svix-gabriel:

PR #1314, opened 2026-08-05
[CI] Manually Install crun 1.28 before codegen to fix podman OCI v1.2.1 error

svix-gabriel:
In response to this lovely failure: https://github.com/svix/diom/actions/runs/31027382776/job/92380933891?pr=1313

The check-codegen job's containerized formatters (goimports, ruff) were
failing with "crun: unknown version specified".

Based on https://github.com/podman-container-tools/podman/issues/27272, it seems like a version mismatch with what Ubuntu 24.04 ships and what the OCI runtime spec requires, or something like that.

Installing a recent static crun before running cargo codegen seems to fix this.


svix-jbrown:
What? Is this running podman inside docker? Why do we have multiple nested container infrastructures?

I think we're just missing a `setup-docker` step if we expect to be building containers here.

svix-jbrown:
I think #1315 will be less fragile

Issue #1316, opened 2026-08-06
port to new APIs from openraft 0.10.0-alpha33


svix-jbrown:
openraft 0.10.0-alpha33 is out with big changes to the snapshot APIs

PR #1317, opened 2026-08-06
cluster: port to openraft 0.10.0-alpha.34

svix-jbrown:
This is a first pass at porting to openraft 0.10.0-alpha.33 and the newly-published `openraft-legacy` crate (so removing the vendorized version of it).

This entirely removes the concept of snapshot IDs. There's now a "snapshot signature" on the `SnapshotMeta` (transmitted in a wire-protocol compatible way by openraft), which we use instead.

First pass is to make sure that it passes tests; still need to test cross-version compatibility. I think it should work across versions on the network, but downgrading an individual host from this version to the previous version might **not** work because `SnapshotMeta` no longer has a snapshot_id, and we will probably crash when trying to extract the idx.

This uses the branch from https://github.com/databendlabs/openraft/pull/2067

Part of #1316


src/core/cluster/state_machine.rs
```diff
@@ -934,10 +912,11 @@ impl StoreSnapshotHandle {
 
         tracing::debug!(
             ?path,
-            last_log_id = ?meta.last_log_id(),
+            last_log_id = meta.last_log_id().map(tracing::field::display),
```
svix-gabriel, line 914: not that it really matters that much, but why the change?

src/v1/endpoints/cluster_admin.rs
```diff
@@ -41,6 +41,21 @@ macro_rules! request_input {
     };
 }
 
+pub fn snapshot_id_from_signature(sig: SnapshotSignature) -> String {
+    let mut buf = String::new();
+    if let Some(last_membership) = sig.last_membership_log_id {
+        write!(buf, "{}-", last_membership.leader_id.term,).unwrap();
```
svix-gabriel:
```suggestion
        write!(buf, "{}-", last_membership.leader_id.term,).expect("writing into a String buffer should be infallible");
```

I thought we had a lint against unwraps, but I guess not?

src/core/cluster/state_machine.rs
svix-jbrown: the `Debug` impl for log ids got a lot bigger in the new version, but the Display version is still pretty compact

svix-jbrown:
gentle nudge on this @svix-gabriel

src/core/cluster/mod.rs
```diff
@@ -29,5 +28,6 @@ pub type NodeId = NodeIdOf<TypeConfig>;
 pub(crate) type Node = NodeOf<TypeConfig>;
 pub(crate) type RaftError<C> = openraft::error::RaftError<TypeConfig, C>;
```
svix-jplatte: Not part of this diff and can be handed in a separate PR but I looked at the upstream thing and `C` is used for the type param we're fixing here. Also should maybe copy their default for the remaining type param?

```suggestion
pub(crate) type RaftError<E = Infallible> = openraft::error::RaftError<TypeConfig, E>;
```

src/core/cluster/network.rs
```diff
@@ -323,25 +315,22 @@ impl RaftNetworkV2<TypeConfig> for NetworkClient {
         &mut self,
         rpc: openraft::raft::VoteRequest<TypeConfig>,
         option: RPCOption,
-    ) -> Result<openraft::raft::VoteResponse<TypeConfig>, RPCError> {
+    ) -> Result<openraft::raft::VoteResponse<TypeConfig>, RPCError<RaftError<TypeConfig>>> {
```
svix-jplatte: Should use our `RaftError` type alias here (can be in a follow-up PR). Maybe we should also have an `RpcError` type alias?

src/core/cluster/state_machine.rs
```diff
@@ -532,28 +508,31 @@ impl Store {
 
     async fn install_snapshot_(
         &mut self,
+
         meta: &SnapshotMeta,
         mut snapshot: StoredSnapshot,
     ) -> anyhow::Result<()> {
```
svix-jplatte, line 513: Extra newline probably not intended?

src/v1/endpoints/cluster_admin.rs
```diff
@@ -41,6 +41,23 @@ macro_rules! request_input {
     };
 }
 
+pub fn snapshot_id_from_signature(sig: SnapshotSignature) -> String {
+    let mut buf = String::new();
+    if let Some(last_membership) = sig.last_membership_log_id {
+        write!(buf, "{}-", last_membership.leader_id.term,)
```
svix-jplatte:
```suggestion
        write!(buf, "{}-", last_membership.leader_id.term)
```

svix-jplatte (dismissed):
Got some nitpicks, nothing blocking.

src/core/cluster/mod.rs
svix-jbrown: sure

src/core/cluster/network.rs
svix-jbrown: sure

svix-jplatte (approved):

PR #1324, opened 2026-08-10
Implement Kafka + Svix Sink

svix-gabriel:
This implements Kafka + Svix as sinks, building off of https://github.com/svix/diom/pull/1299

The "Svix" sink is super straightforward. It's essentially just a convenient wrapper around the HTTP sink.

The Kafka sink is more involved. The major annoyance / hiccup here was adding enough dependencies for rdkafka to build in the docker image and tests.

The other annoyance was new Sink params causing an enum in the SDK to get ballooned in size, triggering a fun clippy lint about chungus enums. I opted to just allowlist this warning, since it seemed pretty unlikely to actually be problematic in the context of the cli build.


tests/it/msgs/sink.rs
```diff
@@ -1,12 +1,22 @@
+#![allow(clippy::disallowed_types)]
```
svix-james, line 1: Do we want this on the whole file?

svix-gabriel: I think it's what we've typically done in test files. It's just because we disallow `serde_json::Value` everywhere but tests.

infra/operator/Dockerfile
```diff
@@ -23,8 +23,18 @@ RUN --mount=target=/var/lib/apt/lists,type=cache,sharing=locked --mount=target=/
     set -euxo pipefail
     export DEBIAN_FRONTEND=noninteractive
     apt-get update -q
+    # build-essential (make plus a C/C++ toolchain), cmake, pkg-config and the OpenSSL, zlib and curl
```
svix-james: Why are we changing the operator?

tests/it/msgs/sink.rs
svix-james: Derp, yeah makes sense.

crates/msgs/src/sink.rs
```diff
@@ -477,6 +584,80 @@ async fn deliver(
...
+            Some(template) => template.apply(&vars).into_bytes(),
+            None => msg.value.to_vec(),
+        };
+
+        let record = FutureRecord::to(self.topic)
+            .payload(&value)
+            .headers(record_headers);
+        // The key branch changes the record's type parameter, so each arm sends independently.
+        let sent = match &key {
+            Some(key) => self.producer.send(record.key(key), Timeout::Never).await,
```
svix-james: We bound the other sinks' timeout. Why not here?

crates/msgs/src/sink.rs
```diff
@@ -463,8 +499,79 @@ async fn deliver(
...
+            None => serde_json::from_slice(&msg.value)
+                .map_err(|e| format!("svix message value is not valid JSON: {e}"))?,
+        };
+        let body = serde_json::json!({
+            "eventType": self.event_type.apply(&vars),
+            "payload": payload,
+        });
+
+        let request = http.post(&url).bearer_auth(self.token).json(&body);
+        send_request(request).await
```
svix-jbrown: no idempotency token?

crates/msgs/src/sink.rs
```diff
@@ -477,6 +584,80 @@ async fn deliver(
...
+        Ok(Self {
+            producer,
+            topic: &config.topic,
+            key: config.key.as_ref().map(|t| t.compile()),
+            value: config.value.as_ref().map(|t| t.compile()),
+            headers,
+        })
+    }
+
+    async fn deliver(&self, msg: &StreamReceiveMsg) -> Result<(), String> {
```
svix-jbrown: so, the only way to get decent performance out of Kafka is to make sure we're sending in batches, but it looks like this only delivers a single message at a time. Is there some sink behavior I'm missing that gets parallelism (which will get coalesced into batch requests by librdkafka), or are we really going to end up, for each sink, only having one delivery in-flight?

crates/msgs/src/sink.rs
```diff
@@ -477,6 +584,80 @@ async fn deliver(
...
+        let record = FutureRecord::to(self.topic)
+            .payload(&value)
+            .headers(record_headers);
+        // The key branch changes the record's type parameter, so each arm sends independently.
+        let sent = match &key {
+            Some(key) => self.producer.send(record.key(key), Timeout::Never).await,
+            None => self.producer.send(record, Timeout::Never).await,
+        };
+        sent.map(|_| ())
+            .map_err(|(e, _)| format!("kafka send failed: {e}"))
```
svix-jbrown: presumably this should be separately handling permanent errors and non-permanent errors?

For example, if a message is too large for the topic, we probably want to disable the sink and somehow flag to the user, right?

svix-jbrown (commented):
should this maybe be behind a feature flag? Kafka support seems niche enough not to force every instance to build and include it?

infra/operator/Dockerfile
svix-gabriel: Originally because rdkafka was bundled by default, but feature flagging it so we don't have to include these deps anywhere wasn't actually too difficult.

crates/msgs/src/sink.rs
svix-gabriel: I agree, but Tom wanted to punt on any kind of disabling for the time being.

crates/msgs/src/sink.rs
svix-gabriel: We are already sending in batches. It's generalized to all Sinks, not kafka specific though, which might be why you missed it. https://github.com/svix/diom/pull/1324/changes#diff-bcce797b41a6865a05b7e32f35fe0299d4f555ee08a0700ba11aff3baaea29feL300

Maybe we can optimize it more, if you have suggestions there.

svix-james (approved):

crates/msgs/src/sink/svix.rs
```diff
@@ -0,0 +1,135 @@
...
+    hasher.update(msg.topic.partition.get().to_be_bytes());
+    hasher.update(b":");
+    hasher.update(msg.offset.to_be_bytes());
+    let digest: [u8; 32] = hasher.finalize().into();
+
+    let mut out = String::with_capacity(digest.len() * 2);
+    for byte in digest {
+        write!(out, "{byte:02x}").expect("writing to a String cannot fail");
+    }
+    out
```
svix-jbrown, line 120:
```suggestion
    hex::encode(digest)
```

the `hex` crate is already in our dependency tree

crates/msgs/src/sink/http.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+impl<'a> CompiledHttpSink<'a> {
+    pub(super) fn new(config: &'a HttpSinkConfig) -> Self {
+        let method = match config.method {
+            HttpMethod::Post => reqwest::Method::POST,
+            HttpMethod::Put => reqwest::Method::PUT,
+            HttpMethod::Patch => reqwest::Method::PATCH,
+        };
+        let headers = config
```
svix-jbrown, line 25: we should probably be setting a default user-agent header (other than whatever reqwest does)

crates/msgs/src/sink/mod.rs
```diff
@@ -503,6 +487,21 @@ fn build_vars(msg: &StreamReceiveMsg) -> HashMap<String, String> {
     vars
 }
 
+/// Sends a prepared request and maps the outcome to the sink delivery contract (any 2xx is success).
+async fn send_request(request: reqwest::RequestBuilder) -> Result<(), String> {
+    let response = request
+        .send()
+        .await
+        .map_err(|e| format!("request failed: {e}"))?;
```
svix-jbrown, line 495: we don't need to configure any timeouts?

crates/msgs/src/sink/svix.rs
```diff
@@ -0,0 +1,135 @@
...
+/// Builds the fallback idempotency key as a hex SHA256 of the sink identity (namespace, topic,
+/// consumer_group) and the message identity (partition, offset). This is stable across retries of
+/// the same message by the same sink, and distinct across sinks and messages. Hashing keeps the
+/// value header safe regardless of topic contents.
+fn default_idempotency_key(
+    namespace_id: NamespaceId,
+    consumer_group: &ConsumerGroup,
+    msg: &StreamReceiveMsg,
+) -> String {
+    let mut hasher = Sha256::new();
```
svix-jbrown, line 104: nit (ignorable)

this isn't security-sensitive, so we don't need a cryptographic hash and could probably save a few nanoseconds using rustc-hash or even defaulthasher

tests/it/msgs/mod.rs
```diff
@@ -3,6 +3,7 @@ mod publish;
 mod queue;
 mod retention;
 mod scheduled;
+#[cfg(feature = "kafka")]
 mod sink;
```
svix-jbrown, line 7: wait, doesn't this also contain some non-kafka tests that we should still enable?

crates/msgs/src/sink/mod.rs
```diff
@@ -150,9 +158,28 @@ async fn drain_sink<F>(
...
+        SinkConfig::Kafka(kafka_config) => match CompiledKafkaSink::new(kafka_config) {
+            Ok(sink) => CompiledSink::Kafka(sink),
+            Err(e) => {
+                tracing::error!(error = %e, "failed to build kafka producer, skipping sink");
+                return;
+            }
+        },
+        #[cfg(not(feature = "kafka"))]
+        SinkConfig::Kafka(_) => {
+            tracing::error!("kafka sink support not compiled in, skipping sink");
```
svix-jbrown, line 178: 👍  glad you covered this case

crates/msgs/src/entities.rs
```diff
@@ -485,6 +485,8 @@ pub enum HttpMethod {
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum SinkConfig {
     Http(HttpSinkConfig),
+    Svix(SvixSinkConfig),
+    Kafka(KafkaSinkConfig),
 }
```
svix-jbrown, line 490: `SinkConfig` ends up persisted inside SinkSettings as a postcard value, which means that adding fields to it may break backwards-compatibility. I don't actually think it will in this case, but still seems risky.

I wonder if we should take some of these not-that-hot fields and serialize them with `rmp_serde` instead so that we can add to them safely?

svix-gabriel:

PR #1327, opened 2026-08-12
Add support for versioning rows in DB

svix-gabriel:
This adds a `PersistibleVersion` trait and derive macro, so we can actually start versioning rows in the DB.

That's basically it. I think derive macro implementations are always hard to parse, but hopefully the [tests](https://github.com/svix/diom/pull/1327/changes#diff-0febf36721079eb4064cb109f41750619b17ad6f3ee4336ce6e0f57f6d251c56R136) are convincing enough.

### DIFF
--- a/crates/diom-derive/src/lib.rs
+++ b/crates/diom-derive/src/lib.rs
@@ -15,6 +15,7 @@ use crate::{
 };
 mod fjall_key;
 mod fjall_key_component;
+mod versioned;
 
 use self::aide::{AideAnnotateArgumentList, expand_aide_annotate};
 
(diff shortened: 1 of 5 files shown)


crates/diom-derive/src/lib.rs
```diff
@@ -85,3 +86,20 @@ pub fn macro_derive_persistable_value(input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as DeriveInput);
     derive_persistable_value(input).into()
 }
+
+/// Derive `Serialize`/`Deserialize` with a leading schema-version tag so trailing fields can be
+/// added over time without breaking reads of existing postcard data.
+///
+/// Use this **instead of** `#[derive(Serialize, Deserialize, PersistableValue)]`. Mark fields added
```
svix-jbrown, line 93: what about the other logic in PersistableValue (e.g., checking for unsafe serde attributes like `skip_serializing_if` on all transitive structs)? I assume we still need those; can se call the other macro from this one?

crates/diom-derive/src/versioned.rs
```diff
@@ -0,0 +1,250 @@
...
+        }
+    });
+
+    let field_idents = fields.iter().map(|f| &f.ident);
+    let inner_types = fields.iter().map(|f| &f.ty);
+    let expecting = format!("{name} versioned tuple");
+
+    Ok(quote! {
+        #[automatically_derived]
+        impl ::serde::Serialize for #name {
```
svix-jbrown, line 188: my worry here is that this `Serialize` impl is also going to get used from the msgpack context (for inter-server replication) and end up generating msgpack that can't be deserialized by old versions of the code, since we didn't put version tags on the msgpack.

We could either try to detect when we're serializing to msgpack and disable the version tag, or we could try to add a custom msgpack deserializer (in a separate PR landing before this one) which looks for a leading varint in the msgpack stream and strips it if so, since I **think** everything we send across the wire is a struct.

svix-gabriel:

PR #1331, opened 2026-08-18
server: improve body limit handling

svix-jbrown:
See also @svix-james's svix/svix-webhooks-private#6995 for a workaround.

We have the Axum [`DefaultBodyLimit`](https://docs.rs/axum/latest/axum/extract/struct.DefaultBodyLimit.html) in Diom right now. The way this works is that it hooks into the actual body reading and, when the limit is hit, emits a [413](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/413) and (relatively quickly) drops the read end of the stream. If the client hasn't finished sending their request yet (which depends on any intermediate buffering load balancers, network latency, whether there are other substreams on the same HTTP/2 stream if HTTP/2 is used, and the phase of the moon), the client will end up seeing this as an EPIPE from trying to write to a half-closed socket.

This adds a new layer which **just** looks at the `Content-Length` and, if it's higher than the configured max body size (but only modestly so), first reads and discards the full request and then sends the 413. This means that clients more-often actually see the 413.

This also disables body limits on the internal/loopback client.

### DIFF
--- a/crates/diom-proto/src/body_limit.rs
+++ b/crates/diom-proto/src/body_limit.rs
@@ -6,6 +6,20 @@ use axum::{
     response::{IntoResponse, Response},
 };
 use futures_util::stream::StreamExt;
+use std::time::Duration;
+
+const MAX_BODY_READ_TIME: Duration = Duration::from_secs(10);
+
+async fn read_body_forever(body: axum::body::Body, hard_body_limit: usize) {
+    let mut stream = body.into_data_stream();
+    let mut remaining = hard_body_limit;
+    while let Some(Ok(data)) = stream.next().await {
+        remaining = remaining.saturating_sub(data.len());
+        if remaining == 0 {
+            break;
+        }
+    }
+}
 
 /// Handler to ensure that large requests (but not huge) get a graceful 413
 ///
@@ -31,16 +45,12 @@ pub async fn limit_requests_body_gracefully(
         .and_then(|v| v.parse::<usize>().ok())
         && content_length > body_limit
     {
-        // consume up to hard_body_limit bytes and discard them
-        let body = request.into_body();
-        let mut stream = body.into_data_stream();
-        let mut remaining = hard_body_limit;
-        while let Some(Ok(data)) = stream.next().await {
-            remaining = remaining.saturating_sub(data.len());
-            if remaining == 0 {
-                break;
-            }
-        }
+        tokio::task::spawn(async move {
+            // consume up to hard_body_limit bytes and discard them; do this in
+            // a background task so that we can send the response right away
+            let body = request.into_body();
+            tokio::time::timeout(MAX_BODY_READ_TIME, read_body_forever(body, hard_body_limit)).await
+        });
         return (
             http::StatusCode::PAYLOAD_TOO_LARGE,
             MsgPackOrJson(ErrorBody::invalid_input(
--- a/crates/diom-proto/src/body_limit.rs
+++ b/crates/diom-proto/src/body_limit.rs
@@ -1,0 +1,55 @@
+use super::msgpack_or_json::MsgPackOrJson;
+use crate::ErrorBody;
+use axum::{
+    extract::{Request, State},
+    middleware::Next,
+    response::{IntoResponse, Response},
+};
+use futures_util::stream::StreamExt;
+
+/// Handler to ensure that large requests (but not huge) get a graceful 413
+///
+/// The axum BodyLimit middleware (and tower-http's RequestBodyLimitLayer) work
+/// by actually shutting down the read of the stream when the limit is hit; this causes
+/// clients to get an EPIPE, and (depending on whether the end-of-stream packet
+/// from the server is received before or after the last bytes are sent), can prevent
+/// clients from actually seeing the 413.
+///
+/// This middleware *only* looks at Content-Length, and emits a 413 without doing anything to the
+/// body. This can be tricked by HTTP/1.1 chunked-encoding or HTTP/1.0 connection-close or a whole
+/// variety of HTTP/2 things, so it's important to also use one of those other middlewares,
+/// configured for a higher value.
+pub async fn limit_requests_body_gracefully(
+    State((body_limit, hard_body_limit)): State<(usize, usize)>,
+    request: Request,
+    next: Next,
+) -> Response {
+    if let Some(content_length) = request
+        .headers()
+        .get(http::header::CONTENT_LENGTH)
+        .and_then(|v| v.to_str().ok())
+        .and_then(|v| v.parse::<usize>().ok())
+        && content_length > body_limit
+    {
+        // consume up to hard_body_limit bytes and discard them
+        let body = request.into_body();
+        let mut stream = body.into_data_stream();
+        let mut remaining = hard_body_limit;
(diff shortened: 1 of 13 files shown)


src/serve.rs
```diff
@@ -148,6 +148,11 @@ pub async fn run_with_listeners(
                 .allow_methods(Any)
                 .allow_headers(AllowHeaders::mirror_request())
                 .max_age(Duration::from_secs(600)),
+            DefaultBodyLimit::max(cfg.max_body_size * 4),
```
svix-james: I don't really get why we want the x4 allowance

svix-jbrown: it has to be bigger than the configured max_body_size, or else it'll kick in first (because the way `DefaultBodyLimit` works is it actually wraps the `Body` type to handle all those annoying things like HTTP/1.1 chunked encoding where you can't know the size of the body until you read it) and we'll be back where we started.

The 4x is just an arbitrary number to hope that the client will receive the 413 before we finish reading and discarding their whole body.

svix-james: Can we do 2x instead? Just to limit the potential overage.

svix-jbrown: okey dokey

svix-james (dismissed):

crates/diom-proto/src/body_limit.rs
```diff
@@ -0,0 +1,55 @@
...
+        // consume up to hard_body_limit bytes and discard them
+        let body = request.into_body();
+        let mut stream = body.into_data_stream();
+        let mut remaining = hard_body_limit;
+        while let Some(Ok(data)) = stream.next().await {
+            remaining = remaining.saturating_sub(data.len());
+            if remaining == 0 {
+                break;
+            }
+        }
```
svix-jplatte: In Svix, we do this in a separate tokio task, so we can immediately return the `PAYLOAD_TOO_LARGE` response. Should we not do the same here?

crates/diom-proto/src/msgpack_or_json.rs
```diff
@@ -75,7 +75,7 @@ where
 
                 _ => {
                     tracing::error!("Error reading body as bytes: {e}");
-                    MsgPackOrJsonRejection::InternalServerError {
+                    MsgPackOrJsonRejection::BadRequestError {
```
svix-jplatte, line 78: Why this change?

src/cfg/defaults.rs
```diff
@@ -136,3 +136,8 @@ pub(super) const fn sink_max_task_duration() -> NonZeroDurationMs {
 pub(super) fn default_database_size() -> MemorySize {
     MemorySize::Percent(20)
 }
+
+pub(super) const fn default_max_body_size() -> usize {
+    // 2 MB
+    2 * 1000 * 1000
```
svix-jplatte, line 142: You don't like MiB? ^^

tests/it/common.rs
```diff
@@ -15,7 +17,7 @@ async fn kv_set(
 ) -> anyhow::Result<TestResponse> {
     client
         .post("v1.kv.set")
-        .json(json!({
+        .msgpack(json!({
```
svix-jplatte, line 20: I'm pretty surprised this actually works. ISTR that encoding `serde_json::Value` w/ rmp_serde is problematic. Or maybe it's just deserialization? Either way, I'd prefer if we serialized a struct here. One alternative _might be_ [`typed_json!`](https://docs.rs/typed-json/0.1.1/typed_json/), which probably contains fewer / no `serde_json`-specific hacks.

src/cfg/mod.rs
```diff
@@ -683,6 +683,15 @@ pub struct ConfigurationInner {
     #[env_overridable(nest_with_prefix("OPENTELEMETRY"))]
     #[dumpable_config(nest)]
     pub opentelemetry: OpenTelemetryConfig,
+
+    /// Maximum size for a request body, in bytes.
+    ///
+    /// This should be set to a large enough value to handle all real requests that you receive,
+    /// but small enough not to consume all of your RAM. Requests more than 4x this size will
```
svix-jplatte:
```suggestion
    /// but small enough not to consume all of your RAM. Requests more than 2x this size will
```

crates/diom-proto/src/msgpack_or_json.rs
svix-jbrown: Because this isn't a 500? There's no reason to be alerted if, e.g., the customer cuts off their body half-way through.

src/cfg/defaults.rs
svix-jbrown: tower-http uses MB instead of MiB so I figured I'd match it

tests/it/common.rs
svix-jbrown: I actually made this change because the blowup from encoding the bytes as serde_json turned my 2MB payload into like a 10MB payload. It seems to work fine...

crates/diom-proto/src/msgpack_or_json.rs
svix-jplatte: Are you sure that's this branch would only be taken due to client errors?

src/cfg/defaults.rs
svix-jplatte: Huh, didn't know that.

crates/diom-proto/src/body_limit.rs
svix-jbrown: ¯\_(ツ)_/¯ it just changes the nature of the error (it's better for HTTP/2 clients, but sometimes worse for HTTP/1.1 clients). I can do it that way if you want to match webhooks.

crates/diom-proto/src/msgpack_or_json.rs
svix-jbrown: This path isn't hittable at all right now because BytesRejection only has one variant (and is non-exhaustive). But any problem reading raw bytes from the request has to be a client problem by definition, right?

crates/diom-proto/src/body_limit.rs
svix-jplatte: Can you elaborate? I thought the whole point here was to make sure the client sees the http status, how can sending it as early as possible be worse in some cases?

crates/diom-proto/src/msgpack_or_json.rs
svix-jplatte: I don't think so. For instance I could imagine the extraction being changed to use fallible allocation APIs, so then this code path could be hit on OOM.

svix-jbrown: I don't think OOM is a meaningfully-likely case, and I think _any_ other case where we have issues reading the request are the client's problem

crates/diom-proto/src/body_limit.rs
svix-jbrown: It depends on how the client is written; HTTP/2 as a protocol is asychronous, so pretty much all HTTP/2 clients will actually read the response as soon as it appears, but traditional synchronous HTTP/1 clients aren't going to read from the socket until after they finish writing, so you're just sending noise that they may never see.

Anyhow, the last push move it to a task.

svix-jplatte (approved):

PR #1336, opened 2026-08-27
implement more-graceful shutdown

svix-jbrown:
This changes shutdown on signal to a two-phase process: first, we make the readiness check fail (so we stop getting traffic from the LB) and issue a TransferLeader request if we're the leader, *then* we actually shut down.

### DIFF
--- a/src/core/cluster/app.rs
+++ b/src/core/cluster/app.rs
@@ -15,7 +15,7 @@ use headers::{Authorization, authorization::Bearer};
 use http::StatusCode;
 use openraft::{
     ChangeMembers,
-    raft::{AppendEntriesRequest, InstallSnapshotRequest, VoteRequest},
+    raft::{AppendEntriesRequest, InstallSnapshotRequest, TransferLeaderRequest, VoteRequest},
 };
 use serde::Serialize;
 use tap::{Pipe, TapFallible};
@@ -59,6 +59,7 @@ pub fn router(cfg: &Configuration) -> axum::Router<AppState> {
         .route("/repl/discover", get(discover))
         .route("/repl/raft/append_entries", post(append_entries))
         .route("/repl/raft/vote", post(vote))
+        .route("/repl/raft/transfer_leader", post(transfer_leader))
         .route("/repl/raft/stream-snapshot", post(stream_snapshot))
         .route(
             "/repl/raft/handle-forwarded-write",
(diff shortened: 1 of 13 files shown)


src/core/cluster/app.rs
```diff
@@ -59,6 +59,7 @@ pub fn router(cfg: &Configuration) -> axum::Router<AppState> {
         .route("/repl/discover", get(discover))
         .route("/repl/raft/append_entries", post(append_entries))
         .route("/repl/raft/vote", post(vote))
+        .route("/repl/raft/transfer_leader", post(transfer_leader))
```
svix-jplatte, line 62: Why do we have a mix of `_`s and `-`s in the repl API ._.

src/core/cluster/app.rs
```diff
@@ -162,6 +163,19 @@ async fn vote(
     state.raft.vote(body).await.pipe(rpc_response)
 }
 
+#[tracing::instrument(skip_all)]
+async fn transfer_leader(
+    Extension(state): Extension<RaftState>,
+    MsgPack(body): MsgPack<TransferLeaderRequest<TypeConfig>>,
+) -> impl IntoResponse {
+    tracing::trace!("recording leadership-transfer request",);
```
svix-jplatte:
```suggestion
    tracing::trace!("recording leadership-transfer request");
```

src/core/cluster/raft.rs
```diff
@@ -289,6 +290,52 @@ pub async fn initialize_raft(
...
+                            }
+                        })
+                })
+                .await
+            else {
+                tracing::warn!("error looking up other peers at shutdown; ignoring");
+                return;
+            };
+            let Some(other_node_id) = other_peer else {
+                tracing::debug!("no other node to hand off leadership to");
```
svix-jplatte: Would it not make more sense to move this to where we first return `None` (this node is not a leader), and log at a higher level (`warn`?) in the second branch (there is no other node)?

src/core/cluster/raft.rs
```diff
@@ -289,6 +290,52 @@ pub async fn initialize_raft(
...
+                            if *node_id != my_node_id {
+                                Some(*node_id)
+                            } else {
+                                None
+                            }
+                        })
+                })
+                .await
+            else {
+                tracing::warn!("error looking up other peers at shutdown; ignoring");
```
svix-jplatte: Does the `Err()` not contain useful information?

svix-jplatte (dismissed):
Got some questions / nitpicks, none of them blocking.

src/core/cluster/app.rs
svix-jbrown: just laziness; I can clean that up in a follow-up

src/core/cluster/raft.rs
svix-jbrown: it's not really a warn because it just means we're in a 1-node cluster, but I can move it up.

src/core/cluster/raft.rs
svix-jbrown: good point, fixed

src/core/cluster/raft.rs
svix-jplatte: So we wouldn't hit it if other nodes exist, but are offline or something?

svix-jplatte (approved):

src/core/cluster/raft.rs
svix-jbrown: they'd still be in the membership array unless they were actually evicted by a vote; no vote can ever leave us with a 1-node cluster because 1 node can never be a quorum.

PR #1337, opened 2026-08-27
deps: bump chacha20

svix-jbrown:
See https://github.com/RustCrypto/stream-ciphers/pull/580; they yanked 0.10.1.

It makes me a little anxious to bump this without a RUSTSEC advisory or any official posting about why they did the yank outside of that github thread, but @tarcieri is the right person to make that call, so hopefully 0.10.2 isn't secretly malware.


svix-lucho (approved):

tarcieri:
It was UB in the SSE2 backend (also we tend to document yanks in CHANGELOG.md)

PR #1340, opened 2026-09-02
ci: Split PR comment actions into separate workflow

svix-jplatte:
This should allow CI to succeed on PRs from forks (such as #1305).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,5 @@
 name: Diom CI
+
 on:
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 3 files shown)


svix-jplatte:
It's expected that there are no coverage / test report comments on this PR, since the new `pr-comments.yml` workflow will only start running for PRs once it's part of the main branch.

svix-jplatte:
Of course, there's no way to really know whether it will work before merging this ^^'

.github/workflows/ci.yml
```diff
@@ -104,41 +106,6 @@ jobs:
             ./junit.xml
             ./coverage-html
```
svix-jbrown, line 108: shouldn't this also be uploading `coverage.json`?

svix-jplatte: Thanks for noticing, yes.

svix-jbrown (approved):

PR #1346, opened 2026-09-03
Add support for versioning rows in DB

svix-gabriel:
This is part 1 of a series of PRs to support versioning data stored in the DB.

One of the main goals I have is that upgrades / deployments to a Diom cluster should be easy and low risk, while still being able to add columns, extend Operations, etc. That imposed a couple of constraints:

1. Data stored in Fjall needs to be backwards compatible, so that new code can read old database rows.
2. Data stored in Fjall needs to be forwards compatible, so we can safely revert versions.
3. During rolling upgrades, nodes of different versions will be talking to each other, and the APIs they use, and replication, should not break during this upgrade.
4. We can't have data loss during an upgrade. e.g. if I add a column in v2, upgrade a cluster from v1 to v2, and node v2 replicates to node v1, we can't have data loss because v1 didn't serialize all the new fields added in v2.

To that end, here are the series of changes, each in their own PR in the stack:

1. Add a `PersistableVersioned` derive macro, to be used explicitly on Table row structs. This generates all of the (de)serialization code necessary to make sure that new columns can be added to data stored in Fjall in backwards AND forwards compatible way. <--- YOU ARE HERE
	* Important to note that no actual data models are changed to use the new macro, and no colums are added, so this PR should be safe to merge and revert in isolation
2. Migrate all extant rows to use `PersistiableVersioned` macro.
3. Add a `feature_version` tracking accessible from the raft state, and consensus logic so that during an upgrade, the `feature_version` field on an individual node is updated only once we know that all nodes are on the same version.
4. Add plumbing so that all Operations have visibility on the `feature_version`, and can gate new fields and even new logic on consensus of all nodes in the cluster being on the same version.

The end result is that, if you need a new column to a table in fjall, it will look something like
```diff

struct UserRow {
    id: UserId,
    name: String,
+   #[since(1)]
+   nickname: Option<String>,
}

// ...

struct CreateUserOperation {
    id: UserId,
    name: String,
+   nickname: Option<String>,
}

impl CreateUserOperation {
    async fn apply_real(self, state: &mut UsersState) -> CreateUserResponse {
-       let row = UserRow { id: self.id, name: self.name };
+       let row = UserRow { id: self.id, name: self.name, nickname: self.nickname };
        state.users.insert(row.id, &row);
        CreateUserResponse { id: row.id }
    }
}

// ...

async fn create_user(repl: &RaftState, req: CreateUserReq) -> Result<...> {
+   let nickname = if repl.feature_version() >= 1 { req.nickname } else { None };
+   let op = CreateUserOperation { id: new_id, name: req.name, nickname };
-   let op = CreateUserOperation { id: new_id, name: req.name };
    repl.client_write(op).await
}

```

You can see this in the final PR in the stack, https://github.com/svix/diom/pull/1356

This does add a decent bit of boilerplate in code when adding new fields, though I don't think the end result
is terrible from a code quality perspective.

The advantages are
1. No data loss during a rolling upgrade, since the Leader will never try to write data that the follower doesn't know about yet
2. Reverts are safe, since deserialization is always forward compatible with old data

So from someone running their own Diom cluster, the deployment/upgrade path isn't any more complex than it was before.

### DIFF
--- a/crates/diom-derive/src/versioned.rs
+++ b/crates/diom-derive/src/versioned.rs
@@ -1,0 +1,287 @@
+use proc_macro2::TokenStream;
+use quote::quote;
+use syn::{
+    DeriveInput, Expr, Ident, LitInt, Token,
+    parse::{Parse, ParseStream},
+    spanned::Spanned,
+};
+
+/// Parsed `#[since(N)]` or `#[since(N, default = EXPR)]` field attribute.
+struct SinceAttr {
+    version: u32,
+    default: Option<Expr>,
+}
+
+impl Parse for SinceAttr {
+    fn parse(input: ParseStream<'_>) -> Result<Self, syn::Error> {
+        let lit: LitInt = input.parse()?;
+        let version = lit.base10_parse()?;
+        let mut default = None;
+        if input.peek(Token![,]) {
+            let _: Token![,] = input.parse()?;
+            let key: Ident = input.parse()?;
+            if key != "default" {
+                return Err(syn::Error::new(key.span(), "expected `default`"));
+            }
+            let _: Token![=] = input.parse()?;
+            default = Some(input.parse()?);
+        }
+        Ok(SinceAttr { version, default })
+    }
+}
+
+struct VersionedField {
+    ident: Ident,
+    ty: syn::Type,
+    since: u32,
+    default: Option<Expr>,
+    nested: bool,
+}
+
+fn parse_since(field: &syn::Field) -> Result<Option<SinceAttr>, syn::Error> {
+    for attr in &field.attrs {
+        if attr.path().is_ident("since") {
+            return Ok(Some(attr.parse_args::<SinceAttr>()?));
+        }
+    }
+    Ok(None)
+}
+
+/// A `#[nested]` marker stores a struct-valued field with length delimiters (as its own postcard blob) so
+/// the nested struct can add trailing fields without corrupting the fields that follow it.
+fn parse_nested(field: &syn::Field) -> bool {
+    field
+        .attrs
+        .iter()
+        .any(|attr| attr.path().is_ident("nested"))
+}
+
+/// Parses an optional `#[versioned(row_type = EXPR)]` container attribute. When present, the derive
+/// also generates a `TableRow` impl wired to the versioned storage envelope.
+fn parse_row_type(input: &DeriveInput) -> Result<Option<Expr>, syn::Error> {
+    for attr in &input.attrs {
+        if attr.path().is_ident("versioned") {
+            let mut row_type = None;
+            attr.parse_nested_meta(|meta| {
+                if meta.path.is_ident("row_type") {
+                    row_type = Some(meta.value()?.parse::<Expr>()?);
+                    Ok(())
+                } else {
+                    Err(meta.error("expected `row_type`"))
+                }
+            })?;
+            return row_type
+                .ok_or_else(|| {
+                    syn::Error::new(attr.span(), "missing `row_type` in #[versioned(...)]")
+                })
+                .map(Some);
+        }
+    }
+    Ok(None)
+}
+
+fn collect_fields(input: &DeriveInput) -> Result<Vec<VersionedField>, syn::Error> {
+    let syn::Data::Struct(data) = &input.data else {
+        return Err(syn::Error::new(
+            input.ident.span(),
+            "PersistableVersioned can only be derived for structs",
+        ));
+    };
+    let syn::Fields::Named(fields) = &data.fields else {
+        return Err(syn::Error::new(
+            input.ident.span(),
+            "PersistableVersioned requires named fields",
+        ));
+    };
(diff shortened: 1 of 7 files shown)


crates/diom-derive/src/versioned.rs
```diff
@@ -0,0 +1,287 @@
...
+fn parse_since(field: &syn::Field) -> Result<Option<SinceAttr>, syn::Error> {
+    for attr in &field.attrs {
+        if attr.path().is_ident("since") {
+            return Ok(Some(attr.parse_args::<SinceAttr>()?));
+        }
+    }
+    Ok(None)
+}
+
+/// A `#[nested]` marker stores a struct-valued field with length delimiters (as its own postcard blob) so
```
svix-jbrown, line 50: we're not worried about conflicting with some other derive macro that uses the bare `#[nested]`?

crates/diom-derive/src/versioned.rs
```diff
@@ -0,0 +1,287 @@
...
+        let (since, default) = match parse_since(field)? {
+            Some(a) => (a.version, a.default),
+            None => (0, None),
+        };
+        out.push(VersionedField {
+            ident,
+            ty: field.ty.clone(),
+            since,
+            default,
+            nested: parse_nested(field),
```
svix-jbrown, line 115: is there any way we can check and lint on whether we have nested structs without a `#[nested]` on them?

crates/diom-derive/src/versioned.rs
```diff
@@ -0,0 +1,287 @@
...
+            input.ident.span(),
+            "PersistableVersioned requires named fields",
+        ));
+    };
+    if !input.generics.params.is_empty() {
+        return Err(syn::Error::new(
+            input.generics.span(),
+            "PersistableVersioned does not support generic types",
+        ));
+    }
```
svix-jbrown, line 101: there are some other checks in PersistableValue (like making sure that `skip_serializing_if` is never set on a struct field, and that enums use the right variant encoding). Since this new derive is _replacing_ the derive for `PersistableValue`, it needs the same checks too, right?

svix-jbrown (commented):
looks good; a couple of notes

crates/diom-derive/src/versioned.rs
```diff
@@ -0,0 +1,287 @@
...
+fn parse_nested(field: &syn::Field) -> bool {
+    field
+        .attrs
+        .iter()
+        .any(|attr| attr.path().is_ident("nested"))
+}
+
+/// Parses an optional `#[versioned(row_type = EXPR)]` container attribute. When present, the derive
+/// also generates a `TableRow` impl wired to the versioned storage envelope.
+fn parse_row_type(input: &DeriveInput) -> Result<Option<Expr>, syn::Error> {
```
svix-jplatte, line 61:
```suggestion
fn parse_row_type(input: &DeriveInput) -> syn::Result<Option<Expr>> {
```

here and elsewhere.

crates/diom-derive/src/versioned.rs
svix-gabriel: Can you clarify what you mean? Since PersistableVersioned implemented these serde traits for you, if you try to do something like add skip_serializing_if, it'll just fail to compile. e.g.

```diff
+  #[derive(diom_core::PersistableVersioned, serde::Serialize, serde::Deserialize, PartialEq, Debug)]
-  #[derive(diom_core::PersistableVersioned, PartialEq, Debug)]
    #[versioned(row_type = 0)]
    struct Fixture {
        a: u32,
        b: u32,
        #[since(1)]
+      #[serde(skip_serializing_if = "Option::is_none")]
        c: Option<u32>,
    }
```

You'll get
```
error[E0119]: conflicting implementations of trait `_::_serde::Serialize` for type `Fixture`
   --> crates/fjall-utils/src/lib.rs:141:47
    |
141 |     #[derive(diom_core::PersistableVersioned, serde::Serialize, serde::Deserialize, PartialEq, Debug)]
    |              -------------------------------  ^^^^^^^^^^^^^^^^ conflicting implementation for `Fixture`
    |              |
    |              first implementation here
```

crates/diom-derive/src/versioned.rs
svix-gabriel: Are there popular macros with a `#[nested]` annotation?

My big worry would be conflicting w/serde, so for now we're clean. I think if that changes it's relatively easy to rename the annotation.

crates/diom-derive/src/versioned.rs
svix-jbrown: aha, that covers it

svix-jbrown (approved):

crates/diom-derive/src/versioned.rs
svix-gabriel: I believe so, I think it makes more sense in https://github.com/svix/diom/pull/1347 so that's where I'm getting it to work.

PR #1347, opened 2026-09-03
Implement snapshot testing for table data and operations

svix-gabriel:
This adds snapshot testing so changes to table data and operations can more easily be audited.

There's two elements to this

1. human readable snapshots that can be updated with `just schema-snapshot`. This way, schema changes are called out for human review automatically
2. tests for each module that will fail if a backwards INcompatible change is made in the future. I intend for these to act as proof that switching to `Persistable*` derive macros are not a breaking change. While these tests aren't automated/will need active maintenance if rows are changed in the future, they at least setup some kind of framework for people to follow in the future.


crates/diom-core/src/schema_shape.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub kind: &'static str,
+    pub members: &'static [MemberShape],
+}
+
+inventory::collect!(SchemaShape);
+
+/// Renders every registered shape into a stable, sorted, human-readable snapshot.
+///
+/// The output is deterministic across runs (sorted by module path then type name), so it can be
+/// committed and diffed by CI.
```
svix-jbrown, line 35: the snapshot is nice, but I can imagine in the future wanting to do spelunking in it without having to write a custom parser for this. Is there any way we could make it still somewhat human readable, but also machine readable? YAML or TOML or RON or something?

crates/diom-core/src/schema_shape.rs
```diff
@@ -0,0 +1,86 @@
...
+        out.push_str(&format!(
+            "{}::{} ({})\n",
+            shape.module_path, shape.type_name, shape.kind
+        ));
+        for member in shape.members {
+            // A unit enum variant has no type, so render just its name (no trailing `: `).
+            let mut line = format!("    {}", member.name);
+            if !member.ty.is_empty() {
+                line.push_str(&format!(": {}", tidy_type(member.ty)));
+            }
```
svix-jbrown, line 51: I notice that this type is a bare type (un-qualified by module); do we ever have any cases where there are types in different modules with the same name but different implementations? Is there a way to get the absolute path to the type?

svix-jbrown (commented):
that turned out to be much less code than I expected

.github/workflows/on-pr.yml
```diff
@@ -2,7 +2,8 @@ name: CI on PR
 
 on:
   pull_request:
-    branches: ['*']
+    # ** matches base branches containing a slash (stacked PRs). A single * does not.
+    branches: ['**']
```
svix-jplatte, line 6: You can just leave off the `branches` property entirely if you want to make it branch-independent.

PR #1350, opened 2026-09-04
Swap models over to use PersistableVersioned 

svix-gabriel:
With the byte-level snapshot tests added in https://github.com/svix/diom/pull/1347/changes/7aa1183142b074fc5cabf5749933872cfc66ab1e still passing, it should hopefully be convincing that this won't break anything on disk, and at least until new columns are added, is also safely revertible.

I had to update the macro a bit to support generics, specifically for Namespace.


crates/diom-derive/src/versioned.rs
```diff
@@ -220,38 +238,43 @@ pub(crate) fn derive(input: TokenStream) -> Result<TokenStream, syn::Error> {
...
-        let nested = f.nested;
-        quote! {
-            diom_core::schema_shape::MemberShape {
-                name: #member_name,
-                ty: #ty_str,
-                since: #since,
-                nested: #nested,
+    // Register this type's shape for the schema snapshot test. Generics are skipped, since
+    // can't register a specific instantiation (concrete instances are guarded by byte
+    // fixtures instead).
```
svix-jbrown, line 243: if it's generic over lifetime only, what do you think about registering it with `'static`?

PR #1355, opened 2026-09-09
Add cluster feature-version coordination

svix-gabriel:
This is the final pillar of the versioning machinery. The goal is for nodes within a cluster to be able to reach consensus on the highest supported version in the cluster. This way, code can selectively gate new data (e.g. columns, operations, etc.) behind a version, and nodes will only ever execute that code when ALL nodes in the cluster support it. This way, we can support adding new columns without risking runtime errors or dataloss during a rolling upgrade.

The way we do this is as follows:
* Add a `feature_version` field, stored locally in the raft state, similar to cluster_id. It's setup so it can only ever increase.
* Each node also gets a "version range", defined in code. The upper bound value will only ever increment when we add new features, like a new column. The lower bound will increment if we ever drop support for a feature. I don't foresee incrementing the min any time soon, but seems like a useful option to have.
* A background worker running on the leader will poll every voting node to see their supported versions, and advance the `feature_version` in the raft state to the highest version number shared by all voting nodes. This is done using the `AdvanceFeatureVersionOperation`, which will fail if it can't increment to the target feature version.
* If a voting node's supported versions are incompatible with the cluster's, then the node simply can't join the cluster.

When adding a new feature that requires versioning, the code will be gated on the feature_version, shown here https://github.com/svix/diom/pull/1356

A few decisions worth calling out that I wasn't sure about
- If a node can't join the cluster, it'll simply shutdown and report an `error!`. This seemed like the behavior I'd want, since it would make a non-version compatible node very obvious during a deploy. Maybe there's a better option here?
- When checking supported versions to decided on the feature version, we only check the voting nodes, not the learner nodes. I don't know how many "learner" nodes we'd expect in a typical deployment, so perhaps we could check against all nodes just fine. In theory, this means you could get all the voter nodes incrementing the feature_version, but a learner node can't recognize that version's data. If this happens, I believe the learner node would simply fail when applying `AdvanceFeatureVersionOperation`, and halt applying future operations, which is what we want. I don't have strong opinions whether this is better or worse than simply checking learners supported feature range before incrementing the feature version across the cluster.
- In order to make this testable, I had to update the shutdown logic to use an explicit cancellation token, so we could test this without shutting down the whole process.


src/core/cluster/version.rs
```diff
@@ -0,0 +1,48 @@
...
+//! discovery and health checks so an incompatible node parks instead of joining a cluster it cannot
+//! interoperate with.
+use diom_operations::FeatureVersion;
+use serde::{Deserialize, Serialize};
+
+/// The range of feature versions this build can operate at.
+pub(crate) const SUPPORTED_VERSION_RANGE: VersionRange = VersionRange { min: 0, max: 0 };
+
+/// An inclusive range of feature versions a node supports.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
```
svix-jbrown, line 13: should this be using a persistable struct itself, or should we just be very careful never to change it?

src/cfg/mod.rs
```diff
@@ -441,6 +445,26 @@ pub struct ClusterConfiguration {
...
+    /// How often the leader polls voters to decide whether to advance the cluster feature version.
+    ///
+    /// This should only be configured when testing the feature-version subsystem.
+    #[serde(
+        rename = "feature_version_advance_interval_ms",
+        default = "defaults::cluster_feature_version_advance_interval"
+    )]
+    #[env_overridable(skip)]
+    #[dumpable_config(skip)]
+    pub feature_version_advance_interval: NonZeroDurationMs,
```
svix-jbrown, line 467: I'm still really skeptical of advancing the protocol and storage versions automatically, given that once it's advanced you can never roll back the cluster. I know @tasn wants us to have automatic advancement, but I'd really vote for something like most other databases, where you separately control the "no going back" phase (e.g., clickhouse's `compatibility` setting or kafka's `inter.broker.protocol.version`) from actually upgrading the code.

I think a way to handle this would be to actually make `supported_feature_versions` a supported (heh) feature, and have a validator that it must be a strict subset of the compiled-in default and that its max must be no more than 1 below the compiled-in max.

That way, users who want a safer upgrade path can, when upgrading from (say) 3..=10 to 3..=11 first manually set `supported_feature_versions` to `Some(3..=10)`, then upgrade and let it bake for a few days, then unset `supported_feature_versions`...

PR #1370, opened 2026-09-21
py: run lint checks on all supported versions; start testing py3.15

svix-jbrown:
This actually runs the build steps under all the supported Python versions, and adds a couple of new ones to the stable.


.github/workflows/python.yml
```diff
@@ -8,12 +8,17 @@ jobs:
   lint:
     name: Lint
     runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14", "3.15.0rc2", "pypy@3.11"]
```
svix-pluiz, line 13: We're already testing different Python versions [in the SDK tests](https://github.com/svix/diom/blob/main/.github/workflows/sdk-integration.yml). Should we really do the same here?

svix-pluiz: If that's the case, let's at least match the same python version in both workflows

svix-jbrown: okey dokey

svix-pluiz (approved):

